PR #115, opened 2015-05-11
first draft of review criteria.

katyhuff:
At the suggestion of @jaberg , I wrote a quick draft of possible suggested review criteria for scipy to assist the authors and reviewers of full papers for the proceedings. 

I honestly didn't put a lot of original thought into this. Rather, I looked at [what IEEE requires](http://www.ieee.org/conferences_events/conferences/publishing/paper_acceptance_criteria.pdf) and thought about what I personally think I would like SciPy papers to strive for. I wrote those things down in this document. 

As was noted by @jaberg , this is important, actually, as these review criteria could, in part, set forth a potentially long-term tone and a flavor for SciPy. Previously, to my knowledge, we have allowed the reviewers to almost entirely set their own review parameters, which was confusing for all involved.

Though this is intended to be a living document and no version should ever be final or overly prescriptive, I would be grateful for first-glance feedback from others on these review criteria. In particular, folks at the intersection of both academia and SciPy may have the most appropriate perspective on what should be contained in this document. For this reason, I am hereby pinging a few academic-types who might like to opine on this before it is pulled in tomorrow evening: 

@scopatz , @ellisonbg , @jakevdp , @sjsrey , @mandli , @kthyng , @MatthewTurk , @jarrodmillman 

If you have a little time, please do your worst.

Thanks. :)

### DIFF
--- a/review_criteria.md
+++ b/review_criteria.md
@@ -1,0 +1,65 @@
+# Suggested Review Criteria for SciPy Proceedings
+
+A small subcommittee of the SciPy 2015 organizing committee has created this 
+set of suggested review criteria
+to help guide authors and reviewers alike. Suggestions and amendments to these 
+review criteria are enthusiastically welcomed via discussion or pull request.
+
+
+## Written Quality
+
+- Prose should be written in English.
+- Prose should clearly communicate the narrative.
+- The written style should convey information that furthers the knowledge or 
+  research of the reader.
+- Due to the interdisciplinary nature of SciPy, highly domain-specific jargon 
+  should be avoided or explained where possible.
+
+## Technical Content
+
+- The technical content should be scientifically sound.
+- Computational content should, likewise, be verified.
+- The work should describe the development or use of python software for 
+  approaching a problem in a domain within the scope of the conference.
+
+## Novelty and Impact
+
+- The work should employ new or innovative methods or approaches to a problem.
+- The work should advance the state of a scientific domain, the practice 
+  of scientific computing itself, or another subject area within the scope of 
+  the conference.
+
+## Verifiability
+
+- Software descriptions should be accompanied by references to or examples of 
+  representative source code.
+- Source code essential to the conclusions of the paper should be made 
+  available to the reader.
+- Data sources should be identified (e.g., with citation to a persistent DOI).
+- Analysis should be accompanied by a workflow description sufficient
+  to reproduce the results.
+
+## Other Requirements
+
+- All mentioned software should be formally cited wherever possible.
+- Acronyms should be spelled out upon first mention.
+- License conditions on images and figures must be respected (Creative Commons, 
+  etc.).
+- Mathematical and other symbols should be defined.
+- Definitions should include consistent units where appropriate.
+
+### Length
+
+- The compiled version should be no longer than 7 pages, including figures.  
+
+### Figures
+
+- All figures and tables should have captions. 
+- Figure text should be of a readable size.
+
+### Code Snippets
+
+- Code snippets should be formatted to fit inside a single column without
+  overflow.
+- If not in Python, the language in the code snippet should be mentioned.
+- Code snippets should follow a common style guide. PEP8 is preferred.
--- a/README.md
+++ b/README.md
@@ -39,6 +39,14 @@ allowed during the open review period which ends July 5th.
   overflow.
 - Avoid custom LaTeX markup where possible.
 
+## Review Criteria
+
+A small subcommittee of the SciPy 2015 organizing committee has created [this 
+set of suggested review 
+criteria](https://github.com/scipy-conference/scipy_proceedings/review_criteria.md) 
+to help guide authors and reviewers alike. Suggestions and amendments to these 
+review criteria are enthusiastically welcomed via discussion or pull request.
+
 ## Other markup
 
 Please refer to the example paper in ``papers/00_vanderwalt`` for
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ paper:
 Pull requests are to be submitted by **June 15th**, but modifications are
 allowed during the open review period which ends July 5th.
 
+
 ## General Guidelines
 
 - All figures and tables should have captions.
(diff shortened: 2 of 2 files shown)


shjb:
Great work, Katy!

In terms of scope, this is exactly what I was thinking about in my earlier email. I hope it will prove helpful to have this spelled out.

In terms of content, this captures really well how I personally think of SciPy.

review_criteria.md
```diff
@@ -0,0 +1,65 @@
...
+
+- All figures and tables should have captions. 
+- Figure text should be of a readable size.
+
+### Code Snippets
+
+- Code snippets should be formatted to fit inside a single column without
+  overflow.
+- If not in Python, the language in the code snippet should be mentioned.
+- Code snippets should follow a common style guide. PEP8 is preferred.
```
shjb: Code snippets should follow the PEP8 coding style.

review_criteria.md
```diff
@@ -0,0 +1,65 @@
...
+- Computational content should, likewise, be verified.
+- The work should describe the development or use of python software for 
+  approaching a problem in a domain within the scope of the conference.
+
+## Novelty and Impact
+
+- The work should employ new or innovative methods or approaches to a problem.
+- The work should advance the state of a scientific domain, the practice 
+  of scientific computing itself, or another subject area within the scope of 
+  the conference.
```
shjb: Let's spell this out a little more rather than referring to "the conference".

Maybe...
"""
The work should advance some practice in mathematics, science, or engineering.
"""

review_criteria.md
```diff
@@ -0,0 +1,65 @@
...
+  research of the reader.
+- Due to the interdisciplinary nature of SciPy, highly domain-specific jargon 
+  should be avoided or explained where possible.
+
+## Technical Content
+
+- The technical content should be scientifically sound.
+- Computational content should, likewise, be verified.
+- The work should describe the development or use of python software for 
+  approaching a problem in a domain within the scope of the conference.
```
shjb: A possibly contentious suggestion -- when the technical contribution is a program or library in Python, then that program or library must be primarily an open-source project. SciPy proceedings is meant to disseminate technical knowledge, and should not be used to advertise closed-source products.

review_criteria.md
katyhuff: Thanks for this comment. I also about that. One problem is that we have been expanding the conference to include things like "social science" as a domain symposium topic or "education" as a main track topic. I wanted to make sure  that when the conference includes those topics, that the papers are still reviewed fairly.

Do you have any thoughts about how to make that work? My solution was to refer to the scope of the conference (which changes yearly)

review_criteria.md
katyhuff: Hi @shjb . Did you see, below, in verifiability where I say that essential source code should be made available to the reader?
```
- Software descriptions should be accompanied by references to or examples of representative source code.
- Source code essential to the conclusions of the paper should be made available to the reader.
```

review_criteria.md
katyhuff: I hesitated with that mostly because of Google's reliance on their own non-pep8 style guide. Is it really essential that it all be PEP8, or can we be open to other styles as long as they are self-consistent?

stefanv:
Here is what we gave the reviewers in the past, for comparison:

https://github.com/scipy-conference/scipy_proceedings/blob/master/reviews/review-template.rst

katyhuff:
Ah! I missed that. @shjb suggested that in the past the guidance had not
been sufficiently detailed and it caused some confusion for the authors and
reviewers.

I remember this template though, from years past. I liked it.. Let me see
what I can do to combine them where there isn't overlap.

On Tue, May 12, 2015 at 11:15 AM, Stefan van der Walt <
notifications@github.com> wrote:

> Here is what we gave the reviewers in the past, for comparison:
>
>
> https://github.com/scipy-conference/scipy_proceedings/blob/master/reviews/review-template.rst
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/scipy-conference/scipy_proceedings/pull/115#issuecomment-101372487>
> .
>



-- 
http://katyhuff.github.com

review_criteria.md
shjb: Good point! I'm happy with the original wording.

review_criteria.md
shjb: Sorry @katyhuff I somehow missed that on my first reading. Looking at it now, I'm not sure that it takes a clear-enough stance on e.g. an open-source Python wrapper around a closed-source library. Should we try to lay out guidelines on when that's a satisfactory contribution for the proceedings, and when it isn't?

If you or anyone else wants to just leave that question to case-by-case discussions I'm okay with that, but suspect in case-by-case analysis the reviewers will tend to want to avoid [public] confrontation. Therefore I suggest setting the review guideline to err on the strict side of things.

Looking at https://github.com/scipy-conference/scipy_proceedings/blob/master/reviews/review-template.rst (thanks @stefanv and @jarrodmillman !) I see they also address this issue. They take a relatively hard stance against advertising closed-source things.  Relative to that guideline, my inclination would be more moderate. If for example, a wrapper provides an elegant and non-trivial interface to a widely-used closed-source library (e.g. PyOpenCL) then that sort of work should still be welcome in the proceedings regardless of the existence of an open OpenCL implementation.

review_criteria.md
shjb: I was thinking about that too. I'm glad you mention it. Maybe just add it to the list of things SciPy is about? Getting a bit over-wrought maybe but ...

'''
The work should advance the practice of scientific computing, including all subject areas that make up the SciPy community (e.g. high performance computation, analytics, visualization, simulation, numerous specific scientific sub-domains, technical education and the application of Python to social sciences and humanities).
'''

review_criteria.md
katyhuff: I totally totally agree. It hink this should be somewhere in the middle... goldilocks style. There are a lot of reasons to take it on a case-by-case basis. In my own field, for example, a lot of nuclear engineering software is export controlled by DOE... so ... let's err on the stricter side of things, with exceptions on a case by case basis. That way, we end up with a strong preference toward open source, but aren't forcing the reviewers to be super hard-lined if there's  a good reason. I'll think about phrasing.

review_criteria.md
katyhuff: Seems good to me. Will add.

Issue #120, opened 2015-06-11
[proceedings] multiple institutions?


bmcfee:
In the 2015 proceedings template and build system, are multiple institutions per author accepted/supported?

It seems to work, in terms of generating the correct footnotes for affiliations, but I'm getting incorrect footnote markers next to the names.

Example source:
```rst
:author: First author
:email: foo@bar.baz
:institution: Home of Foo
:institution: Home of Bar

:author: Second author
:institution: Home of Bar

:author: Third author
:institution: Yet another place
```
and the rendered tex/pdf output:
![Author names](http://i.imgur.com/LnkG0ls.png)
...
![Footnotes](http://i.imgur.com/he7RBmI.png)

`Second author` should have a single dagger, and only `Third author` should have the double-dagger.

bmcfee:
(clarification) actually the third institution is completely missing, so maybe multi-institution just doesn't work?

stefanv:
This use-case isn't currently supported. Are you interested in making a
pull request, Brian?

bmcfee:
Ok, thanks for clarifying!

I'll take a stab at fixing it, but no promises that it'll work!  (Beats writing, anyway! :))

Is the PR process the same as submission, just with a non-year branch name?

stefanv:
Thanks, Brian! Yes, please make the changes on the master branch, update
the example paper to include the new feature, and then we'll merge the
changes into 2015 as well.

katyhuff:
I think this can be closed now that this is merged. Closing.

PR #121, opened 2015-06-11
added support for multiple institution affiliations per author

bmcfee:
ref #120

This PR changes the author-institution data structure into many-to-many mapping, so that each author can have multiple affiliations without redundant \thanks clauses.

I've only tested this with pdf output on the sample doc (with some modifications to make sure it works).


stefanv:
Thanks, Brian, this looks great.

bmcfee:
Happy to help!

For the record: what's the best way to use this in an existing branch off 2015?  Will merging from upstream master cause troubles down the line when PRs come in?

stefanv:
I've merged your changes onto 2015, so rebasing on 2015 should work

PR #126, opened 2015-06-13
Fixed print statements in build_paper.py to be python 3 compatible

michaelpacer:
Added parens to the print statements so that it would be compatible with both python 3 and python 2.

### DIFF
--- a/publisher/build_paper.py
+++ b/publisher/build_paper.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+from __future__ import print_function
 
 import docutils.core as dc
 import os.path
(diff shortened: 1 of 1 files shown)


stefanv:
To make this compatible with Python 2, we also need

```
from __future__ import print_function
```
at the top of the file.

michaelpacer:
I didn't realize that just by pushing that to my forked branch it would automatically add it to this pull request — but i guess that makes sense. 

But is that correct?

michaelpacer:
Ok didn't realize beginning meant literally beginning.

stefanv:
Thanks, @michaelpacer !

PR #129, opened 2015-06-15
Paper: HoloViews: Building Complex Visualizations Easily for Reproducible Science (Stevens)

jlstevens:
This PR is our Scipy 2015 paper submission regarding our new package called HoloViews.

We would like to be able to state in the paper that both Jean-Luc Stevens and Philipp Rudiger contributed equally to this paper.

We look forward to hearing your comments!

[Edit by jaberg] http://nipy.bic.berkeley.edu:5000/download/27


philippjfr:
Just wanted to mention that we are currently preparing a new release of HoloViews. If you want to try it out for yourself we'd recommend installing v1.2 via pip until v1.2.5 is released. Our Tutorials are available under ``doc/Tutorials`` in the [v1.2 zip file](https://github.com/ioam/holoviews/archive/v1.2.0.zip).

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+underlying floating-point value.  This default visualization may then
+be customized declaratively to achieve the desired aesthetics, without
+complicating the objects themselves with such customizations. This
+separation of content and presentation is already a well established
+design principle outside of science, and is analogous to the
+relationship between HTML content and CSS in web pages.
+
+In the next section we will discuss the data structures that hold the
+important content. Starting with the simple primitive elements, we
+examine how they can be composed into complex figures and embedded
```
prabhuramachandran: Do you mean "embedded in high-dimensional spaces", i.e. the "in" is missing?

jlstevens: It was definitely meant to be "embedded in" as you point out.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+
+The particular ``Image`` shown in Figure :ref:`layout` **A** was 
+constructed as a visualization of the Mandelbrot Set,
+defined in the complex plane. In particular, the ``kdims`` argument
+declares that the *x*-axis is along the real axis and that the
+*y*-axis is along the imaginary axis. This information is then
+reflected in the visual output by assigning
+the appropriate axis labels. This semantic information is also passed
+to the ``Curve`` object generated by sampling the image using
+``im.sample(Im=1)``.
```
prabhuramachandran: A note on what kind of sampling happens when the sample point does not fall exactly on the data would he handy to note.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+highly flexible, compositional tree-based data structure: intermediate
+nodes correspond either to ``Layout`` nodes (``+``) or ``Overlay``
+nodes (``*``), with element primitives at the leaf nodes. Even in this
+potentially complex tree, all the raw data corresponding to every
+visual element is conveniently accessible via key or attribute access
+by selecting a leaf
+element using its path through the tree, and then inspecting the
+``.data`` attribute, making it simple to declare which part of a
+complex dataset you want to work with at a given time.
+
```
prabhuramachandran: I think a simple code example here would help make this easier for the user to follow.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+solution is to present the data sequentially over time as an
+animation.  A third solution is to provide interactive control,
+allowing the user to reveal further dimensionality by interacting with
+the plots using various widgets.
+
+In HoloViews, we solve this problem with composable data structures
+that embed homogeneous collections of ``Element`` objects in any
+arbitrarily dimensioned space. Fundamentally, this set of data
+structures (subclasses of ``NdMapping``) are multi-dimensional
+dictionaries that allow the user to declare the dimensionality of the
```
prabhuramachandran: It isn't too clear what you mean by a "multi-dimensional dictionary"?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+space via a list of key dimensions (``kdims``). The multi-dimensional
+location of the items held by the dictionary are defined by tuples,
+where the values in the tuple match the declared key dimensions by
+position. In addition to regular Python dictionary indexing semantics,
+these data structures also support slicing semantics to select
+precisely the subregion of the continuous multi-dimensional space that
+the user wants to explore.
```
prabhuramachandran: This is a bit too dense and a nice example would serve to clarify this a *lot*.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+   various other domain specific applications such as the visualization
+   of Bloch Spheres in quantum physics. We have found the basic design
+   general and extensible, making it possible to wrap third party
+   libraries with little effort.
+  
+Conclusion
+----------
+
+Based on the key principles of:
+(1) making data immediately and transparently visualizable,
```
prabhuramachandran: This doesn't render as a list if that is what you intended.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+the options can easily be shared across groups of plots, it is simple
+to build up new, highly customized plot types, without necessarily
+requiring custom code.  Third, HoloViews makes it trivial to add
+completely novel types of Elements with corresponding plots (or to
+override specific code in existing plots) using custom code when
+needed, and these custom plots will then combine seamlessly with other
+HoloViews objects to make composite figures.  Thus in HoloViews,
+default plots are simple and straightforward, but it can handle even
+the most complex figures possible with other packages, and should
+offer important benefits at every level of complexity.
```
prabhuramachandran: I think that it would help if you were to present a few links to fairly complex examples in the documentation or code that highlight these points -- you can pick each of them. For example one example for point 1 which shows the easy customizations of settings, and perhaps showing the sharing of the options.  An example demonstrating how easy it is to add new elements would also be helpful as users can see how easy it really is.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+files.  (2) Because HoloViews can directly express the complex
+relationships between different bits of data as subfigures, it can
+capture entire figures within notebooks that would previously have
+required unreproducible work in external drawing programs. (3)
+HoloViews exports the actual data alongside published figures,
+allowing it to be tested automatically (as is done for the HoloViews
+web site) without conflating it with arbitrary display choices.
+HoloViews makes it possible to reproduce results from every step of
+the project, up to and including the final published figures, in a way
+that has not previously been practical.
```
prabhuramachandran: It may help emphasize that holoviews is not just a plotting package but a lot more by mentioning that holoviews can have different backends for the actual plotting (I know this is mentioned elsewhere but it may be nice to emphasize this again in the discussion).

prabhuramachandran:
This is a very well written paper and the HoloViews package is very interesting.  It feels like a breath of fresh air in comparison to other available options!  I have marked a few minor comments that are relatively easy to address.  I think the authors should link to the online documentation of holoviews in the paper as the documentation is very handy and showcases the package well.  The paper itself will also benefit with a little more in the way of code examples (I have indicated a few places where this will help the discussion).  It would also be good to mention explicitly that it is easy to use the package without IPython.

jlstevens:
Thanks for the review!

We'll get working on those fixes shortly but first I would like announce the release of HoloVIews 1.3.0 on [PyPI](https://pypi.python.org/pypi/holoviews/) and on [anaconda](https://anaconda.org/ioam). Our website [holoviews.org](holoviews.org) has also been updated to present material relevant to version 1.3.0.

This release allows HoloViews to use any rendering backend subpackage placed in ``holoviews.plotting``. For instance, we can now have a bokeh backend so you could use HoloViews without matplotlib installed.

In addition to the changes above, we want to update the ``OptionTree`` repr shown in figure 4 as this has also been improved in 1.3.0. It should now be:

```python
options = hv.Store.custom_options()[content.id]
print repr(options)

OptionTree(groups=['plot', 'style', 'norm'],
   plot={
            'Curve ' : dict(aspect=1.25),
            'Layout' : dict(aspect_weight=1, sublabel_format='{Roman}')},

   style={
            'Curve' : dict(linewidth=3),
            'Image' : dict(cmap='Blues')}
)
```

jlstevens:
I've gone and updated the included notebook as suggested above. I would note that ``mandelbrot.npy`` required by this notebook is not included in this PR. Would you like us to include it? 

This file is 1.3MB uncompressed but  far smaller when compressed, including when compressed by git. If you don't want us to add this file to the PR, you can find it in [this directory](https://github.com/ioam/holoviews/tree/master/doc).

We now need to update ``customization_example.svg`` and ``customization_example.pdf``.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+representation. This artificial distinction is a lingering consequence
+of a text-only world and has resulted in a strict split between how we
+conceptualize 'simple' and 'complex' data. Simple data returned from
+an operation is rendered directly in the interpreter (as strings,
+short lists, etc.), but more complex data requires an extremely
+detailed set of explicit steps using external plotting packages such
+as Matplotlib [Mpl]_ before results can be revealed in a graphical
+representation.
+
+Here we introduce HoloViews, a library of simple classes designed to
```
jlstevens: Maybe here we could point people to [holoviews.org](http://holoviews.org)? In addition we may want to mention that the latest version is 1.3.0 at the time of writing.

prabhuramachandran:
@jlstevens could you please make the minor changes I have suggested in my review?

jlstevens:
Yes sure, we were planning to wait for the feedback from the second reviewer before reworking the document (which I believe was due today!).

amueller:
Sorry for the late review, I'll do it today / tonight.
First remark: I think you should contrast / compare with vistrails which tries to solve the same problem but takes a very different approach.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+tools such as the IPython Notebook [Per07]_ and SAGE
+MathCloud [Ste05]_. These projects offer interactive computation sessions
+in a notebook format instead of a traditional text prompt. Although
+similar in design to the traditional text-only interpreters, these
+notebooks allow embedded graphics or other media (such as video) while
+maintaining a permanent record of useful commands in a rich document
+that supports the gradual development of a document with interleaved
+code, results, and exposition. 
+
+Yet despite the greatly improved interactive capabilities of these
```
amueller: While I agree with the statement and sentiment, I feel you could give IPython Notebook more credit, as it has a built-in mechanism for registering html-based visualization. Pandas and scikit-learn already use this mechanism to display tables using html or directly plotting images as images.

amueller: I'd be happy if you say that ipython notebook gives the opportunity to extend the interactive capabilities beyond text based representations, but that this has not really been embraced, and methods for visualizing complex data are rare.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+selects the second key dimension (the imaginary axis ``'Im'``) from
+*0.85* to *1.05*. You can apply a similar slice along the real axis to
+select a portion of the curve object shown in Figure :ref:`layout`
+**B**.
+
+There are many additional element classes, one for each of the 
+common visual representations for data. These elements form an
+extensible library of primitives that allow the composition of data
+structures with complex, meaningful visualizations. Within the set of
+all elements, there are subsets forming equivalence classes according
```
amueller: This sentence is hard to comprehend for non-mathematicians and I'm not sure what you want to express. Can you list the equivalence classes? Can you give an example?

jlstevens:
>  ... I think you should contrast / compare with vistrails which tries to solve the same problem but takes a very different approach.

Sure. I have tried VisTrails before and I did quite like it!

As you point out, the approaches are quite different. With HoloViews, we wanted to integrate well with the IPython Notebook which works more like a traditional interactive session than what I've seen in VisTrails. However, we can mention VisTrails more explicitly in our discussion if you think it appropriate.

amueller:
I think it would be nice to mention it though I don't have very strong feelings. Sorry I didn't finish my review last night, will get back to it now.

philippjfr:
No problem, looking forward to your feedback. Sorry we haven't made Prabhu's suggested corrections yet, we'll start working on them tonight and hopefully complete them before or during our flight to Austin on Friday.

amueller:
Maybe it is my lack of coffee, but I find some parts of the paper hard to follow. I'll leave more detailed in-line comments. I agree with @prabhuramachandran that more code examples might be helpful.

More generally, the paper starts of mostly by motivating immediate representations of complex objects. I feel that is not really picked up much in the discussion of the holo views library.

For the other main theme of separating the concerns of data, plotting and styling, I feel it could be made more explicit how these three work.
So data is represented as elements, styling is represented as options via the %%opts magic (how are these available outside the notebook?), but it is slightly unclear to me how the plotting is represented. There are collections and spaces which both control parts of the plotting. But how do I specify if I want to do a histogram or a line plot?
If I call "hist" (saw it in an example), is the object I get back data (an element) or plot?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+representation, but allows comparison between related data items in
+order to illustrate similarities or differences. In other words, a
+typical figure is an object composed of many visual representations
+combined together. HoloViews makes it trivial to compose elements in
+the two most common ways: concatenating representations side-by-side
+into a single figure, or overlaying visual elements within the same
+set of axes.
+
+These types of composition are so common that both have already been
+used in Figure :ref:`layout` as our very first example. The ``+``
```
amueller: Is that always just one row, so I can't vertically stack plots? Can I do uneven spacing? Is ``a + b + c`` the same as ``a + (b + c)``? If the only purpose of "+" is to do evenly spaces rows of plots, and the rest is handled by ``GridSpaces`` this seems like a very limited use-case.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+dimensionality, we face a tradeoff between representing the full
+dimensionality of our data, and keeping the visual representation
+intelligible and therefore effective. In practice we are limited to
+two or at most three spatial axes, in addition to attributes such as
+the color, angle, and size of the visual elements. To effectively
+explore higher dimensional spaces we therefore have to find other
+solutions.
+
+One way of dealing with this problem is to lay out multiple plots
+spatially.  Some plotting packages [Was14]_ [Wic09]_ have shown how
```
amueller: Maybe mention the packages names seaborn and ggplot2 here as you do in other places.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+this can be done easily using various grid-based layouts. Another
+solution is to present the data sequentially over time as an
+animation.  A third solution is to provide interactive control,
+allowing the user to reveal further dimensionality by interacting with
+the plots using various widgets.
+
+In HoloViews, we solve this problem with composable data structures
+that embed homogeneous collections of ``Element`` objects in any
```
amueller: So I can not have a grid where the first line is a scatter plot, the second line is a line plot and the third line is an image?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+
+One way of dealing with this problem is to lay out multiple plots
+spatially.  Some plotting packages [Was14]_ [Wic09]_ have shown how
+this can be done easily using various grid-based layouts. Another
+solution is to present the data sequentially over time as an
+animation.  A third solution is to provide interactive control,
+allowing the user to reveal further dimensionality by interacting with
+the plots using various widgets.
+
+In HoloViews, we solve this problem with composable data structures
```
amueller: I find this paragraph is really hard to read.

amueller: I'm not sure what you mean by composable data structures. Can a GridSpace contain a GridSpace? Can a HoloMap contain a GridSpace? If so, they don't need to contain homogeneous collections of Element, right?

papers/jean-luc_stevens/jean_luc_stevens.rst
amueller: I guess dictionaries where the keys are tuples?

papers/jean-luc_stevens/jean_luc_stevens.rst
amueller: +10

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+   two dimensions?  If you cast a 3d HoloMap into a GridSpace, won't
+   it just end up being a 2D grid of 1D HoloMaps?  If so you don't
+   need to mention anything here about such a restriction.
+
+All of the above classes are simply different ways to package and view a
+high-dimensional dataset. Just as with ``Elements``, it is possible to
+cast between these different spaces via the constructor (although
+``GridSpace`` is restricted to a maximum of two dimensions). In
+addition, they can all be tabularized into a HoloViews ``Table``
+element or a pandas ``DataFrame`` [pandas]_, a feature that is also supported by
```
amueller: the pandas part here renders oddly.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+location of the items held by the dictionary are defined by tuples,
+where the values in the tuple match the declared key dimensions by
+position. In addition to regular Python dictionary indexing semantics,
+these data structures also support slicing semantics to select
+precisely the subregion of the continuous multi-dimensional space that
+the user wants to explore.
+
+The full list of currently supported ``NdMapping`` classes includes:
+
+* ``HoloMaps``: The most flexible high-dimensional data structure in
```
amueller: How is this most flexible? Videos allow only one additional dimension to vary along, right?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+  arbitrarily high-dimensional space, to be rendered either as a
+  video animation or as an interactive plot that allows
+  exploration via a set of widgets.
+
+* ``GridSpaces``: A data structure for generating spatial layouts
+  with either a single row (1D) or a two-dimensional grid. Each
+  overall grid axis corresponds to a key dimension.
+
+* ``NdLayouts``/``NdOverlays``: Similar to ``Layout`` or ``Overlay``
+  objects, but allowing only homogeneous types of elements, which lets
```
amueller: I don't understand this.

amueller: Or maybe I do? So you give it an object with an extra dimension and it will do "+" along the axis you specify? Maybe consider reformulating it.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+   selected using the widget at the right.  In **B**, only a single
+   curve is shown, with the three sliders at the right together
+   selecting the appropriate curve from the 3D HoloMap space.
+   HoloMaps allow users to explore arbitrarily high-dimensional data
+   naturally and conveniently, with sliders appearing automatically
+   whenever there is more data than can fit into the available display
+   dimensions.  :label:`spaces`
+
+We can immediately go ahead and display this HoloMap either as an
+animation or using the default widgets, as in Figure
```
amueller: maybe try to get the figure on this page.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+:ref:`spaces` **B**. Visualizing individual curves
+in isolation is not very useful, of course; instead we probably want
+to see how the curves vary across ``Frequency`` and ``Amplitude`` in a
+single plot. A ``GridSpace`` provides such a representation and by
+using the space conversion method ``.grid()`` we can easily transform
+our three-dimensional HoloMap into a two-dimensional GridSpace (which
+then allows the remaining dimension, the choice of trigonometric
+function, to be varied via the drop-down menu). Finally, after
```
amueller: Does the drop-down menu control the function for A and B? That is unclear to me. Otherwise I don't know where the function dimension for A went, though. How is decided which plots share a widget?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+   the two are separate yet linked.  :label:`customization`
+
+To explore how option setting works in practice, Figure
+:ref:`customization` shows an example of customizing Figure
+:ref:`layout` with some basic display options.  Here we use an
+optional but highly succinct method for setting the options, an
+IPython cell magic ``%%opts``, to specify aspect ratios, line widths,
+colormaps, and sublabel formats. By printing the string representation
+of the content (``Out[6]``) and the options (``Out[7]``), we can see
+immediately that there are two distinct objects, where each entry in
```
amueller: I feel the start of the sentence unnecessary. Why should they be the same object?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+the options tree matches an applicable object type. Finally, in the
+actual rendered output, we can see that all these display options have
+taken effect, even though the actual data structure differs from the
+object rendered in Figure :ref:`layout` only in the ``id`` value.
```
amueller: Is this visible or clear from any of the code / outputs?
Also, does the user ever have to worry about the ``id``? Is the ``id`` specific to an object? Also, why isn't the ``custom_options`` dict simply indexed by the object itself instead of the ``id``?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+accessible via a tab-completable IPython magic and are documented via
+the ``help`` function. This ease of discovery enables a
+workflow where the visualization details of a plot can be easily and
+quickly iteratively refined once the user has found data of interest.
+
+The options system is also inherently extendable.  New options may be
+added at any time, and will immediately become available for
+tab-completion. In fact, the plotting code for each element and
+container type may be switched out completely and independently, and
+the options system will automatically reflect the changes in the
```
amueller: I am not sure I understand this paragraph. So if the options system automatically reflects the changes, what happens with options that became unavailable? Are they simply ignored?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+available customization options. This approach lets the user work with
+a variety of plotting backends at the same time, without even having
+to worry about the different plotting APIs.
+
+This three-part design explicitly supports the workflows that are
```
amueller: Maybe mention the three parts again explicitly

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+common in science, repeatedly switching between phases of exploration
+and periods of writing up.  Interesting data can be collected and
+curated over time, instantly and transparently visualizable without
```
amueller: This phrasing seems odd.

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+   :align: center
+
+   This view of the HoloViews display and customization systems
+   illustrates the complete separation between the content (data) to
+   be displayed, the display options, and the rendering/plotting
+   system. The display options are stored entirely separately from the
+   content, with the appropriate options being selected with
+   user-controllable levels of specificity: general options for all
+   objects of a given type, more specific options controlled by
+   user-definable ``group`` and ``label`` strings, or arbitrarily
```
amueller: Why is the "group" text in Options centered right? Is that on purpose?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+This paper demonstrates a succinct, flexible, and interactive approach
+for data exploration, analysis, and visualization.  By eliminating the
+need to write plotting code for individual plots, this approach
+extends the quick and easy interaction typical when working with an
+interpreter to include even complex data. At the same time, it allows
+the user to keep a concise and reproducible recipe of their work, from
+exploration to the final publication. HoloViews thus allows scientists
+to capture the entire workflow involved in a research project.
+
+Previous workflows for producing figures from data required an
```
amueller: I feel this was already eliminated by using ipython notebooks and is not necessarily addressed by holoviews. You still need pandas for preprocessing, right?

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+data exploration.  While each of these packages can provide extremely
+polished interactive graphics, getting them set up for specific sets
+of data requires significant additional effort and custom code,
+placing a barrier to their primary use case, the interactive
+exploration of data.  HoloViews instead tries to avoid custom coding
+altogether as far as possible, with users instead supplying metadata
+to declare the properties of the data and option settings to control
+its visual appearance.
+
+Because much of the actual work in creating scientific visualizations
```
amueller: As someone new to holovis I am not sure I understand the concern and maybe phrase it differently.
Maybe rather say "even though we try to be very out-of-the-boxy you can do very custom things, look at these colorful examples"

papers/jean-luc_stevens/jean_luc_stevens.rst
```diff
@@ -0,0 +1,924 @@
...
+needed, and these custom plots will then combine seamlessly with other
+HoloViews objects to make composite figures.  Thus in HoloViews,
+default plots are simple and straightforward, but it can handle even
+the most complex figures possible with other packages, and should
+offer important benefits at every level of complexity.
+
+Although HoloViews is a general purpose library for working with data
+at every stage, it actually represents a significant advance over
+previous approaches focused only on achieving reproducibility of the
+final result. (1) Simply by keeping specifications for figures
```
amueller: The sentence preceding the (1) doesn't make it clear that a list is following, and I'm not sure what your are enumerating.

amueller:
Looking at more examples, I now think this is a pretty cool library. I feel the paper could sell it a bit better, though.
I haven't really understood how exactly the separation of data and plotting works, and it would be great to have some example just showing this off, say showing how easy it is to plug new data into a custom visualization. Maybe just refer to the website?

What I feel is a pretty killer feature are the automatic widgets combined with the slicing.
I haven't really studied the examples on the website in detail, but what I think would be awesome to see would be reading a complex (say five attribute mixted date type) csv in pandas, and then slicing it in different ways, creating holo maps and grids along different axes.

jaberg:
Reminder to everyone that our official review deadline was June 30. Please give the authors some time to make revisions before SciPy.

amueller:
Sorry for the late review. I would not expect the authors to address the possibly larger comments about the three part architecture, but maybe clarify some of the parts pointed out in the inline comments.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Vertically stacking plots is as easy ``(a + b + c).cols(3)``. Controlling the spacing of the plots is a bit more complex as it depends on the plotting backend. Using matplotlib you can specify the aspects of each plot via the plot options and tell the ``Layout`` to what extent it should respect the aspects of each subplot with, [this section](http://ioam.github.io/holoviews/Tutorials/Containers#Layout) of the Tutorials makes this a bit clearer. We're also considering allowing explicit nesting of Layouts to support more complex plot arrangements. 

We'll try to expand a little bit on the fact that even complex, multi-row, arrangements of plots are possible.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Will do.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Currently this is possible but requires you to wrap each Element in an ``Overlay``. We've already got a prototype of another ``Grid`` type, which allows heterogeneous Element types, which you can check out [here](http://philippjfr.com/work/work-in-progress/gridmatrix/) (also showcases the new Bokeh backend).

amueller: Makes sense. I was just checking if I understood correctly.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: There is an overall hierarchy in the way objects can be composed, which you can view [here](http://holoviews.org/Tutorials/Composing_Data). Generally these dictionary-like object types have to be uniform in terms of the type of their values. This ensures that the data in these objects remains easily accessible, i.e. you can easily deconstruct a GridSpace of HoloMaps of Curves into a ``Table`` element or dataframe and then visualize it in some other way. It also let's you perform common numeric operations along the various dimensions (or axes) of the data.

This is in contrast to the non-homogenous types, namely ``Layout`` and ``Overlay``, which let you compose any objects together. The only thing missing is the non-homogenous ``Grid`` type shown above.

I don't know what we can do to make this clearer for the paper, the nesting diagram linked to above probably provides the best overview but would take up way to much space in the paper. I think Prabhu's suggestion of linking directly to the documentation may be appropriate here as well.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Clearly need to phrase this better. HoloMaps don't simply render as videos, the default visualization is the multi-dimensional slider/dropdown widget shown in the ``spaces`` figure, which lets you vary each dimension individually.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Will clarify this, but basically yes. It will do the equivalent of ``+`` and ``*`` but keep the dimension information so that for the NdOverlay type it can generate an appropriate legend and the NdLayout type can generate appropriate titles.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Plots displayed in a single cell will always share a widget. It's a fairly complex thing to explain but it always attempts finds the union of all dimensions. In that example the 'Amplitude' and 'Frequency' dimensions are laid out across the ``GridSpace``, but the simple Curve plot on the right represents a single amplitude and frequency. Therefore changing the amplitude or frequency via the widget will only change the Curve plot, while changing the function via the dropdown will control both plots. If there is no overlap at all in the dimensions of plot **A** and **B** it would just treat it as a sequence of frames. Does that make any sense?

papers/jean-luc_stevens/jean_luc_stevens.rst
amueller: The other option is to just leave out things that are hard to understand without more context and give some pointers that there is more. The interaction of different objects and classes is somewhat non-trivial and you don't need to give all detail here. You should make sure that what is explained here can be understood by itself, though.

papers/jean-luc_stevens/jean_luc_stevens.rst
amueller: I think I got that part. This feature is amazing btw ;)

papers/jean-luc_stevens/jean_luc_stevens.rst
amueller: Yes, it makes sense. Maybe you should mention that there will be one widget per cell and that the axis/selection is shared between all plots to which it applies? Maybe I'm just being picky ;) Looking at the screen shot and thinking about it for a bit I expected something like what you explained, but maybe adding a sentence on it wouldn't hurt.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: No the id value is not immediately visible in the output. Each object is assigned an id value appropriately depending on the options applied to it via the ``%%opts`` magic or via the pure python equivalent, the ``__call__`` method. The ``custom_options`` are not indexed by the object itself because multiple objects may share the same id value. Say you're setting the colormap for a GridSpace of Images, both of these snippets:

```python
hv.GridSpace({i:hv.Image(np.random.rand(10,10)) for i in range(3)})(style={'Image': dict(cmap='gnuplot')})
```

```python
%%opts Image (cmap='gnuplot')
hv.GridSpace({i:hv.Image(np.random.rand(10,10)) for i in range(3)})
```
will apply a newly generated id corresponding to the cmap setting to all the Image Elements in the GridSpace so the plotting code can independently look up the correct settings for each Element and generate this output:

image

amueller: I'm not entirely sure what objects have an id. So here the three ``Image``s share an id?
You said earlier " options based on a unique integer ``id`` assigned to each content object" - so if they might be shared, I don't understand how they are assigned.

Even if you are able to explain all these details in the comments here, maybe I will understand them, but the reader will not have these additional explanations. The paper should be self-contained. 
I realize that the library is complex, and it might not be possible to give all the details. It is not necessary to do so. 
As I said in another place, maybe leave things out that are very complex, and rather give an example. The reader can then go to your docs and figure out the details.

After reading this, I feel slightly confused on what ``id`` does, and it kind of feels to me like an implementation detail.

amueller:
Just to make it clear: I am in favor of accepting this paper. I made comments where I think it can be made more accessible, but I realize that it might not be possible to address all of these.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Agreed, it is definitely not worth getting into the complexities of this as part of the paper. The reason why the ``id`` is introduced at all is because it's the single link between the users data inside the Element and container datastructures and the plotting system. I'm not yet sure what's best to do here, I'll discuss it with the other authors and we'll try to come up with a clearer explanation that avoids getting into the complexities.

jlstevens: I also agree that the ``id`` is an implementation detail. We should try to find a way to avoid mentioning it if we can still making the same basic point - very little state is required on each object in order to link it to the associated options.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: This has just been fixed.

papers/jean-luc_stevens/jean_luc_stevens.rst
jlstevens: Yes, I agree! 

This is a fair observation and I certainly wouldn't dispute that other packages have made use of the display hook machinery in the IPython Notebook. Your suggestion is a good way to put it: HoloViews uses the display hooks extensively in order to extend the core idea of display hooks to far more complex, compositional data types.

We'll be reworking this section shortly.

philippjfr: Addressed this in commit 12b73ab.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: In commit 33c2b99 we addressed this by removing point about equivalence classes entirely and framing it in terms of casting Elements with matching numbers of key and value dimensions.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Done.

papers/jean-luc_stevens/jean_luc_stevens.rst
jlstevens: We decided on this inline list style to save space. Thanks for pointing out what could have been an important formatting error!

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: We've replaced this section with a more concrete example of a space, which we hope will be less confusing.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Just added a sentence to the caption explaining this.

papers/jean-luc_stevens/jean_luc_stevens.rst
jlstevens: The ``cols`` method for setting the number of columns in layouts is mentioned in 4d63b7f.

papers/jean-luc_stevens/jean_luc_stevens.rst
jlstevens: Added a concrete example of data access in c81900c

papers/jean-luc_stevens/jean_luc_stevens.rst
jlstevens: We've now removed references to ``id`` except for figure 3 where it is used (a20876c). This should help avoid confusing the user with implementation details while getting the main point across.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: This is a good point and something we've been meaning to address (but haven't had to because we only had one backend). Our proposed solution is mentioned here https://github.com/ioam/holoviews/issues/201, but we won't get into it as part of the paper.

papers/jean-luc_stevens/jean_luc_stevens.rst
philippjfr: Fixed.

jlstevens:
We've finished updating the paper and believe that our fixes address the majority of the points made.

Once again, we thank the reviewers for their time and feedback.

amueller:
Thank you for your revisions. I will be giving two tutorials tomorrow and on Tuesday, so I am not sure I'll have time to have a look.

PR #131, opened 2015-06-15
Paper: The James Webb Space Telescope Data Calibration Pipeline (Bushouse)

hbushouse:
[Edit by jaberg] http://nipy.bic.berkeley.edu:5000/download/25

### DIFF
--- a/papers/howard_bushouse/howard_bushouse.rst
+++ b/papers/howard_bushouse/howard_bushouse.rst
@@ -1,0 +1,379 @@
+:author: Howard Bushouse
+:email: bushouse@stsci.edu
+:institution: Space Telescope Science Institute
+
+:author: Michael Droettboom
+:email: mdroe@stsci.edu
+:institution: Space Telescope Science Institute
+
+:author: Perry Greenfield
+:email: perry@stsci.edu
+:institution: Space Telescope Science Institute
+
+--------------------------------------------------------
+The James Webb Space Telescope Data Calibration Pipeline
+--------------------------------------------------------
+
+.. class:: abstract
+
+   The James Webb Space Telescope (JWST) is the successor to the Hubble Space
+   Telescope (HST) and is currently expected to be launched in late 2018.
+   The Space Telescope Science Institute (STScI) is developing the pipeline
+   systems that will be used to provide routine calibration of the science
+   data received from JWST. The JWST calibration pipelines use a processing
+   environment provided by a Python module called "stpipe" that provides
+   many common services to each calibration step, relieving step developers
+   from having to implement such functionality. The stpipe module provides
+   multi-level logging, command-line option handling, parameter validation
+   and persistence, and I/O management.
+
+   Individual steps are written as Python classes that can be invoked
+   individually from within Python or from the stpipe command line. Pipelines
+   are created as a set of step classes, with stpipe handling the flow of
+   data between steps. The stpipe environment includes the use of standard
+   data models. The data models, defined using json schema, provide a means of
+   validating the correct format of the data files presented to the pipeline,
+   as well as presenting an abstract interface to isolate the calibration
+   steps from details of how the data are stored on disk.
+
+.. class:: keywords
+
+   pipelines, astronomy
+
+Introduction
+------------
+
+Data coming from the electronic detectors in scientific instruments
+attached to telescopes (both on the ground and in space) look nothing like
+the end product on which astronomers do their analysis or the pictures that
+show up in the media. Raw images and spectra contain artifacts and extra
+signals that are intrinsic to the instrumentation itself, rather than the
+source being observed. These artifacts include things like dead detector
+pixels, pixel-to-pixel variations in sensitivity, background signal from
+the detector and instrument, non-linear detector response, anomalous
+signals due to impacts of cosmic-rays, and spatial distortions due to the
+optics.
+The removal of or correction for
+all anomalies must be performed before the data are suitable for scientific analysis.
+In addition, processing such as combining the data from multiple exposures
+and extracting one-dimensional spectra from the two-dimensional
+detector format in which they were recorded must also be performed.
+This is the job of astronomical data reduction and calibration pipelines.
+
+The Space Telescope Science Institute (STScI), which is the science operations
+center for the Hubble Space Telescope (HST), has developed and maintained
+the calibration pipelines for all of the HST scientific instruments and is now
+in the process of developing the pipelines that will be used for the James
+Webb Space Telescope (JWST) after it is launched in late 2018. The HST
+pipelines for the different scientific instruments on the telescope were
+developed over a span of more than 20 years and hence show an evolution in
+both software languages and design. The pipelines for each instrument, which
+now number 11 over the 25 year history of HST, were all written
+independently of one another and used an assortment of programming languages
+ranging from the Subset Pre-Processor (SPP), which is unique to the astronomical
+community, Fortran, C, and eventually Python. This assortment of languages
+and the somewhat monolithic design of each instrument pipeline naturally
+led to increased burdens on maintenance and enhancements, and also made code
+sharing rather difficult.
+
+The JWST calibration pipelines are being developed from scratch using a
+completely new design approach and of course using almost nothing but Python.
+There is a common framework for all 4 of the scientific instruments, with
+extensive sharing of routines and a common code base. The new design allows
+for flexibility in swapping in and out specific processing steps, easily
+changing the ordering of steps within pipelines, and the ability for
+astronomers to plug-in custom processing. This flexibility is necessary due
+to the fact that the knowledge of the science instruments and the intricacies
+of the data they produce is constantly evolving, often over the entire
+lifetime of the mission.
+
+stpipe
+------
+
+The heart - or perhaps more appropriately, the nervous system - of the JWST
+calibration pipeline environment is a Python module called "stpipe". Stpipe
+manages individual processing steps that can be combined into pipelines. The
+stpipe environment provides functionality that is common to all steps and
+pipelines so that they behave in a consistent manner. Stpipe provides:
+
+- running steps and pipelines from the command line
+- parsing of configuration settings
+- composing steps into pipelines
+- file management and data I/O between pipeline steps
+- interface to the Calibration Reference Data System (CRDS)
+- logging
+
+Each pipeline step is embodied as a Python class, with a pipeline being
+composed of multiple steps. Pipelines can in turn be strung together, just
+like steps, to compose an even higher-order flow. Steps and pipelines can
+be executed from the command-line using stpipe, which is the normal mode of
+operations in the production environment that processes data in real-time as
+it is downlinked from the telescope. The step and pipeline classes can also
+be instantiated and executed from within a Python shell, which provides a
+lot of flexibility for developers when testing the code and to astronomers
+who may need to occasionally tweak or otherwise customize the processing of
+their particular data sets.
+
+When run from the command line, stpipe
+handles the parsing of configuration parameters that can be provided either
+as arguments on the command line or within configuration files. Configuration
+files use the well-known ini-file format and stpipe uses the ConfigObj 
+library to parse them.
+stpipe handles all of the file I/O for each step and the passing of data between
+pipeline steps, as well as providing access within each step to a common
+logging facility. It also provides a common interface for all steps to
+reference data files that are stored in the STScI Calibration Reference
+Data System (CRDS). Having all of these functions handled by the stpipe
+environment relieves developers from having to include these features in
+each step or pipeline and provides a consistent interface to users as well.
+
+Command-line Execution
+======================
+
+Stpipe can be used from the command line to execute a step or pipeline by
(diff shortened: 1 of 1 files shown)


papers/howard_bushouse/howard_bushouse.rst
```diff
@@ -0,0 +1,379 @@
...
+now number 11 over the 25 year history of HST, were all written
+independently of one another and used an assortment of programming languages
+ranging from the Subset Pre-Processor (SPP), which is unique to the astronomical
+community, Fortran, C, and eventually Python. This assortment of languages
+and the somewhat monolithic design of each instrument pipeline naturally
+led to increased burdens on maintenance and enhancements, and also made code
+sharing rather difficult.
+
+The JWST calibration pipelines are being developed from scratch using a
+completely new design approach and of course using almost nothing but Python.
```
mrocklin: Suggest removing "of course".  Makes it seem like the other languages don't have value.

mrocklin:
This paper presents a data pipeline abstraction to be used in the JWST.

It is clearly written and has a consistent narrative.  

Suggestions for improvement:

1.  It would be nice to see a non-trivial use case, perhaps with multiple pipelines connected together in a real-world application.  Such an experiment would give this paper more depth.
2.  It would be nice to see a comparison with other data pipeline projects or at least a discussion of background.  Systems like scikit-learn have similar pipelines, Luigi would arguably be a general purpose approach.  There are undoubtedly others.  What distinguishes this work from those?
3.  It would be nice to see discussion of the "big ideas" of data pipelining, and how this work might be generalized outside of JWST.

papers/howard_bushouse/howard_bushouse.rst
```diff
@@ -0,0 +1,379 @@
...
+
+    # setup logging
+    import logging
+    log = logging.getLogger()
+    log.setLevel(logging.DEBUG)
+
+    # the pipeline class
+    class SloperPipeline(Pipeline)
+
+        spec = “””
```
kbarbary: These unicode quote marks aren't rendering well in the PDF, at least in my viewer; replace with `"`.

papers/howard_bushouse/howard_bushouse.rst
```diff
@@ -0,0 +1,379 @@
...
+--------------------------------------------------------
+
+.. class:: abstract
+
+   The James Webb Space Telescope (JWST) is the successor to the Hubble Space
+   Telescope (HST) and is currently expected to be launched in late 2018.
+   The Space Telescope Science Institute (STScI) is developing the pipeline
+   systems that will be used to provide routine calibration of the science
+   data received from JWST. The JWST calibration pipelines use a processing
+   environment provided by a Python module called "stpipe" that provides
```
kbarbary: You could use the rst inline-code markup
```rst
``stpipe``
```
throughout when refering to the Python module or executable.

papers/howard_bushouse/howard_bushouse.rst
```diff
@@ -0,0 +1,379 @@
...
+
+    # the pipeline class
+    class SloperPipeline(Pipeline)
+
+        spec = “””
+            save_cal = boolean(default=False)
+        “””
+
+        # step definitions
+        step_defs = {"dq" : dq_step.DQInitStep,
```
kbarbary: The relation between these `step_defs` and instance methods of the same names wasn't entirely clear to me. Presumably the class constructor looks at `step_defs` and creates attributes for each one? It might be worth a few sentences explaining this and the benefit of this approach.

papers/howard_bushouse/howard_bushouse.rst
```diff
@@ -0,0 +1,379 @@
...
+   The James Webb Space Telescope (JWST) is the successor to the Hubble Space
+   Telescope (HST) and is currently expected to be launched in late 2018.
+   The Space Telescope Science Institute (STScI) is developing the pipeline
+   systems that will be used to provide routine calibration of the science
+   data received from JWST. The JWST calibration pipelines use a processing
+   environment provided by a Python module called "stpipe" that provides
+   many common services to each calibration step, relieving step developers
+   from having to implement such functionality. The stpipe module provides
+   multi-level logging, command-line option handling, parameter validation
+   and persistence, and I/O management.
```
kbarbary: I suggest replacing "multi-level logging, command-line option handling"  with "common configuration handling" or some such. (My first reaction to this was "don't `logging` and `argparse` already provide this?" :)

kbarbary:
Overall, I agree with Matt's comments: This is a clearly written paper, but could be improved by more discussion of the broader context.

jakevdp:
I agree with much of @mrocklin's evaluation. A couple particulars:

- A comparison with other approaches would be nice. I found myself wondering if the previous 11 (?) attempts at a data reduction pipeline tried similar things. Is it mainly the language that's outdated, or are they non-modular, or something else?

- I'd like to see an example of a more realistic use-case which strings together multiple steps.

- The current paper ends pretty abruptly; some sort of conclusion statement should be added before merge.

papers/howard_bushouse/howard_bushouse.rst
```diff
@@ -0,0 +1,379 @@
...
+center for the Hubble Space Telescope (HST), has developed and maintained
+the calibration pipelines for all of the HST scientific instruments and is now
+in the process of developing the pipelines that will be used for the James
+Webb Space Telescope (JWST) after it is launched in late 2018. The HST
+pipelines for the different scientific instruments on the telescope were
+developed over a span of more than 20 years and hence show an evolution in
+both software languages and design. The pipelines for each instrument, which
+now number 11 over the 25 year history of HST, were all written
+independently of one another and used an assortment of programming languages
+ranging from the Subset Pre-Processor (SPP), which is unique to the astronomical
```
jeanconn: I think if you "range from" something you need to range "to" something else.  It might be good to just replace the "ranging" with "including" or the like.

papers/howard_bushouse/howard_bushouse.rst
```diff
@@ -0,0 +1,379 @@
...
+the end product on which astronomers do their analysis or the pictures that
+show up in the media. Raw images and spectra contain artifacts and extra
+signals that are intrinsic to the instrumentation itself, rather than the
+source being observed. These artifacts include things like dead detector
+pixels, pixel-to-pixel variations in sensitivity, background signal from
+the detector and instrument, non-linear detector response, anomalous
+signals due to impacts of cosmic-rays, and spatial distortions due to the
+optics.
+The removal of or correction for
+all anomalies must be performed before the data are suitable for scientific analysis.
```
jeanconn: Maybe turn this sentence around for a bit more readability? as is "All anomalies must be removed or corrected before the data are suitable for scientific analysis" or "Processing steps must remove or correct all anomalies..."

jeanconn:
I also agree with much of the comments of @mrocklin , @kbarbary , and @jakevdp ; this paper has a clear and consistent narrative and seems to describe the tool well.  I also agree that it could be improved with a conclusion and perhaps some more context / description of the bigger picture.  

I note that the introduction talks about a primary driver in this development being flexibility, but for me the example given only allows the reader to infer flexibility from the interface.  If the more realistic / non-trivial use case suggested by @jakevdp and @mrocklin could also include a less subtle demonstration of the flexibility gained from the new structure, that would be great!

taldcroft:
Yes, one more person independently having similar comments.  I enjoyed reading the paper and it would definitely be improved after addressing all the comments.  Some others:

- It is implied that users can run most or all of the production data processing pipelines, and potentially tweak them.  If that is really the case then that will be a great accomplishment and should be highlighted.  I wonder about requirements for having the CRDS available, system configuration, dependencies and so on.  This places a non-trivial burden on developers, like providing usable documentation.  :smile:  I know that the Chandra data processing pipelines require a specialized environment, with dedicated and knowledgable staff to run them.

- Is there any concept of step dependencies?  For instance you might have a long processing pipeline that creates a sequence of files.  If you improve/tweak a step near the end and want to reprocess everything there is no need to re-do earlier steps.

- The paper mentions being able to stop at an arbitrary point (in "Steps and Pipelines"), but it isn't clear how this is accomplished.  Maybe this is related to the `step_defs` class attribute?  Does this require the user to be aware of whether the pipeline is a simple linear pipe with no logic?

- In the example `sloper.cfg` file the output file is hardcoded as `output.fits`.  This seems in conflict with the idea of a file-format-independent data model.  It might be worth mentioning how or if the file naming is abstracted out.

- In real-world pipelines there might be hundreds of files for each run or observation so I'm wondering if there is some framework for defining and maintaining file naming conventions.

eteq:
@hbushouse - I have read over this submission and have little to add to the many existing comments, except for one minor item: it's not clear to me that all readers will actually know what astronomers mean when they say "pipeline". Perhaps just a single sentence in the intro clarifying this (in particular that a pipeline is *not* necessarily just linear, as you mention later on)

papers/howard_bushouse/howard_bushouse.rst
hbushouse: Done. Removed in next version.

papers/howard_bushouse/howard_bushouse.rst
hbushouse: Fixed in the next version.

papers/howard_bushouse/howard_bushouse.rst
hbushouse: Changed all occurrences in next version.

papers/howard_bushouse/howard_bushouse.rst
hbushouse: @mrocklin For the next version I've added a section to the end of the introduction that better clarifies what exactly we mean in this context by the term "pipeline". We are not talking about a high-level task and workflow management system like Luigi, but rather a lower-level pipeline that simply applies a series of operations of imaging data. From what I can tell the scikit-learn pipelines are strictly for the purpose of applying a series of transforms to collections of data values, which is not at all what we are doing here. Hopefully the revised text will make this more obvious to readers.

papers/howard_bushouse/howard_bushouse.rst
hbushouse: The purpose and functionality of the step_defs is now explained in the paragraph preceding the code example.

papers/howard_bushouse/howard_bushouse.rst
hbushouse: This suggestion is being incorporated into the next version of the paper.

papers/howard_bushouse/howard_bushouse.rst
hbushouse: This wording has been fixed in the updated version of the paper.

papers/howard_bushouse/howard_bushouse.rst
hbushouse: Done. Wording fixed in next version.

hbushouse:
@taldcroft - I have expanded the discussion in the introduction to indicate explicitly that we do plan to distribute the calibration pipeline package to users so that they can run it on their home systems, much in the same way that we currently distribute the HST pipelines via IRAF/STSDAS. The package is relatively self-contained and hence should not be difficult for users to install and run on their own. The only dependency (other than common python libraries) is the CRDS, which I've now explained works via a server/client system between STScI and the user's system.

I've also added a bit of discussion to the data models section to indicate how step dependencies are handled, when they exist, and how the use of data models themselves eliminate many dependencies.

I've expanded the linear vs. non-linear pipeline discussion a bit to make it more explicit that only linear pipelines can be started/stopped mid-stream and that those options don't even exist for a user when running a non-linear pipeline.

I've updated the discussion of sloper.cfg to try to make it more obvious that this is an example of a case where the user is overriding the default behavior to specify a desired output file name and also briefly mention how stpipe handles the default situation when the output name is not specified. We're still working on defining all the file name conventions for the various products that will be generated by the different observing modes. So I can't address that point yet!

hbushouse:
@eteq - I have expanded the introduction to explain in more detail just what kind of pipeline we're talking about in this context. We're not talking about a big, high-level process management system. We're just feeding the pixel data from a given image through a series of operations.

papers/howard_bushouse/howard_bushouse.rst
hbushouse: Oh, and I've also added another example that shows a pipeline being used from within another pipeline (it's still a fairly trivial case, but ...)

papers/howard_bushouse/howard_bushouse.rst
hbushouse: I'm also adding a conclusion section that summarizes and addresses several of the high-level issues.

hbushouse:
I've committed an updated version of the paper that incorporates as many of the review comments as feasible. The new version tries to do a better job of explaining how the use of the term "pipeline" in this context does not refer to a high-level task and process management system, but rather a much more light-weight middle-ware system that guides data in a fairly simple way through a set of processing steps. Hence I did not think it was necessary to devote time and effort to a discussion of existing high-level process management systems and how our pipelines relate to them (or why we didn't use them), because that's not the type of processing that we need here. Regardless, it is a much improved paper (IMHO) thanks to all of the suggestions and comments received. Thanks!

taldcroft:
:+1:

PR #132, opened 2015-06-15
Paper: Dask: Parallel Computation with Blocked algorithms and Task Scheduling (Rocklin)

mrocklin:
[Edit by jaberg] http://nipy.bic.berkeley.edu:5000/download/24

PR #135, opened 2015-06-15
Paper: Creating a Real-Time Recommendation Engine using Modified K-Means Clustering and Remote Sensing Signature Matching Algorithms (Lippa & Vertrees)

dalippa:
Added introduction information, as well as methods/implementation. Added working source code that can be copied and pasted into iPython.

[Edit by jaberg] http://nipy.bic.berkeley.edu:5000/download/22


mrocklin:
This paper describes the solution of a clustering problem on Google App Engine.

Suggestions for improvement:

1.  The first line of the abstract sounds like a sales pitch for a company, not like the beginning of a technical document.
2.  On a first reading it is not clear to me on what features we are clustering.  There is significant discussion of spectral imaging (indeed, one of the two figures is on the spectrum of deciduous trees) but the application seems to be in real estate and discusses very different features.
3.  There is an interesting problem here, notably "how can we do k-means clustering on the restrictive environment of GAE."  This could use more depth and discussion.
3.  Details of the solution are sparse, possibly requiring analysis of the source code attached in the Appendix.  
4.  The paper is very short, consisting of about two pages of prose.

dalippa:
Hi Matthew,

1. I didn't know if we could change the abstract as I had previously submitted. I will definitely submit a new abstract.
2. I I'll add some clarification on what we use for clustering and why. I think editing the abstract will help with this. The idea in the introduction was to show that there is some similarity between a multi-spectral signature match (like Landsat, which has 7 bands including RGB; http://landsat.usgs.gov/band_designations_landsat_satellites.php) and a recommendation match in a way that is coherent for those who do not have a remote sensing background.
3. I'll add some additional paragraphs to discuss the GAE restrictions in more detail and how we implemented clustering on GAE.
4. The details of the clustering solution are indeed sparse. I can go into some more detail in the paper on this so that it can stand alone without the source code. Do you agree that the snippet of SAM code earlier in the paper is sufficient?
5. The paper is pretty short, in terms of prose; I looked at past proceedings (most notably Google's App Engine SciPy paper here: http://conference.scipy.org/proceedings/scipy2011/pdfs/starnes.pdf) and some were as short as 3 pages. Most of the breakthrough in solving the problem was in its transformation, and less the actual solution. In some ways, we came up with the solution backwards, from the search engine (the most limiting component) up.

mrocklin:
5.  Being short isn't a vice; concision is a virtue.  However brevity is a common sign of "not enough content to warrant a significant contribution."  It's quite likely that, in skimming, I didn't glean the breakthrough here.  It may require some more effort to frame that contribution and demonstrate its worth in a way that is obvious to the casual reader.  It didn't come across to my time-pressured eye.

dalippa:
The breakthrough was less technical, and more about how the restrictions forced us to transform a problem to overcome GAE's weaknesses (complex real-time data processing). The SAM algorithm required a concept of a signature, which required clustering and the KDTree, etc.

We were nearly hogtied with the available tools to solve the problem, first by the lack of traditional libraries and techniques, and next by the limitations of GAE search.

dalippa:
@mrocklin Hi Matthew,

I've made some substantial edits, following your suggestions. Please re-review and provide another round of feedback.

I removed the clustering source code completely; if you think it should be added back, please let me know.

dalippa:
@mrocklin @katyhuff Since this work was done while employed, my employer, RealMassive, owns the copyright. Is there a way to change the template for our paper so that Jason and I are still shown as authors, but RealMassive is marked as the copyright holder?

katyhuff:
@dalippa Gotcha. We'll have to change the template (I just made issue #169). If no one else solves it, I will try to get to it before the actual publication date (omg two weeks from now).  

You're welcome to give it a shot and submit a PR!! The offending line of python is, more or less, here: https://github.com/scipy-conference/scipy_proceedings/blob/master/publisher/writer/__init__.py#L183).

katyhuff:
(note that it should be simple, we should add something like an optional :copyrightholder: item to the header variables and use that instead of the authors, if it exists.

dalippa:
@katyhuff I'll give it a shot to fix.

dalippa:
@katyhuff I have a fix and will be submitting a PR shortly. Do you want me to submit it against 2015 or against master?

dalippa:
Katy,

That's exactly what I did :)

--David

On Thu, Jun 25, 2015 at 12:43 PM, Katy Huff <notifications@github.com>
wrote:

> (note that it should be simple, we should add something like an optional
> :copyrightholder: item to the header variables and use that instead of the
> authors, if it exists.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/scipy-conference/scipy_proceedings/pull/135#issuecomment-115341879>
> .
>



-- 
David Lippa
Principal Software Engineer
david.lippa@realmassive.com
RealMassive <http://www.realmassive.com/>
t: @realmassive


* Commercial Real Estate in Real-Time*

*-----*
This message contains confidential information and is intended only for the
individual named. If you are not the named addressee you should not
disseminate, distribute or copy this e-mail. Please notify the sender
immediately by e-mail if you have received this e-mail by mistake and
delete this e-mail from your system. E-mail transmission cannot be
guaranteed to be secure or error-free as information could be intercepted,
corrupted, lost, destroyed, arrive late or incomplete, or contain viruses.
The sender therefore does not accept liability for any errors or omissions
in the contents of this message, which arise as a result of e-mail
transmission. If verification is required please request a hard-copy
version.

mrocklin:
Hi @dalippa, substantial edits indeed.  To my eye this is a big improvement and your method and the motivations behind it are now much clearer to me.

If I might ask for a little more, how does this method perform?  Can you compare it both in terms of performance and accuracy to an alternative, perhaps doing this on a single machine.  Can you provide some motivating evidence that this is the right approach for this application?

dalippa:
@mrocklin Thanks.

The method I described is incredibly fast on a local machine. When SAM is used on a 30 MP spectral image, even a lightly outfitted desktop computer can process it in almost a second. However, we have very little to no control when it comes to deployment hardware.

As for performance, there are two ways the service runs:
1. On *every* page view of a space (under "similar spaces", such as here: https://www.realmassive.com/us/tx/austin/office-space/oak-creek-plaza/101/4531509935472640), using the currently viewed space as the user-input (no clustering needed).
2. And on the survey view, which requires the clustering step before providing a recommendation.

In practice, it runs quite fast, though we have a very limited choice of hardware. We're running on a 2.4 Ghz F4_1G GAE instance (https://cloud.google.com/appengine/docs/python/modules/) with only 1 GB of RAM available (and each instance serves about a dozen or more requests per second). The technique we used had to work on a front-end instance (for auto-scaling), use a minimum of RAM, could not load data on-demand (due to latency), had to operate in real time, and was limited by the mathematical functions described in the paper. Given the implementation hardware and time constraints (3 weeks from design to production) and our real-time needs, I feel we made the right choice.

From my benchmarking analysis, it takes about 50 ms for GAE search to fetch a single document (representing an entire object and its children) by its hash ID; fetching 100 items via a query takes roughly 200-400 ms (without the angle calculation); we use 20 as the default number of items returned in the recommendation list, with a performance range of 6 ms (cached results) - 850 ms (complex query). When I did stress testing (100+ recommendations from 100+ user inputs - well outside the parameters for what it's used), performance lagged quite a bit (6 ms (cached) - 1950 ms), but we were returning whole objects rather than the necessary fields. A projection query would bring response time back to sub-second speeds.

Based on your comments, I probably should add a note that future work could involve parallelizing the clustering algorithm similarly to how you suggest in your paper #132.

mrocklin:
I think that it would be very interesting to include these comments in the paper itself.  This might give weight to the method.

While I appreciate the potential nod to my work it seems somewhat unrelated at the moment.

dalippa:
@mrocklin I'm going to add a "results" section similar to the poster to integrate some of these comments into the paper. I'll also add some statistics over the past 10 months.

dalippa:
@mrocklin Ready for re-review. I added a results section that had the time constraints, performance considerations, and a better breakdown of the performance cost of each step. I think it works well as a bookend to the initial statement in the introduction that the hardware was limiting.

I also did a new benchmark against the desktop hardware that is more apples-to-apples, executing the spectral data cube using memory-mapped file I/O. Subsecond speeds on the 30 MP dataset are expected when fully loaded into RAM.

I also pulled the event stats to show average and peak throughput, which goes to the point of scale.

Please let me know if there are any additional changes, or if it's ready to merge.

Thanks!

dalippa:
P.S. When it's ready to merge, I will remove the poster file from the directory. I just wanted to keep it in the same place as the paper for simplicity's sake.

dalippa:
@mrocklin @katyhuff @jaberg Is the paper ready for acceptance? I have not heard back since my last round of edits submitted ~ 4 days ago.

papers/david_lippa/david_lippa.rst
```diff
@@ -0,0 +1,160 @@
...
...elevance, but by a modified version of a spectral angle |---| a simple computation borrowed from the domain of linear algebra and spectral analysis. The Spectral Angle Mapper (SAM) algorithm treats each pixel of an image as an *n*-dimensional vector :math:`\vec{v_{ij}}` and computes the angle :math:`\theta` between :math:`\vec{v_{ij}}` and a vector :math:`\vec{s}` for all rows *i* and all columns *j*: :math:`\cos^{-1}\left(\frac{\vec{s} \cdot \vec{v_{ij}}}{\left|\vec{s}\right| \left|\vec{v_{ij}}\right|} \right)`. A potential candidate match usually has an angle between 5 and 10 degrees, while a collinear match has an angle of 0. For remote sensing applications, SAM has a roughly 83% accuracy rate when predicting exact signature matches [Pet11]_ in a variety of applications and domains including: determining the chemical composition of stars [Ric15]_, analyzing the health of vegetation [Zha09]_, measuring the quality of an RGB image, and detecting camouflage in times of war [Lan05]_. Unfortunately, one of the weaknesses of the SAM algorithm is that a collinear match can show up as a false positive [#]_, requiring additional algorithmic steps that takes vector magnitude into account.
```
katyhuff: For readability, the equation in the middle of this paragraph might deserve its own line.  That is, I'd recommend:

```
...The Spectral Angle Mapper (SAM) algorithm treats each pixel of an image as 
an *n*-dimensional vector :math:`\vec{v_{ij}}` and computes the angle 
:math:`\theta` between :math:`\vec{v_{ij}}` and a vector :math:`\vec{s}` for all rows 
*i* and all columns *j*: 

..math::
  :type: equation
  :label: sam

  \cos^{-1}\left(\frac{\vec{s} \cdot \vec{v_{ij}}}{\left|\vec{s}\right|\left|\vec{v_{ij}}\right|}\right).

A potential candidate match usually has an angle between 5 and 10 degrees, 
while a collinear match has an angle of 0....
```

papers/david_lippa/david_lippa.rst
```diff
@@ -0,0 +1,160 @@
...
...ly, one divides the *n* data points into an initial set of clusters, whose new center points are calculated as an unweighted average, causing the clusters to shift in response to the distribution of their contents. Iteratively, clusters are merged and center points are re-calculated until no cluster intersects any other [Vat11]_. We chose not to use an iterative approach which has a worst-case complexity of :math:`2^{\Omega(n)}` even in 2 dimensions [Vat11]_. Though the worst case scenario doesn't seem to arise in practice, we use a quick guess-and-check method that has good asymptotic complexity and converges quickly, even though another algorithm may produce better results. The algorithm takes advantage of a few known attributes of the data: there is a limited amount of overlap between data points because they represent physical objects in 3-dimensional space; the data points have a limited range since they are latitude and longitude coordinates; and since we use the clusters as a geofence in our search parameters, using a global KDTree of all building coordinates in our datastore allows us to make a good estimation of the initial cluster sizes. The algorithm executes as follows:
```
katyhuff: In the middle of this paragraph, where you say "another algorithm" are you referring specifically to the iterative approach that you chose to avoid, or a broader set of other more-precise algorithms that you are aware of? If the former, consider saying "even though the iterative approach". If the latter, consider saying "even though other algorithms". If something else, maybe clarify?

papers/david_lippa/david_lippa.rst
```diff
@@ -0,0 +1,160 @@
...
..., it has generated more than 302,925 recommendations, sifting through over 80,000 spaces at sub-second speeds. The workload averaged in the thousands per work day, with loads peaking at 18,327 per day and 1,407 per minute [#]_. These speeds were reached when deployed as part of the RealMassive API [#]_ on F4_1G instances, each outfitted with a 2.4 Ghz proecssor, 1 GB RAM, and configured with automatic scaling on a per-request basis [Goo15]_. Benchmarks of the GAE search service showed that search queries limited to 100 results clocked in between 6-600 |nbsp| ms depending on caching and query complexity. The clustering and SAM algorithm added up to 200 |nbsp| ms, mostly in the form of reads from the datastore prior to building clusters. At 6-800 |nbsp| ms, GAE performed nearly 8 times slower than consumer hardware [#]_ but scaled well during traffic spikes. Recently, we performed a stress test outside of a standard use case: 80+ recommendations generated from 100+ user inputs lagged at an unacceptable 3 seconds |---| a performance hit caused by returning entire objects rather than utilizing a projection query, an optimization that would lower response time back to sub-second speeds.
```
katyhuff: It looks to me like this section satisfies @mrocklin's desire to know more about performance of the algorithm. I might recommend, if it makes sense, to present some of this data in a table, as it becomes somewhat dense to read inline.

papers/david_lippa/david_lippa.rst
```diff
@@ -0,0 +1,160 @@
...
+                        [(4, 3, 2), (0, 1, 1)],
+                        [(1, 1, 1), (4, 4, 1)]])
+    >>> SAM(img, sig)
+    """
+    matches = []
+    sig_norm = sig/np.linalg.norm(sig)
+    for r in range(len(img)):
+      for c in range(len(img[r])):
+        pix = img[r][c]
+        cos_theta = pix.dot(sig_norm)/np.linalg.norm(pix)
```
katyhuff: This line of code slightly overflows its bounds. I think it's acceptable as is, but if it can be ameliorated with a linebreak, all the better.

katyhuff:
@dalippa Thanks for a nice, short, readable discussion of a practical improvement to a constrained challenge in the commercial sector. I recommend to accept. A few minor comments are noted inline.


## Written Quality

The prose successfully communicated a narrative without extensive reliance on jargon. It clearly describes a challenge, the constraints of the solution, the approach to the challenge, and notes potential alternative solution paths that were not selected. The written style is clear and error free, with excellent communicative grammar. Where possible, I would recommend utilizing the space available by presenting a the denser pieces of data in a table or two and presenting the key equation in the display equation environment, to help draw the eye of the reader.

## Technical Content

The technical content is sound and presentsed clearly. It describes the use of python to solve a challenge of interest to conference attendees who may also be using the GAE or implementing recommender systems in other capability-constrained webframeworks.

## Novelty and Impact

The implementation of the spectral angle method seems to be an appropriate approach and represents creative problem solving on the part of the author.

## Verifiability

Software descriptions accompany descriptions of the algorithm. While the code is presumably not open source, the essential kernel of information is shared and described in sufficient detail for reproduction. 

## Other Requirements

ok

### Length

This paper complies with our page limit.

### Figures

A caption accompanies the figure, which is of a readable nature. 

### Code Snippets

With the exeption of one line of code that overflows its boundaries slightly, the code snippets are formatted to fit and are presented within PEP8 compliance.

papers/david_lippa/david_lippa.rst
dalippa: @katyhuff It's fairly simple, given that it's the dot product. My inclination would be to leave it in-line.

papers/david_lippa/david_lippa.rst
dalippa: @katyhuff The standard K-means clustering algorithm produces better results, but has the worst case scenario of 2^n. I'll make the change "even though other algorithms". TODO.

papers/david_lippa/david_lippa.rst
dalippa: I'll look into putting it into a table. TODO.

papers/david_lippa/david_lippa.rst
dalippa: @katyhuff I think I can get the desired result with "cos_t" instead of "cos_theta". TODO.

dalippa:
@katyhuff I will implement the changes for some of the recommendations (marked as TODO) and re-push.

papers/david_lippa/david_lippa.rst
dalippa: @katyhuff I just tried to put the results into a table, and it disrupted the entire flow of the results section. I would like to keep it as is, if at all possible.

dalippa:
@katyhuff @jaberg Ready for re-review.

papers/david_lippa/david_lippa.rst
katyhuff: no problem. leaving it as-is will do fine. Thanks for considering.

katyhuff:
@dalippa , my issues are resolved. I'll change the label to "ready" assuming that @mrocklin's concerns have also been addressed, as I looked it over and they seem to have been addressed as far as I can tell. @mrocklin : if you have a chance over the next few days, please double check and let us know if any of the recommendations you made were not adequately addressed.

Thanks!

mrocklin:
Sorry for the prolonged absence.  Recommend accept.

PR #136, opened 2015-06-15
Paper: Automated Image Quality Monitoring with IQMon (Walawender)

joshwalawender:
Add paper from Josh Walawender


papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+.. class:: keywords
+
+   astronomy, automated telescopes, image quality
+
+.. [#] Source code at ``https://github.com/joshwalawender/IQMon``
+
+Introduction
+------------
+
+Using existing tools such as astropy [astropy]_, astrometry.net, source extractor, SCAMP, and SWARP, IQMon analyzes images and provides the user a quick way to determine whether the telescope is performing at the required level.
```
mdboom: "provides the user **with** a quick way"

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+IQMon can provide provide a determination of whether the telescope is focused (from the typical FWHM of stars in the image), whether it is pointing accurately (obtained from a comparison of the target coordinates with the astrometrically solved coordinates), whether the tracking or guiding is adequate (from the typical ellipticity of stars in the image), and whether the night is photometric (obtained from the typical photometric zero point of stars in the image).  For wide field systems which detect many stars in each image, these metrics can be spatially resolved allowing for more detailed analysis such as differentiating between tracking error, focus error, and optical aberration or determining if the dome is partially obscuring the telescope aperture.
```
mdboom: "provide ~~provide~~"

mdboom: Provide definition of FHWM

papers/josh_walawender/josh_walawender.rst
mdboom: References for source extractor, SCAMP and SWARP.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+Because the system is designed to do quick evaluations of image quality, the primary concept is an object representing a **single** image.  IQMon does not do any image stacking or other processing which would be applied to more than one image at a time nor is it built around other organizational concepts such as targets or visits.  It is not intended to supplant a full data reduction and analysis package.  The output of IQMon, however, can be stored in a MongoDB[#]_ database making it potentially useful for collecting information on observing concepts which span multiple images such as targets, nights, or visits.  It might also be useful as a preprocessing step for a more complex data pipeline.
+
+.. [#] http://www.mongodb.org
+
+To date, IQMon has been deployed on three disparate systems: a 735mm focal length wide field imager with a monochrome CCD camera which undersamples the PSF, an 0.5 meter f/8 telescope with a monochrome CCD camera with well sampled PSF, and an 85mm focal length camera lens and DSLR camera (with Bayer color array) designed for very wide field photometry.  IQMon has provided valuable diagnostic information about system performance in all cases.
```
mdboom: Define PSF (point spread function) for the first instance.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+
+Structure and Example Use
+-------------------------
+
+IQMon operates by using ``Telescope`` and ``Image`` classes.  The ``Telescope`` object contains basic information about the telescope which took the data.  When a ``Telescope`` object is instantiated, a configuration file is read which  contains information on the telescope and controls various user configurable parameters and preferences for IQMon.  The configuration file is a YAML document and is read using the ``yaml`` module.
```
mdboom: "user-configurable"

Reference for "PyYAML" package.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+An ``Image`` object is instantiated with a path to a file with one of the supported image formats and with a reference to a ``Telescope`` object.  The image analysis process is simply a series of calls to methods on the ``Image`` object.
+
+The IQMon philosophy is to never operate on the raw file itself, but instead to create a "working file" (using the ``read_image`` method) and store it in a temporary directory.  If the raw image file is a FITS file, then ``read_image``  simply copies the raw file to the temporary directory and records this file name and path in the ``working_file`` property.  If the file is a raw image file from a DSLR (e.g. ``.CR2`` or ``.dng`` format), then ``read_image`` will call ``dcraw`` using the subprocess32 module to convert the file to ``.ppm``.  The file is then converted to FITS format using either ``pamtofits`` or ``pnmtofits`` tools from the ``netpbm`` package.  To date IQMon has only been tested with FITS and ``.CR2`` files, but should in principle work with numerous DSLR raw format images.
```
mdboom: Reference for "dcraw".  Reference "subprocess32" -- is it importantly different from the stdlib "subprocess" module?  If not, maybe just say "subprocess module (from the Python standard library)".  Reference for "netpbm".

"To date, IQMon"

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+    # Read the fits header
+    im.read_header()
+
+Once the image has been read in and a working file created, IQMon uses various third party tools to perform image analysis.  The following sections describe some of the analysis steps which are available.
+
+
+PSF Size Measurements with Source Extractor
+```````````````````````````````````````````
+
+Source Extractor (SExtractor; [Bertin1996]_[Bertin2010_SExtractor]_) is called using the ``run_SExtractor`` method which invokes the command using the subprocess32 module.  Customization parameters can be passed to Source Extractor using the telescope configuration file.
```
mdboom: Same comment about `subprocess` from above.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+    ## and can be manipulated as such.  For example:
+    print('Image FWHM = {:.1f}'.format(im.FWHM))
+    print('Image FWHM = {:.1f}'.format(\
+          im.FWHM.to(u.arcsec, equivalencies=\
+          im.tel.pixel_scale_equivalency)))
+    print('Zero Point = {:.2f}'.format(im.zero_point))
+    print('Pointing Error = {:.1f}'.format(\
+          im.pointing_error.to(u.arcmin)))
+
+These results can also be stored for later use.  Methods exist to write them to an ``astropy.Table`` (the ``add_summary_entry`` method) and to a YAML document (the ``add_yaml_entry`` method), but the preferred storage solution is to use a mongo database.
```
mdboom: Just for additional clarity: Why is a mongo database preferred, and why is it preferred to another database system or type?

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+.. figure:: PSFplot.png
+   :scale: 65%
+   :figclass: w
+
+   An example of the plot which can be produced by the ``determine_FWHM`` method.  The plot shows histograms of the FWHM and ellipticity values (upper left and upper right respectively), the spatial distribution of FWHM and ellipticity values (middle left and middle right), ellipticity vs. radius within the image (lower left), and the correlation between the measured PSF position angle and the position angle of the star within the image (lower right). :label:`PSFplot`
```
mdboom: "produced ~~by~~ using"

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+In the example plot (Fig. :ref:`PSFplot`), we can see several different effects.  First, from the spatial distribution of FWHM and ellipticity, as well as the ellipticity vs. radius plot, we see that image quality is falling off at large radius.  This image is from a wide field imaging system and we are seeing the signature of off axis aberrations.  This is also suggested in the plot of the correlation between the measured PSF position angle and the position angle of the star within the image which shows strong diagonal components indicating that position within the image influences the PSF.  There is also, however, a vertical component in that plot at :math:`PA \sim 0` which is suggestive of image drift perhaps due to slight polar misalignment or flexure.
```
mdboom: "large ~~radius~~ radii"

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+JPEGs and Plots
+---------------
+
+In addition to generating single values for FWHM, ellipticity, and zero point to represent the image, IQMon can also generate more detailed plots with additional information.
+
+A plot with PSF quality information can be generated when ``determine_FWHM`` is called by setting the ``plot=True`` keyword.  This generates a .png file (see Fig. :ref:`PSFplot`) using matplotlib which shows detailed information about the point spread function (FWHM and ellipticity metrics) including histograms of individual values, a spatial map of FWHM and ellipticity over the image, and plots showing the ellipticity vs. radius within the image (which can be used to show whether off axis aberrations influence the ellipticity measure) and the correlation between the measured PSF position angle and the position angle of the star within the image (which can be used to differentiate between tracking error and off axis aberrations).
```
mdboom: Unless the fact that it's a PNG is meaningful, maybe just say "image" (and there are some other instances of this below).

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+
+Flags
+`````
+
+For the four primary measurements (FWHM, ellipticity, pointing error, and zero point), the configuration file may contain a threshold value.  If the measured value exceeds the threshold (or is below the threshold in the case of zero point), then the image is "flagged" as an indication that there may be a potential problem with the data.  The flags property of an ``Image`` object stores a dictionary with the flag name and a boolean value as the dictionary elements.
+
+This can be useful when summarizing results.  For example, the Tornado web application provided with IQMon (see the `Tornado Web Application`_ section) lists images and will color code a field red if that field is flagged.  In this way, a user can easily see when and where problems might have occurred.
+
+JPEGs and Plots
```
mdboom: This section header says "JPEGs", but "PNGs" are mentioned below.  Since the file format isn't really important to the discussion maybe just say "images"

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,251 @@
...
+
+This web application is intended to be the primary interface for users.  It provides three levels of interaction to the user.  First, a custom plot of IQMon results over the course of a night is easy to generate from the mongo database entries and represents the highest level of interaction.  Using such a plot, serious problems which affect many images can be detected at a glance.  Users can then drill down to see a list of images for that UT date and see system performance as a table of IQMon results with flagged values highlighted in red.  Finally an individual image can be examined as a jpeg with overlays or by using the PSF quality plots or zero point plots to examine detailed performance.
+
+References
```
mdboom: We should proabably have a formal Conclusion section here.  Maybe something describing how IQMon has impacted the productivity of the installations where it is used?

joshwalawender:
@mdboom Thanks for the comments!  I've incorporated them and updated the paper.

eteq:
@joshwalawender - It looks like this paper is not currently building (see http://nipy.bic.berkeley.edu:5000/ ) - are you currently working on it or are you unaware of the failing build?

katyhuff:
@joshwalawender : you can see the build error if you click on the words "build failed" in the bot. 

```
...
Underfull \hbox (badness 1454) in paragraph at lines 194--195
\T1/ptm/m/n/10 Using ex-ist-ing tools such as as-tropy [[]], as-trom-e-try.net
Package hyperref Warning: Ignoring empty anchor on input line 200.
Runaway argument?
{
! Paragraph ended before \hyperlink was complete.

\par
l.220 ...iteral start-string without end-string.}}
! ==> Fatal error occurred, no output PDF file produced!
Transcript written on paper.log.
...
```

joshwalawender:
@eteq @katyhuff Sorry I missed the build error.  I think I have it fixed.  Some of the reference formatting looks weird to me, but I'm not sure what it is supposed to look like.

katyhuff:
> I think I have it fixed. 

It appears so!

>  Some of the reference formatting looks weird to me

If you want to be more specific, I can help you determine whether it's intentional.

joshwalawender:
@katyhuff In the first paragraph, the pdf output looks like:

"Using existing tools such as astropy [astropy], astrometry.net [Lang2010], ..."

The references being links in brackets following the statement looks odd to me, is that the intended formatting?

katyhuff:
You, as the author, are in complete control of the reference text that appears in those brackets. I see two reasonable options:

"Using existing tools such as [astropy], astrometry.net [Lang2010], ..."

or

"Using existing tools such as Astropy [AstropyCollaboration2013], astrometry.net [Lang2010], ..."

For the latter, you would need to:

1) Change the text in the in-line reference to `[AstropyCollaboration2013]_`
2) and change the text in the brackets at the bottom of the doc to `.. [AstropyCollaboration2013]  Astropy Collaboration ", Robitaille, T.~P., Tollerud, E.~J., et al. *Astropy: A community Python package for astronomy* 2013..."`

joshwalawender:
Thanks for the tip.  I've reformatted the references to be a bit clearer (and made some other minor fixes) and have pushed the latest version.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+
+.. [#] http://www.mongodb.org
+
+To date, IQMon has been deployed on three disparate optical systems: a 735mm focal length wide field imager with a monochrome CCD camera which undersamples the point spread function (PSF), an 0.5 meter f/8 telescope with a monochrome CCD camera with well sampled PSF, and an 85mm focal length camera lens and DSLR camera (with Bayer color array) designed for very wide field photometry.  IQMon has provided valuable diagnostic information about system performance in all cases.
```
eteq: It might be useful to specify where/what the purpose of these applications was?  That helps the reader better understand the motivation and suitable use cases.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+
+.. [#] http://pyyaml.org
+
+An ``Image`` object is instantiated with a path to a file with one of the supported image formats and with a reference to a ``Telescope`` object.  The image analysis process is simply a series of calls to methods on the ``Image`` object.
+
+The IQMon philosophy is to never operate on the raw file itself, but instead to create a "working file" (using the ``read_image`` method) and store it in a temporary directory.  If the raw image file is a FITS file, then ``read_image``  simply copies the raw file to the temporary directory and records this file name and path in the ``working_file`` property.  If the file is a raw image file from a DSLR (e.g. ``.CR2`` or ``.dng`` format), then ``read_image`` will call ``dcraw`` [#]_ using the subprocess32 module [#]_ to convert the file to ``.ppm``.  The file is then converted to FITS format using either ``pamtofits`` or ``pnmtofits`` tools from the ``netpbm`` [#]_ package.  To date, IQMon has only been tested with FITS and ``.CR2`` files, but should in principle work with numerous DSLR raw format images.
```
eteq: I imagine a certain amount of customization is necessary for different kinds of DSLR cameras.  Perhaps there should be brief mention of this? (Or at least how much work was necessary to get it to work with the peculiarities of the one DSLR it's been used with thus far)

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+
+.. [#] http://www.cybercom.net/~dcoffin/dcraw/
+
+.. [#] The subprocess32 module "is a backport of the subprocess standard library module from Python 3.2 & 3.3 for use on Python 2.4, 2.5, 2.6 and 2.7" (from https://pypi.python.org/pypi/subprocess32).  It is used instead of the standard subprocess module due to its support for timeout functionality.
+
+.. [#] http://netpbm.sourceforge.net
+
+IQMon has been tested with Python 2.7.X, testing with Python 3.X is pending.  Python 3.X compatibility notes will be posted to the readme file on the git repository.
```
eteq: Are you planning to use `2to3` to manage compatibility?  Probably worth mentioning it if so (or using it if you aren't already!)

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+
+In the following sections, I will describe a simple example of evaluating image quality for a single image.  A more complex example which is updated in concert with IQMon can be found in the ``measure_image.py`` script at the git repository for the VYSOS project [#]_.  That process can then be wrapped in a simple program to monitor a directory for images and analyze them as they are written to disk (see the ``watch_directory.py`` script in the same VYSOS repository for an example).  This enables automatic near real time analysis.
```
eteq: Do you use some other tool to monitor the directory for new files?  Might be worth mentioning (I say because I've found it can be non-trivial in a platform-independent manner...)

papers/josh_walawender/josh_walawender.rst
eteq: Also might be useful to mention what platform it's being tested/supported on (I'm guessing linux?).  Occassionally CCD or small telescope control software is windows-only, so some readers may actually care.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+
+.. [#] http://www.mongodb.org
+
+To date, IQMon has been deployed on three disparate optical systems: a 735mm focal length wide field imager with a monochrome CCD camera which undersamples the point spread function (PSF), an 0.5 meter f/8 telescope with a monochrome CCD camera with well sampled PSF, and an 85mm focal length camera lens and DSLR camera (with Bayer color array) designed for very wide field photometry.  IQMon has provided valuable diagnostic information about system performance in all cases.
+
+
+Structure and Example Use
+-------------------------
+
+IQMon operates by using ``Telescope`` and ``Image`` classes.  The ``Telescope`` object contains basic information about the telescope which took the data.  When a ``Telescope`` object is instantiated, a configuration file is read which  contains information on the telescope and controls various user-configurable parameters and preferences for IQMon.  The configuration file is a YAML document and is read using the pyyaml [#]_ module.
```
eteq: Is it also possible to do this programtically without a YAML file?  Might be worth mentioning that if so.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+``````````````````````````````````````
+
+After importing IQMon, the first step would be to instantiate the ``Telescope`` object which takes a configuration file as its input.  The next step is to instantiate an ``Image`` object with the path to the image file and the ``Telescope`` object representing the telescope which took that image.
+
+.. code-block:: python
+
+    tel = IQMon.Telescope('~/MyTelescope.yaml')
+    im = IQMon.Image('~/MyImage.fits', tel)
+
+IQMon writes a log which is intended to provide useful information to the user (not just the developer) and shows the progress of the analysis.  We can either pass in a ``logger`` object from Python's logging module, or ask IQMon to create one:
```
eteq: Good trick!  Creating loggers can be an awkward business, so I like that you've made it easier while still being compatible.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+    # Read the fits header
+    im.read_header()
+
+Once the image has been read in and a working file created, IQMon uses various third party tools to perform image analysis.  The following sections describe some of the analysis steps which are available.
+
+
+PSF Size Measurements with Source Extractor
+```````````````````````````````````````````
+
+Source Extractor (SExtractor) is called using the ``run_SExtractor`` method which invokes the command using the subprocess32 module.  Customization parameters can be passed to Source Extractor using the telescope configuration file.
```
eteq: Should cite SExtractor paper here.  Also definitely need a sentence or two explaining what it *is*.  You probably shouldn't assume all readers know that (although of course many will).

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+Determining the PSF size from the SExtractor results is done with the ``determine_FWHM`` method.  The full width at half maximum (FWHM) and ellipticity values for the image are a weighted average of the FWHM and ellipticity values for the individual stars.
+
+These steps not only provide the typical FWHM (which can indicate if the image is in focus), they can also be used to guess at whether the image is "blank" (i.e. very few stars are visible either because of cloud cover or other system failure).  For example:
+
+.. code-block:: python
+
+    im.run_SExtractor()
+    # Consider the image to be blank if <10 stars
+    if im.n_stars_SExtracted < 10:
+        im.logger.warning('Only {} stars found.\
```
eteq: I think this `\` is supposed to be a `'`?

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+    else:
+        im.determine_FWHM()
+
+
+Pointing Determination and Pointing Error
+`````````````````````````````````````````
+
+IQMon also contains a ``solve_astrometry`` method to invoke the ``solve-field`` command which is part of the astrometry.net software.  The call to ``solve-field`` is only intended to determine basic pointing and orientation and so IQMon does not use the SIP polynomial fit of distortion in the image.
+
+Once a world coordinate system (WCS) is present in the image header, then the ``determine_pointing_error`` method can be called which compares the right ascension (RA) and declination (DEC) values read from the RA and DEC keywords in the header (which are presumed to be the telescope's intended pointing) to the RA and DEC values calculated for the center pixel using the WCS.  The separation between the two coordinates is reported as the pointing error.
```
eteq: perhaps you should note that you're using `astropy.wcs` for the WCS parsing and processing?

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+    if not im.image_WCS:
+        im.solve_astrometry()
+    # Determine pointing error by comparing telescope
+    # pointing coordinates from the header with WCS.
+    im.determine_pointing_error()
+
+Astrometric Distortion Correction
+`````````````````````````````````
+
+In order to make an accurate comparison of the photometry of stars detected in the image and stars present in a chosen stellar catalog, many optical systems require distortion coefficients to be fitted as part of the astrometric solution.  IQMon uses the SCAMP software to fit distortions.
```
eteq: I find it a bit curious that you use SCAMP for distortion correction even though you mention above that  astrometry.net gives SIP coefficeints.  Perhaps a sentence or two motivating SCAMP by saying why astrometry.net is not suitable?

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+        if im.SCAMP_successful:
+            # Remap the pixels to a rectilinear grid
+            im.run_SWarp()
+
+Estimating the Photometric Zero Point
+`````````````````````````````````````
+
+With a full astrometric solution, SExtractor photometry, and a catalog of stellar magnitude values, we can estimate the zero point for the image and use that as an indicator of clouds or other aperture obscurations.
+
+The ``get_catalog`` method can be used to download a catalog of stars from VizieR using the astroquery module.  Alternatively, support for a local copy of the UCAC4 catalog is available using the ``get_local_UCAC4`` method.
```
eteq: Perhaps note that `get_catalog` uses the `astroquery` package? (I don't think there's a paper to cite, but mentioning the package/linking to the web site is probably good enough)

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+
+It should be noted that unless custom code is added to handle reduction steps such as dark/bias subtraction and flat fielding, the zero point result will be influenced by systematics due to those effects.  In addition, the choice of catalog and the relative response curve of the filter in use and the filter defined by the catalog's photometric system will also introduce systematic offsets.  For many systems (especially typical visible light CCDs), the zero point value from IQMon can be used to compare throughput from image to image, but should not be used to compare different equipment configurations.
+
+Analysis Results and Mongo Database Integration
+-----------------------------------------------
+
+Results of the IQMon measurements for each image are stored as properties of the ``Image`` object as ``astropy.units.Quantities``.  For example, the ```FWHM`` value is in units of pixels, but can be converted to arcseconds using the equivalency which is automatically defined by the ``Telescope`` object (``tel.pixel_scale_equivalency``) for this purpose.
```
eteq: (minor pedantic point:) it's `astropy.units.Quantity` - I usually do this as `astropy.units.Quantity`'s or something like that.

Also, I think there's an extra backtick in FWHM you didn't mean to be there.

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+
+.. code-block:: python
+
+    ## Results are typically astropy.units quantities
+    ## and can be manipulated as such.  For example:
+    print('Image FWHM = {:.1f}'.format(im.FWHM))
+    print('Image FWHM = {:.1f}'.format(\
+          im.FWHM.to(u.arcsec, equivalencies=\
```
eteq: This is a very nice use of equivalencies!

papers/josh_walawender/josh_walawender.rst
```diff
@@ -0,0 +1,271 @@
...
+
+Images and Plots
+----------------
+
+In addition to generating single values for FWHM, ellipticity, and zero point to represent the image, IQMon can also generate more detailed plots with additional information.
+
+A plot with PSF quality information can be generated when ``determine_FWHM`` is called by setting the ``plot=True`` keyword.  This generates a .png file (see Fig. :ref:`PSFplot`) using matplotlib which shows detailed information about the point spread function (FWHM and ellipticity metrics) including histograms of individual values, a spatial map of FWHM and ellipticity over the image, and plots showing the ellipticity vs. radius within the image (which can be used to show whether off axis aberrations influence the ellipticity measure) and the correlation between the measured PSF position angle and the position angle of the star within the image (which can be used to differentiate between tracking error and off axis aberrations).
```
eteq: Should probably cite matplotlib (http://matplotlib.org/citing.html)

papers/josh_walawender/josh_walawender.rst
eteq: Oh, and perhaps here or later also the use of `astropy.coordinates`? (I see that in the source code)

eteq:
@joshwalawender - I have now reviewed this contribution.  It is clearly organized, reasonably clearly written, and describes well a python tool of clear use for astronomy.  I find this to be an *excellent* example of a "hybrid" project, where Python is used both to glue together non-python tools, and is used on its own in situations where it makes sense.  I've left a variety of inline comments, tending towards the stylistic (which you can consider optional if you strongly disagree).  Below I've included some more general comments/suggestions.

1. The introduction is a bit light on *motivating* the tool.  As an astronomer who has some contact with people who run automated telescopes, I see the need, but other readers might be less familiar.  So something about how automated telescopes are becoming more common in astronomy and their data volumes are large (and quick-responding) enough to need automated tools.  A short paragraph or even a few sentences in the first paragraph should be sufficient.
2. Despite this project using a lot of non-python tools, there's little mention of how the "plumbing" works.  That is, how are these tools invoked?  A cursory glance at the source code suggests you mostly use `subprocess`... But this should be mentioned in the paper, ideally with any experience you gained while developing it about why you picked that approach (if you have any).
3. Along similar lines, I imagine the install process is non-trivial because all of the other various sub-tools need to be installed and accessible from the python process.  If there are some non-trivial aspects to this or clever tricks you've used for this, you may want to mention them.
4. I saw little mention of how quickly any of this runs (beyond some references to "near real-time").  That's probably quite important to some users, so I suggest at least a sentence or two somewhere that says if it takes seconds, minutes, hours, etc for your three use cases. 
5. I found the section on the tornado application interesting but a bit terse.  That's understandable given the situation, but I think because of that it would be valuable to show it in action.  E.g., a screenshot of it in use, or a link to it in actual operation, etc.
6. I left a few inline comments suggesting you mention specific parts of `astropy` that you use.  Of course I'm a bit biased as one of the coordinators of `astropy`, so feel free to take these as suggestions and ignore them if you think they disrupt the flow or something.  I just think it might be useful for readers to see where these pieces of astropy fit into a very practical use case like this one.

(One final item that is not strictly about the paper but I figure I should mentiopn: have you considered passing back some of the code putting your tables into mongoDB  to `astropy.table`?  Not sure how much of it is of general use, but it seems like it might be!)

papers/josh_walawender/josh_walawender.rst
joshwalawender: At the moment, no, it requires a YAML file, but it would be a relatively simple tweak to allow that.  I'll add it to my to do list.

papers/josh_walawender/josh_walawender.rst
joshwalawender: I've added text to clarify the customization needed.

papers/josh_walawender/josh_walawender.rst
joshwalawender: Not sure yet, how I'll handle the compatibility.  Past use has been on systems using python2, but will be installing and running this for a project which has software based on python3.  I'd hoped to sort all this out before SciPy, but we didn't quite get to this step on that project yet.

I've added text indicating what platforms this has been used on.

papers/josh_walawender/josh_walawender.rst
joshwalawender: I don't use any particularly clever directory monitoring program.  In the past I've had trouble finding something that works reliably, so for this script I just did a relatively simple brute force loop which is specific to that directory structure and platform rather than build something general.

There's a distinction I'm trying to make here, which I'm not sure is clear in the text:  IQMon is a package which contains the methods for image quality analysis, but does not contain a fully formed script for doing so.  Making a general purpose script seemed too difficult a problem, so instead of trying to build one, I encourage people to look at the scripts which are customized to the VYSOS project and build their own.

papers/josh_walawender/josh_walawender.rst
joshwalawender: Good point.  Done.

papers/josh_walawender/josh_walawender.rst
joshwalawender: Done.

papers/josh_walawender/josh_walawender.rst
joshwalawender: Use of both ```astropy.wcs``` and ```astropy.coordinates``` are now included in the text here.

papers/josh_walawender/josh_walawender.rst
joshwalawender: Yeah, I debated how much detail to go in to here and ended up leaving it out.  In short, this choice was based on my own knowledge (really a lack thereof) of how to get a reliable and timely solve from astrometry.net.

In my initial explorations, I was not getting reliable distortion results from astrometry.net without the possibility of the solve-field operation "hanging" and delaying subsequent processing for very long periods (many minutes).  Since my initial test system was an older machine and was handling two telescopes, it was not able to keep up with the data rate if this was the case.  I never found quite the right settings to use to get reliable and timely results.

Using SCAMP allowed me to add an "ahead" file for the telescope which incorporates previous knowledge about the distortions rather than doing a blind solve.  Once the ahead file is tuned to the system, SCAMP was fast enough to keep up with the data flow and more reliable.

Lastly, if I use SCAMP, then SWARP to de-distort the image, I can ignore distortion from then on and interpret the WCS using just the standard WCS header keywords and not worry about whether any particular software I might use could handle the particular distortion keywords used.

I'm not sure how much of this discussion to include.  If a user can get good results from astrometry.net, then they can choose to only use that (with a SIP keyword in the ```solve_astrometry()``` method).  It's actually been on my wish list to revisit this topic as the SWARP pixel modification breaks a relatively minor feature which I originally included that I'd like to add back in.

Do you think a discussion like that above would be useful in the paper?

papers/josh_walawender/josh_walawender.rst
joshwalawender: Astroquery is mentioned, but the text was not formatted properly.  I've fixed that and I hope it addresses this issue.

papers/josh_walawender/josh_walawender.rst
joshwalawender: Fixed.

papers/josh_walawender/josh_walawender.rst
joshwalawender: Yeah, I'd initially had functions to do the conversion in both directions, but once I discovered equivalencies, I was able to replace that with just a definition of the equivalency.

papers/josh_walawender/josh_walawender.rst
joshwalawender: Done.

papers/josh_walawender/josh_walawender.rst
joshwalawender: @eteq Are you referring to the three optical systems mentioned?  If so, I debated going in to more detail on each, but decided against that because I didn't want to get bogged down discussing those projects.  I just wanted to point out that IQMon is general enough to handle three very different imaging systems.  I'd be happy to add such a description (it would take a couple of paragraphs), but I was worried it would be tangential to the main goal of the paper.  What do you think?

joshwalawender:
@eteq I've responded to your inline comments and made corrections to address them.  For your comments 1-6 above:

1. I added a paragraph to Introduction section to have a bit more discussion of motivation.

2. The plumbing is fairly simplistic at the moment.  It uses subprocess as you surmised. I've added a couple of sentences explicitly stating that to the "Structure and Example Use" section.

3. Unfortunately, I don't have a good install process.  That's an item on my to do list for this.  We're currently integrating IQMon in to the control system itself on one of the projects I'm involved in and we will probably look at making an install script of some sort as part of this, but for now it is TBD.

4. I've added a couple of paragraphs discussing analysis time to the "Structure and Example Use" section.

5. I've added a screenshot (Fig. 4) showing the application in action.  A public link is, unfortunately, impractical due to the limited bandwidth to the site.

Regarding, my mongoDB and ``astropy.table``, my code does not use an astropy.table for this step, so it really is just a very generic use of the PyMongo methods.

Thank you very much for the useful comments!  I've pushed a series of commits to the repo.

papers/josh_walawender/josh_walawender.rst
eteq: Ah, gotcha.  I definitely suggest either using `2to3` (easiest because it requires almost no effort and is built into the stdlib), or [six](https://pypi.python.org/pypi/six) (a bit more up-front work and maintenance, but less overhead in the long run), depending on how much you want to do right away vs. later.  

(But that's not important for the paper, just suggestions from my experiences...)

papers/josh_walawender/josh_walawender.rst
eteq: Gotcha - that makes sense, @joshwalawender . Probably worth mentioning this explicitly, then (unless you added that in the latest update and I missed it?)

papers/josh_walawender/josh_walawender.rst
eteq: Ok, cool - but if it's not in right now no need to mention in the paper.

papers/josh_walawender/josh_walawender.rst
eteq: @joshwalawender - yep, definitely I think this is worth including, as it's hard-earned wisdom you gained that might be useful to others who have similar needs.  A perhaps slightly abriged version of what you just said in this comment would do just fine.

papers/josh_walawender/josh_walawender.rst
eteq: Yep, I see it now.  You might add a footnote with a link to the astroquery docs (as there's no paper to cite) for convenience to the reader, but I'm fine with it either way.

papers/josh_walawender/josh_walawender.rst
eteq: @joshwalawender - I see your point here: I think a brief summary of the context is important, but not necessarily morivating the *science* goals beyond just a phrase or so. It might be this better fits further up as part of the intro: something like (based on your mention below of VYSOS): "Robotic telescopes are critical to some modern astronomical efforts like monitoring surveys of highly variable stars (this is the 0,5 meter telescope we mention below)."

Another possibility is to just provide links to the survey web pages or something like that?  Then the interested reader can drill deeper *if they want*, but without distracting from the focus of the paper.

eteq:
@joshwalawender - thanks, this addresses most of my comments.  I have a few follow-ups above, but none of them are critical, so I'll leave it to your discretion whether you want to do anything with them.

So, @katyhuff @jaberg, once @joshwalawender has decided on those, I consider this ready (@mdboom, do you agree?).

joshwalawender:
@eteq @katyhuff @jaberg @mdboom I have pushed additional edits based on the most recent comments and the submission is ready for a final review.  Thanks very much to all of you for the useful feedback.

eteq:
Looks good to me!

PR #149, opened 2015-06-16
Paper: Scientific Data Analysis and Visualization with VTK and ParaView (Quammen)

cquammen:
Due to circumstances beyond my control involving a sick child and traveling spouse, I regret to report that this submitted draft is incomplete. I will be working to complete the paper over the next day or two. I am looking forward to this collaborative review process.

[Edit by jaberg] http://nipy.bic.berkeley.edu:5000/download/9


cquammen:
This draft is now complete and ready for review.

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,871 @@
...
+
+VTK is provided as a single ``vtk`` module that can be imported like
+any other module. When building VTK from source, if Python support is
+enabled, the build process will produce an executable named
+``vtkpython``. This is the standard Python executable with environment
+variables set to make it simple to import the ``vtk`` module. It is
+also possible to use the same ``python`` executable from the Python
+installation against which VTK was built by prepending the location of
+VTK's shared libraries and the location of VTK's ``__init__.py`` file
+to the PYTHONPATH environment variable.
```
prabhuramachandran: FWIW, this is actually not very standard for Python packages but I guess there never has been a good way to work around this.  The shared library linking has also required LD_LIBRARY_PATH additions unless RPATH has been used.  There used to be a `setup.py` at one point but I guess that was removed later.  Building and packaging VTK is not the easiest and this is a bit of a pain for the average Python user.  I guess this is completely tangential to the paper or its review.  It may help a reader to know the easiest ways to install VTK-Python.  Both Canopy and conda do provide VTK builds on all major platforms.

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,871 @@
...
+   vtk_norm_pressure = np.numpy_to_vtk(norm_pressure, 1)
+   vtk_norm_pressure.SetName('normalized pressure')
+   pd.AddArray(vtk_norm_pressure)
+
+The second argument to ``np.numpy_to_vtk`` indicates that the NumPy
+array should be deep copied to the VTK array. This is necessary if no
+reference to the NumPy array will otherwise be kept. If a reference to
+the numpy array will be kept, then the second argument can be omitted
+and the NumPy array will be shallow copied instead, saving memory and
+time because the array data does not need to be copied.
```
prabhuramachandran: It may be good to warn readers that if no reference is kept and the second argument is 0, one can crash the interpreter.

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,871 @@
...
+   VTKArray(847.02)
+
+This example illustrates nicely the power of combining a NumPy-like
+interface with VTK's uniform API for computing various quantities on
+different types of data sets.
+
+Another distinct advantage of the ``numpy_interface.algorithms``
+module is that all operations are supported in parallel when data sets
+are distributed across computational nodes. [Aya14] describes this
+functionality in more detail.
```
prabhuramachandran: This is very nice functionality indeed.

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,871 @@
...
+to produce publication-quality plots.
+
+Second, VTK itself incorporates some of matplotlib's rendering
+capabilities directly when possible. When VTK Python wrapping is
+enabled and matplotlib is available, VTK uses the
+``matplotlib.mathtext`` module to render LaTeX math expressions to
+either ``vtkImageData`` objects that can be displayed as images or to
+paths that may be rendered to a ``vtkContextView`` object, VTK's
+version of a canvas.
+
```
prabhuramachandran: A short example may help here.

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,871 @@
...
+ParaView provides many data filters for transforming data and
+performing analysis tasks. However, there are an infinite number of
+operations one may want to perform on a data set. To address the need
+for custom filters, ParaView supports a rich plugin architecture that
+makes it possible to create additional filters in C++. Unfortunately,
+creating a plugin this way is a relatively involved process.
+
+Aside from the C++ plugin architecture, ParaView provides a
+Programmable Filter that enables a potentially faster plugin
+development path. This filter is more versatile than the Python
```
prabhuramachandran: What are the downsides of this path (apart from performance if implemented in pure Python)?  Will these filters work in parallel?

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,871 @@
...
+of the data arrays in a data set are copied to the client. By default,
+no arrays are copied. Arrays can be requested using methods available
+in the ``vtkPythonView`` class instance that is passed in as an
+argument to the ``setup_data`` function, e.g.,
+
+.. code-block:: python
+
+   def setup_data(view):
+       view.SetAttributeArrayStatus(0, \
+           vtkDataObject.POINT, "Density", 1)
```
prabhuramachandran: Perhaps a link to where more information on what the above is and how it works would be useful to readers.

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,871 @@
...
+
+       dens = do.GetPointData().GetArray('Density')
+
+       # Convert VTK data array to numpy array
+       from paraview.numpy_support import vtk_to_numpy
+
+       ax.hist(vtk_to_numpy(dens), bins=10)
+
+       return python_view.figure_to_image(figure)
+
```
prabhuramachandran: I think that a more useful example would be to show how one may embed ParaView in a Python + Qt (PyQT/PySide) application -- if that is possible.  I think this will be of great interest.  In fact, this would be more useful than the VTK-Python-Qt integration example you show earlier as there are many examples of various tools that use/do this including Mayavi which is well known in the Python community.

prabhuramachandran:
Thanks for a very informative paper.  The paper is well written and quite clear with many examples. The paper also highlights some of the recent improvements made to the Python support in VTK and ParaView. I have made specific comments throughout the paper.

I think it would help to mention which version of VTK or ParaView you are
referring to, as there are many different versions people use in the community.

I was able to follow the paper well as I am very familiar with VTK.  I have no
familiarity with ParaView though and was happy that I was able to follow that
content.  However, I fear that many users unfamiliar with VTK (and there are a
great many of them), will struggle to see the Scientific data analysis and
visualization parts in the first part of the paper.  As a result, the first
part of the paper is more like a collection of useful recent (and older)
features of the VTK-Python wrappers that a VTK user would appreciate.

I think the problem is that the paper really talks about the use of Python and
the Python integration provided by VTK and ParaView assuming some knowledge of both.  This is different from the title which suggests that the user will learn about how they can analyze and visualize their data using VTK and ParaView.  So if the title read "Using VTK and ParaView from Python", then
it would make a lot more sense as a reader.  

If this is indeed the case, I would also like to know if there are any important aspects that have not been covered?  For example, I think VTK 6.x features some WebGL integration, this is not mentioned -- do you think this is of relevance?  Another very useful feature would be off-screen-rendering capabilities.  This has always been tricky with VTK requiring Mesa and a separate VTK build.  Some users who use VTK to batch process visualizations would probably find this useful.  I don't know if this has changed recently. 
Of course, this is all perhaps biased and based on my interests, so you can use your judgement on these.

papers/cory_quammen/cory_quammen.rst
cquammen: Prabu, just a quick note to let you know I have received your comments and will be addressing them shortly. Thanks for your review.

cquammen: I checked and LD_LIBRARY_PATH is not needed to make VTK available in my system python. VTK now loads the libraries with Python's dl package if it is available. I have added a section on where to obtain VTK.

papers/cory_quammen/cory_quammen.rst
cquammen: Good suggestion. I added a warning.

papers/cory_quammen/cory_quammen.rst
cquammen: I added a comment specifying VTK and ParaView versions discussed in the paper. In my revised version, I am assuming my audience knows little to nothing about VTK and ParaView.

papers/cory_quammen/cory_quammen.rst
cquammen: I will look into adding an example if it doesn't require too much additional context to make sense. It might just be a screen shot.

papers/cory_quammen/cory_quammen.rst
cquammen: The Programmable Filter will indeed work in parallel. I'll add a note on where one can learn more about parallel processing in VTK/ParaView since it alone is a large topic.

papers/cory_quammen/cory_quammen.rst
cquammen: Added.

cquammen:
Thanks for your feedback, Prabhu. I have added some background on the VTK data model and pipeline concept. Hopefully that will make it more accessible to those with no experience with either VTK or ParaView. They are indeed my target audience, but it is a little challenging to fit both an introduction to VTK/ParaView and overview of all the ways Python is integrated into 8 pages.

jaberg:
@prabhuramachandran Have your comments been addressed?

hplgit:
I have read the most recent version of this paper and think it is a clear and very informative presentation of Python's role in the important C++ projects VTK and ParaView, and how Python simplifies the interaction with the packages. Both VTK and ParaView are quite old projects now and it is nice with a recent update on the status. As a plain ParaView user I learned a lot more of the capabilities offered by the Python interface.

@jaberg: This paper is can be accepted.

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,1081 @@
...
+.. code-block:: c++
+
+   double t, x[3]
+   plane->IntersectWithLine(point1, point2, t, x);
+
+In Python, the equivalent is
+
+.. code-block:: python
+
+   t = mutable(0.0)
```
prabhuramachandran: This should really be `vtk.mutable`.

papers/cory_quammen/cory_quammen.rst
```diff
@@ -0,0 +1,1081 @@
...
+The ``paraview.simple`` layer provides a simplified layer of Python
+functions to create pipelines and modify filter and visualization
+properties. The same example above expressed with``paraview.simple``
+functions is reduced to
+
+.. code-block:: python
+
+   from paraview import simple
+
+   Sphere(Radius=2.0)
```
prabhuramachandran: Should this be `simple.Sphere`? Otherwise it is surprising -- I am aware that paraview already does `from paraview.simple import *` but in that case the previous import does not seem necessary.

prabhuramachandran:
@jaberg : LGTM.  There are couple of really minor points which can be addressed if there is time otherwise, you can go ahead.  Thanks @cquammen for making the suggested changes and reorganizing the paper.

papers/cory_quammen/cory_quammen.rst
cquammen: In fact, no, it does not need to be ``simple.Sphere``. Importing ``paraview.simple`` runs a function that creates new global variables referring to Python functions for various ParaView objects, one of which is ``Sphere``. Perhaps they should be created in the ``simple`` module namespace, but that isn't presently how it is done.

papers/cory_quammen/cory_quammen.rst
cquammen: Sorry, I didn't have time to come up with an example. I'm not even entirely sure if this is possible the way you described.

papers/cory_quammen/cory_quammen.rst
cquammen: Very nice catch. I have modified the example.

cquammen:
Thank you for the reviews! I'm looking forward to seeing you all in about a week.

PR #154, opened 2015-06-16
Paper: Testing Generative Models of Online Collaboration with BigBang (Benthall)

sbenthall:
Pull request for Sebastian Benthall paper.

Regrettably, the current state of this paper is a (very) incomplete draft. More revisions to follow.


sbenthall:
Lots of edits. Still in the works--will be working heavily today and tomorrow. Reviews appreciated at any point.

katyhuff:
@sbenthall When you (not sure why) merged allen downey's paper branch into your branch (in this commit: c06c563) , you made it such that the build server builds Allen Downey's instead. 

This is a problem because the reviewers rely on the buildbot to review your paper. Please revert commit 	c06c563 .

sbenthall:
Thanks for the heads up @katyhuff . I've reverted the commit. Please let me know if there are any more troubles with the build.

katyhuff:
@sbenthall , thanks! It looks like it's building the right paper now. I'm changing the tag to needs-review. I will review your paper myself this evening, but am hopeful that a social scientist may be able to look it over as well.

(I know it's short notice for most folks, but perhaps @AllenDowney would be an excellent reviewer, if he can find the time.)

sbenthall:
Thank YOU, @katyhuff . In fact, the paper has a bit more of a complex systems bent to it than traditional social science. So a physicist's review would be excellent. The feedback of a lucid statistician like @AllenDowney would of course be most welcome as well.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+   Mailing lists serve as critical communications infrastructure for
+   many communities, including several of the open source software 
+   development communities that build scientific Python packages.
+   BigBang provides tools for analyzing mailing lists.
+   In this paper, we present results from testing a generative
+   model of online collaboration in mailing lists.
+   We derive a social network fro archival mailing list history
+   and test the Barabasi-Alpert model of network formation
+   against this data.
+   We find the model does not fit the data, but that mailist list
```
takluyver: mailist -> mailing

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+
+Thorough study of these kinds of communities requires the collection and
+rationalization of many heterogenous and high-dimensional data sources,
+including but not limited to:
+
+- mailing lists
+- version control systems such as Git
+- issue trackers such as GitHub and Bugzilla
+
+This data is complex in it has many dimensions that afford very different
```
takluyver: in *that* it has?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+repository for review through the GitHub Pull Request system.
+Source code that performs preprocessing that is usable by multiple computational
+experiments is in a separate source code directory that can be imported as a
+Python module.
+
+In the context of Scientific Python, BigBang is deliberately recursive.
+It is a Python project that depends on many other scientific Python projects.
+It is designed to study, among other things, dependencies and interactions between 
+the Scientific Python technologies and communities.
+Our goal is for BigBang to providing a new means for these communities to engage i
```
takluyver: stray i at the end of this line

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+in scientific self-management.
+
+Email data collection
+---------------------
+
+BigBang supports collection of email data.
+It can do this either by scraping the archival pages of a Mailman instance,
```
takluyver: Mailman 3 is now out, and I know there's a new archiving system. I've not yet seen any project using the new version, but you may want to make it clear that it's the Mailman 2 archive pages it curently scrapes.

IIRC one of the key features of Mailman 3 is a REST interface, so when you do need to support that, hopefully no scraping will be needed. :-)

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+that are essential for email processing and presentation.
+
+For our study, we have been interested in extracting the implied social
+network from an email archive of a public mailing list. To construct this
+network, we have used the `From`, `Message-ID`, and `In-Reply-To` headers
+of the email.
+
+The `From` field of an email contains data about the sender of the message.
+This often includes both a full name and an email address of the sender.
+As this is set by the email client (note: or is it the mailing list server?)
```
takluyver: I believe it's the email client.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+a single person is often represented differently in the `From` field
+across many emails. See *Entity Resolution* for our strategies for
+resolving entities from divergent email headers.
+
+The `Message-ID` header provides a globally unique identification string
+for every email.
+This is accomplished by (how?).
```
takluyver: I didn't know about Message-ID headers, but Wikipedia has a nice summary - assuming it's accurate, of course ;-)

https://en.wikipedia.org/wiki/Message-ID

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+The `In-Reply-To` header is set when an email is sent as a reply to
+another email.
+The reply's `In-Reply-To` header will match the `Message-ID` of the
+original email.
+
+We construct the directed *interaction graph* :math:`G` for a set of emails as follows:
+
+* For every email :math:`e`, add a node :math:`v_{f(e)}` to :math:`G` corresponding 
+  to its `From` header :math:`f(e)`  and set its `sent` attribute :math:`v_{f(e)}` 
+  (notation is sloppy here -- can we write a pseudocode algorithm?)
```
takluyver: I agree, this is not very clear. It sounds like the nodes represent senders, but if the same sender has sent multiple emails in a conversation, do you add multiple nodes with the same name? Or does networkx automatically ignore attempts to add the 'same' node twice?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+
+The motivation for constructing interaction graphs in this way is to build a
+concise representation of the social network implied by email data.
+We posit that building a social network representation based on actual messages
+sent between people provides a more granular and faithful description of
+social relationships than one based on higher-level descriptions of social
+relationships or ties from web services such as Facebook 'friends' and
+Twitter 'followers'
+
+It may be objected that since the data we are considering comes from public
```
takluyver: The passive voice feels a bit forced here.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+By looking at the *Reply-To* header of the emails, we
+are able to construct a graph of who replies to who in the email list. (How?)
+Drawing on (x,y, and z) we are studying the empirical properties of these
+networks in order to come up with a generative model of community graph.
+Whereas (Barabasi reference) models graphs with power-law degree distribution,
+we find (verify using Clauset et al. method) that degree distribution in
+these graphs is log normally distributed. Moreover, whereas [Newman2002]_
+hypothesizes that socially generated graphs will be characterized by high
+degree assortativity, we find empirically that these interaction graphs
+have degree assortativity comparable with biological and technical networks.
```
takluyver: Whew. I can see that this paragraph is still a work in progress, but the paper has suddenly gone from pleasantly readable to dense jargon. I see there's more detail about this below, so I'll read on, but maybe this paragraph should be moved into that section.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+   | ``"Tyrion Lannister" <Tyrion.Lannister@gmail.com>``                        |
+   +----------------------------------------------------------------------------+
+
+Variation in entity representation is a source of noise in our research and an
+ongoing area of development for BigBang.
+For the study in this paper, we have implemented a heuristic system for
+entity matching.
+First we standardize the data by converting it to lower case and normalizing
+" at " and "@". Then we parse out the email and full name subfields of the
+data and look for exact matches.
```
takluyver: exact matches of both fields, or of either one?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+Under this procedure, all of the above email addresses would be collapsed into a single
+entity.
```
takluyver: Is this robust to the order they are processed in? i.e. the process that you've described would not collapse `Tyrion.Lannister@gmail.com` and `halfman@council.kings-landing.gov (halfman@council.kings-landing.gov)` if they were the only two `From` headers for that sender. If it later sees additional variants, can it still combine two nodes that it has previously seen and decided are distinct?

I understand that this may be more detail than you want to go into in the paper, but it interested me reading through it.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+Deriving interaction graphs
+---------------------------
+
+Email is archived in the same text format that email is sent in, as specified in
+RFC2822 [RFC2822]_.
+Every email is comprised of a message body and a number of metadata headers
+that are essential for email processing and presentation.
+
+For our study, we have been interested in extracting the implied social
+network from an email archive of a public mailing list. To construct this
```
takluyver: Does it work with the archive of only a single list, or can you build a network across multiple lists? There are people who are active on multiple lists within a broader community, and I'd expect to see a different network structure when multiple lists were included in one graph.

papers/sebastian_benthall/sebastian_benthall.rst
takluyver: Is there any way to evaluate the effectiveness of the entity resolution mechanism? E.g. if you manually resolved a sample of senders - assuming that a human can do a near perfect job by using external information and spotting less clear-cut similarities - you could compare this to the algorithm's result for the same data and count the rate of false negatives and false positives.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+The edges to and from the derived node are weighted by the sum of the edges of the original
+nodes.
+The `sent` attribute of the new node is also set as the sum of the `sent` attribute of the
+original nodes.
+
+Preferential attachment model
+-----------------------------
+
+A early result in the study of complex networks was the observation that many networks
+existing in nature exhibit a scale-free degree distribution.
```
takluyver: 'scale-free' sounds like it has a significant meaning, but it's not exactly clear to me what that is, or why that relates to a power-law distribution.

My comments are doubtless biased by my own knowledge; I'm very familiar with mailing lists, and not in the least familiar with network analysis.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+In particular, the Barabasi-Alpert model attributes the scale-free distribution of
+node degree to a *preferential attachment* mechanism. 
+The network is formed by beginning with a small number :math:`m_0` of nodes and adding
+new nodes, connecting each new node to :math:`m < m_0` nodes, where the probability
+of connecting to node :math:`i` is relative to the prior degree of that node,
+:math:`k_i`:  adding new nodes to the network one by one. 
+
+.. math::
+
+   \Pi(k_i) = \frac{k_i}{\sum_{j} k_j}
```
takluyver: This is probably revealing my ignorance of this whole topic, but the only symbols I recognise in this equation are `k_i` (which was defined above) and sum. What's Pi? What's j?

I'd guess that the overall meaning is: the probability of connecting to node `i` is the degree of `i` over the sum of the degrees of all nodes in the graph.

How does this probabilistic model interact with the `m < m_0` above, which sounds like a hard limit? Is that actually an expectation rather than a hard limit, or does the new node stop connecting once it has degree `m_0 - 1`?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+--------------------
+
+Degree assortativity is the correlation between degrees of adjacent nodes in the network.
+Following the definitions of [Newman2003]_, the degree assorativity coefficient is
+
+.. math::
+
+   r = \frac{\sum_{jk}jk(e_{jk} - q_{j}q_{k}))}{\sigma_{q}^{2}}
+
+In the above formula, :math:`e_{jk}` is the fraction of edges the connect vertices
```
takluyver: edges *that* connect?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+.. math::
+
+   \sum_{j} e_{jk} = q_{k}
+
+The value :math:`\sigma_{q}` is the standard deviation of :math:`q_k`. [TODO: I'm using directed
+assortativity here, yes? And what about weighted degrees?]
+
+Degree assortativity in complex networks is studied by [Newman2002]_, who makes the intriguing 
+claim that observed social networks, such as coauthorship networks, exhibit positive degree
+assortativity, while technical and biological networks exhibit negative degree assortativity.
```
takluyver: I think I follow what these terms mean - *positive degree assortativity* means that highly connected nodes are even more likely to be connected to other highly connected nodes, right? - but it would be nice if there was a brief clarification of that.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+A further challenge to the Barabasi-Alpert model comes from [Clauset2007]_, who argue that
+many conventionally accepted techniques for fitting power law distributions to empirical data 
+are biased and unsound.
+They propose a Bayesian technique for testing power law distributions.
+By computing the likelihood of the data being generated by a power law distribution and
+comparing it with the likelihood of it being generated by other heavy-tail distributions,
+such as the log normal distribution, they provide a statistically sound basis for model
+comparison.
+The Clauset et al. method also carefully considers only the tail of the data, picking a
+cutoff value below which data are ignored.
```
takluyver: Is this cutoff in terms of degree? What's the advantage of doing this?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+We built interaction graphs according to the above procedure for 10 mailing lists from open
+collaborative communities.
+We then computed the degree assortativity of these networks.
+We also used the Alstott package to test the degree distribution of these networks using
+the Clauset method.
+
+Results
+-------
+
+Every mailing list of the 10 we analyzed exhibits degree disassortivity and a significantly
```
takluyver: Is disassortativity the same as negative assortativity, or a lack of assortativity in either direction?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+   +---------------+----------------+-----------+-----------------------+---------+------------+
+   | list name     | Source         | :math:`n` |  Degree Assorativity  | R value | :math:`p`  |
+   +===============+================+===========+=======================+=========+============+
+   | ipython-dev   | SciPy          | 689       | -0.246441169106       | -0.518  |  0.080     |
+   +---------------+----------------+-----------+-----------------------+---------+------------+
+   | potlatch-dev  | OpenStreetMap  | 75        | -0.0568958403876      | -0.001  |  0.969     |
+   +---------------+----------------+-----------+-----------------------+---------+------------+
+   | scipy-dev     | SciPy          | 1056      | -0.276991197113       | -0.331  |  0.578     |
+   +---------------+----------------+-----------+-----------------------+---------+------------+
+   | ipython-user  | SciPy          | 1085      | -0.267104106913       | -0.334  |  0.227     |
```
takluyver: It would be interesting to know whether the pattern is different for -user mailing lists, where anecdotally the typical pattern is of many users (low-degree nodes) asking questions which a few experienced people (high-degree nodes) reply to, versus -dev lists which we think of as more homogeneous discussion. It doesn't really look like there's a difference here.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+-------
+
+Every mailing list of the 10 we analyzed exhibits degree disassortivity and a significantly
+(:math:`p` > .05) better fit to log normal instead of power law distribution.
+
+.. table:: Results of analysis. For each mailing list archive, number of participants :math:`n`,
+           computed degree assortativity of the interaction graph, and loglikelihood ratio R and
+           statistical significance :math:`p` of comparison of fit between power law and log normal
+           distributions. In all cases the interaction graph is disassortative with significantly
+           more log normal degree distribution. :label:`mtable`
```
takluyver: You say 'significantly more', but none of the p values is particularly small, and some of them are very close to 1. Is this a different kind of `p` from what I'm used to?

takluyver: Oh, I see now your note above about `p > 0.5`. Is this standard in the Bayesian world?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+collaborative communities.
+We then computed the degree assortativity of these networks.
+We also used the Alstott package to test the degree distribution of these networks using
+the Clauset method.
+
+Results
+-------
+
+Every mailing list of the 10 we analyzed exhibits degree disassortivity and a significantly
+(:math:`p` > .05) better fit to log normal instead of power law distribution.
```
takluyver: Is this difference something that can be illustrated? i.e. can you draw a histogram of one of the datasets, overlaid with the best-fit log normal and power law distributions?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+                 Scale Free: Twitter’s Retweet Network Structure. 
+                 At Network Science 2012, Evanston, US.
+
+.. [Zanetti2012] Zanetti, M. and Schweitzer, F. 2012.
+                 "A Network Perspective on Software Modularity"
+                 ARCS Workshops 2012, pp. 175-186.
+
+.. [Zhou2011] Zhou T, Medo M, Cimini G, Zhang Z-K, Zhang Y-C (2011) 
+              Emergence of Scale-Free Leadership Structure in Social 
+              Recommender Systems. PLoS ONE 6(7): e20648.
```
takluyver: Formatting note: these last two references are clashing with the table, which is wider than a single column. I don't know what, if anything, you can do about this.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+to include the \math:`x_{min}` values and include data sets above a certain size threshold ]
+
+
+Discussion
+----------
+
+The regularity in these data sets suggests that there is a need for a new network generation
+model that results in disassortative graphs with log normal degree distribution.
+Building this graph generation model may help us understand better how collaborative communities
+grow and form over time.
```
takluyver: Is there anything we can say from these results about the communities themselves? Does it mean anything, in human terms, that they have negative assortatitivity, or that they fit a log normal degree distribution better than a power law?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+
+Possible models include nonlinear Barabasi-Alpert: http://arxiv.org/pdf/cond-mat/0402315.pdf
```
takluyver: Should this be a citation?

takluyver:
Thanks, this was an interesting read. To summarise the major points from my inline comments:

1. Some more explanation of the graph theory concepts involved would be valuable to make the second half of the paper more broadly accessible.
2. If there are any conclusions to be made about the communities themselves, that would be very interesting. What other kinds of networks produce similar statistics, for instance? There's an intriguing sentence "these interaction graphs have degree assortativity comparable with biological and technical networks." What are those more specifically? 'Biological networks' could be food webs, gene networks, observed interactions of individuals... and I haven't got a clue about the technical ones! Are there other similarities with these networks?
3. It looks like the results each represent a single mailing list, but it's not uncommon for people to post on multiple mailing lists. Have you looked at joining the networks across lists? Would you expect similar results from the combined graph(s)?

sbenthall:
Thank you for these thorough and considerate comments, @takluyver

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+
+This paper introduces BigBang, a Python project whose purpose is the collection,
+preprocessing, and analysis of data from open collaborative communities.
+Built for the use case of studying the Scientific Python communities in particular,
+it generalizes to other communities and supports fruitful comparisons between them.
+
+To demonstrate the potential of this approach, this paper will explore the
+structure of mailing list discussions in the context of open collaborative projects.
+We extract social network data from the archives of public mailing lists and test
+the plausibility that these graphs were generation by Barabasi-Alpert network model.
```
ankurankan: generation -> generated ?

AllenDowney:
Interesting paper.  Thanks for the chance to comment.  A few suggestions:

1) Page 3, col 2, "the Erdos-Renyi model..."  Check the spelling of Erdős–Rényi, and add the accent marks, if possible.  Same with Barabási–Albert.

2) Same location, "produces networks with normal degree distribution".  It's actually binomial, I believe, converging to Poisson as n increases.

3) Page 4, col 1, discussion of "degree assortativity":  It might be nice to provide some intuition for this statistic.  If I understand correctly, in social and co-authorship networks positive assortativity indicates that high-degree participants tend to work together (which is not too surprising), whereas in technical and biological networks (provide examples, maybe), high-degree nodes tend to be connected to low-degree nodes (this is less obvious -- is there an explanation?).

4) Page 4, col 2, "would not expect log normal power law distribution"  I think you mean "log normal degree distribution"  (since the lognormal distribution does not follow a power law).  Also, I suggest lognormal as one word, or hyphenated.

5) Same location, you suggest that preferential attachment models are not expected to produce lognormal degree distribution.  They can, with some tweaks.  In general, the distinction between Pareto and lognormal distributions tends to be subtle, in two ways: often when you are looking at data, it is all but impossible to distinguish them in a robust way, and when you are modeling generative processes, small differences in the model can cause the outcome to flip from one to the other.  See Mitzenmacher's discussion:  https://www.stat.berkeley.edu/~aldous/Networks/1089229510.pdf

6) Table 2 exceeds the column width.  You might be able to beat it into shape by rounding all values to two sig-figs (which is a good idea anyway).  And right-justify the numerical values.

7) In the log-likelihoods, what is the base of the log?  For Bayes factors, base 10 seems to be common, and it is what I understand intuitively, but I am not sure how universal that is.

8) I’m not sure about how you are interpreting the p-values.  It looks like none of them are below 5%, so if you treat the lognormal distribution as the null hypothesis, then you are failing to reject the null.  I am not wild about this methodology, since I am not sure why the lognormal should be the null.  And be careful about interpreting a negative result: you can’t really say whether this means that the lognormal model is better, or if you just don’t have enough data to distinguish them.  In my opinion, the best way to get a grip on the tail behavior of an empirical distribution is to plot the CCDF on a log-log scale.  A power law/Pareto should be a straight line; a lognormal distribution drops off more quickly.  If you have the time to generate a figure like that, it would be a nice addition.  If you want to send me data, I could take a look, although no guarantee I could turn anything around in time for the proceedings.

9) I think your conclusion is valid: it seems like there is something about these datasets that is not captured by the ER or BA models.  It will be interesting future work to see if there are other models that do the job better.  You might consider Watts and Strogatz small world graphs: https://en.wikipedia.org/wiki/Watts_and_Strogatz_model

(I am not optimistic that it will any better, but it is the other standard model to consider, and the exercise might be enlightening)

One thing to keep in mind about these kinds of modeling efforts: you can always find (or make) models that fit the data better.  The question is when to stop; that is, when to decide that that the model is good enough for the intended purpose.  But to answer that question, you have to have an intended purpose.  What do you want this model for?

sbenthall:
Thanks so much for these helpful comments, @AllenDowney Especially the substantive corrections.

I find the argument in the Clauset et al. paper for their test of power law vs. log normal compelling (as an alternative to the more common linear test). I encourage you to take a look at it if it's interesting to you:
http://arxiv.org/abs/0706.1062

I will address this in more depth in the next version of the paper. Thanks again.

papers/sebastian_benthall/sebastian_benthall.rst
jseabold: providing -> provide?

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+We find that of the mailing lists we've analyzed, none exhibit two features of
+Barabasi-Alpert networks: power law degree distribution and zero degree assortativity.
+This result suggests the possibility of future work of scientifically developing a
+generative model of collaboration.
+
+BigBang Overview
+----------------
+
+Launched last year and presented at SciPy 2014, BigBang is a software project
```
jseabold: Maybe a specific date instead of `last year`? We have enough context here, but generally relative dates in papers always confuse me.

papers/sebastian_benthall/sebastian_benthall.rst
jseabold: +1

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+in a way that's similar to the (cite) blockmodel technique.
+The edges to and from the derived node are weighted by the sum of the edges of the original
+nodes.
+The `sent` attribute of the new node is also set as the sum of the `sent` attribute of the
+original nodes.
+
+Preferential attachment model
+-----------------------------
+
+A early result in the study of complex networks was the observation that many networks
```
jseabold: A -> An

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+comparing it with the likelihood of it being generated by other heavy-tail distributions,
+such as the log normal distribution, they provide a statistically sound basis for model
+comparison.
+The Clauset et al. method also carefully considers only the tail of the data, picking a
+cutoff value below which data are ignored.
+This method of testing power law distributions is implemented in Python in the `powerlaw`
+package by [Alstott2014]_
+
+From a Bayesian perspective, the ratio of likelihoods represents how much one should
+update ones beliefs based on observation of data.
```
jseabold: one's

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+In this case, the computed likelihood ratio of the data being generated by a power law
+over a log normal distribution would be interpreted as how much one should be
+persuaded that the data came from a power law distribution based on ones prior
```
jseabold: one's

papers/sebastian_benthall/sebastian_benthall.rst
jseabold: +1. I'd like to see a visualization here, if possible. I'd also like to know what test you're using to measure the goodness of fit.

papers/sebastian_benthall/sebastian_benthall.rst
jseabold: This definitely needs some clarification.

papers/sebastian_benthall/sebastian_benthall.rst
jseabold: I'll echo this one as well. I think it would be super nice to help readers to build up intuition about these things.

papers/sebastian_benthall/sebastian_benthall.rst
```diff
@@ -0,0 +1,574 @@
...
+belief in the originary distribution of data in general.
+
+While this may be a perplexingly abstract way of considering the problem for a
+non-Bayesian, there is an argument that log normal distributions should be
+given a higher prior probability than power law distributions.
+This is because of the Central Limit Theorem, from which it follows that
+a log normal distribution would result from the multiplication of many
+other otherwise distributed factors.
```
jseabold: May be nice to be a little more explicit that you're making a frequentist argument for the use of the (log-) normal distribution. It's pretty clear to me, but may not be to the average reader.

papers/sebastian_benthall/sebastian_benthall.rst
jseabold: I had the same reaction. It may be worth gently introducing and discussing power law distributions briefly a little earlier and also why it matters. What's at stake for incorrectly assuming a power-law distribution vs. a log-normal?

cchoirat:
Hi @sbenthall:

Here are some comments about the paper:

- p. 2, Section "Deriving interaction graphs": it might be easier to follow if the paragraphs "The motivation for constructing interaction graphs in this way" etc. were before the code snippet.
- p .3. Could you provide a reference "A early result in the study of complex networks was the observation that many networks existing in nature exhibit a scale-free degree distribution"?
- p .3. "formulahere" to be fixed?
- p. 3, "The Barabasi-Alpert model is now widely known in the literature."  Could you please elaborate?
- p .4, Section "Methods".  "We built interaction graphs according to the above procedure for 10 mailing lists from open collaborative communities."  Could you please explain how the mailing lists were selected and how stable your results are wrt changes in sample size?
- p. 4. "Every mailing list of the 10 we analyzed exhibits degree disassortivity and a significantly (p > .05) better fit to log normal instead of power law distribution."  I agree with @AllenDowney that some work is needed here.

sbenthall:
Thanks, @jseabold and @cchoirat for these comments!
I'm excited to begin revisions today.

papers/sebastian_benthall/sebastian_benthall.rst
sbenthall: addressed here: https://github.com/sbenthall/scipy_proceedings/commit/a87f11331d5b636980b46f68e585b64564426173

papers/sebastian_benthall/sebastian_benthall.rst
sbenthall: I expect it will be hard to get your feedback on this over the holiday weekend, but reading this again...I'm not sure I understand your point.

Can you elaborate about why you say the argument for use of the log normal distribution is frequentist?

jseabold: Well, I guess that's what I'm asking you to do, and I thought that's what
you were getting at. The argument here seems to be appealing to
"non-Bayesians," calling on the CLT for justification. I suppose limiting
distributions aren't strictly a frequentist thing, but it does underlie all
of the large-sample theory of estimation leaned on by the repeated sampling
framework. If that's not the appeal you're trying to make then perhaps a
rephrasing? Just looking for some clarification but overall a minor point.

sbenthall: gotcha. Thanks.

sbenthall:
As per suggestion by @AllenDowney I would like to right-justify numerical values in tables. I was wondering if somebody, maybe @katyhuff , knew how to do this in rst. The only way I've been able to find so far involves unicode whitespace :(

http://stackoverflow.com/questions/7348208/how-to-right-align-columns-content-in-restructuredtext-simple-tables

AllenDowney:
I used raw latex for the tables in mine (there's an example in the template
document).

On Sat, Jul 4, 2015 at 7:48 PM, Sebastian Benthall <notifications@github.com

sbenthall:
Ah, I see. Thanks!

sbenthall:
Ok. I've completed my revisions to the paper.

katyhuff:
@sbenthall Glad to hear you feel your paper is ready. 

It's best if I can get a confirmation from @AllenDowney @cchoirat and @jseabold that your changes have satisfied their needs. So, @sbenthall : it would  help if you gave a summary of how you addressed their issues (or, if they're willing to look it over once more, or look over your commit history to check on their own, that would work too). I think @takluyver's comments were satisfied early on. 

Thanks All!

AllenDowney:
My comments were all suggestions, so I am satisfied.

sbenthall:
Roger, @katyhuff . Here is a non-exhaustive list of changes:

All reviewiers:

1. I've corrected my use of p-values and explained what they mean in the context of the Clauset power law test method. Thanks for calling me out on my mangling of this before.
2. I've also written more to defend my use of the Clauset method and explain it.
3. I have provided an example of a log-log pdf plot by way of explanation of the usual power law test method. This plot is for the aggregated interaction graph (see below) that actually is significantly not power law.
4. I've included a discussion of weighted and unweighted assortativity, and the relevant data.

@takluyver : 

1. I've expanded considerably on the graph theory sections.
2. In the graph theory sections and in the Discussion section, I've provided more intuitions and interpretation for what disassortativity and preferential attachment mean in the context of a social graph.
3. On your recommendation, I have included in the revised paper an analysis of an interaction graph created by combining the individual graphs. I've included a visualization of this graph and a log-log plot of the PDF of its degree distribution, in addition to the other metrics.

@AllenDowney :
I've corrected (1) accenting and (2) the degree distribution of the ER model, and another (4) typo. I've also fixed the table. (6)

I've provided more intuition for the assortativity statistic (3), corrected the discussion of p-values and otherwise included more material defending my method (8). I've also folded in the insights of Mitzenmacher (5) and adjusted my interpretation accordingly.

I have provided more general motivation for this paper in the discussion section.

@cchoirat :

I've (a) rearranged the section about deriving interaction graphs for clarity, (b) provided a reference for the early study of scale-free networks, (c) fixed the missing formula, (d) minced a words a little https://github.com/sbenthall/scipy_proceedings/commit/bbb13ebd82baa56e4fc06be709c4da0ee9cb5bde

(e) explained how data were selected and discussed stability of results wrt to sample size especially in the context of the aggregate list,

and corrected the (f) p-value discussion.

p .3. Could you provide a reference "A early result in the study of complex networks was the observation that many networks existing in nature exhibit a scale-free degree distribution"?
p .3. "formulahere" to be fixed?
p. 3, "The Barabasi-Alpert model is now widely known in the literature." Could you please elaborate?
p .4, Section "Methods". "We built interaction graphs according to the above procedure for 10 mailing lists from open collaborative communities." Could you please explain how the mailing lists were selected and how stable your results are wrt changes in sample size?
p. 4. "Every mailing list of the 10 we analyzed exhibits degree disassortivity and a significantly (p > .05) better fit to log normal instead of power law distribution." I agree with @AllenDowney that some work is needed here.

@jseabold : I believe I've addressed most of your comments but don't have a summary of yours to work with; many of the most critical changes you also requested have been listed above.

katyhuff:
Thanks @sbenthall , that helps me see the changes more clearly. Appreciated!!

katyhuff:
@sbenthall : I hard-committed your paper (cuz I couldn't fix the weird commit history with the deletion of Allen Downey's paper and it kept screwing up my publication-building-main-2015  branch.) So, this has been merged, even though it doesn't look like it. !!

sbenthall:
Thanks Katy! Sorry about the hassle.

katyhuff:
No worries at all. I learned a couple things about git refs in the process. Thanks for your submission (and congratulations!). The proceedings will be online post haste (I'll send a link).

PR #156, opened 2015-06-16
Paper: PyRK: A Python Package For Nuclear Reactor Kinetics (Huff)

katyhuff:
Paper: PyRK: PyRK: A Python Package For Nuclear Reactor Kinetics (Huff)

[Edit by jaberg] http://nipy.bic.berkeley.edu:5000/download/6


papers/kathryn_huff/kathryn_huff.rst
```diff
@@ -0,0 +1,625 @@
...
+   \rho &= \frac{k-1}{k}
+   \intertext{where}
+   \rho &= \mbox{reactivity}\\
+   k &= \mbox{neutron multiplication factor}\\
+     &= \frac{\mbox{neutrons causing fission}}{\mbox{neutrons produced by fission}}.
+
+The reactor power is stable (*critical*) when the effective multiplication
+factor, :math:`k`, equals 1. For this reason, in all power reactors, the scalar
+flux of neutrons determines the power. The reactor power, in turn, affects the
+temperature. Reactivity feedback then results, due the temperature dependence
```
jeanconn: Should there be a "to" after "due" or is this a domain-specific phrasing?

katyhuff: I should drop the comma. ... Reactivity feedback results due to the temperature dependence...

Would that be more clear?

jeanconn: I think the sentence can work with or without the comma, as long as you use "due to the temperature dependence" instead of the current "due the temperature dependence".    It does seem easier to get through the list at the end of the sentence if you take out the comma.

papers/kathryn_huff/kathryn_huff.rst
```diff
@@ -0,0 +1,625 @@
...
+------------
+
+Time-dependent fluctuations in neutron population, fluid flow, and heat transfer are
+essential to understanding the performance and safety of a reactor. Such
+*transients* include normal reactor startup and shutdown as well as abnormal scenarios
+including Beyond Design Basis Events (BDBEs) such as Accident Transients
+Without Scram (ATWS). However, no open source tool currently exists for
+reactor transient analysis. To fill this gap, PyRK (Python for Reactor
+Kinetics) [Huff2015]_, a new python package for nuclear reactor kinetics, was
+created. PyRK is the first to offer an open source tool designed to conduct:
```
jeanconn: I really like the way you are setting up for the little list of the features of PyRK here, but I'm not entirely clear what PyRK is the "first" *of* in this construction.  The first new python package?  The first project?  The ambiguity may be fine for your purposes though.

wfspotz:
I don't see any numerical results (my browser gives an incomplete rendering of the RTF, but looking at the raw RTF itself, I'm pretty sure I'm not missing anything). Shouldn't you give some plots as described earlier?

wfspotz:
In the Neutronics section under "PyRK Implementation”, the last sentence with `1+i+k` should probably be `1+j+k`. (At least I don't see a definition of `i` previous to this...)

wfspotz:
First sentence of the THSystem section, also under "PyRK Implementation”: “Any reactor is made up of a handful of neutronic object needs a temperature." This looks like an editing snafu.

lederernc:
Last sentence of Background Neutronics section does not make sense. Possibly should read "... in a reactor and are dependent on ...". I think the background sections for Reactivity and The PRKE need a lot of work. I see inconsistent use of rho and p. The equation with \alpha_i appears to be missing a summation or product symbol. Equation 5 appears to use three different lambdas and the two lower case ones are not distiguishable as they are in equations 10 and 14.

The PyRK Implementation section on Neutronics contains duplication in the first and last paragraphs.

The first sentence in the THSystem paragraph does not read well in English. There is also a lot of clean up needed in the markup of that section, e.g. gls{PBFHR}, gls{TRISO}, LaTeX throughout, the word radiation is misspelled before the equations for Q.

katyhuff:
So sorry for the sloppiness of this submitted draft, all (@jeanconn, @wfspotz, and lederernc). Thank you for your comments! I'll have a chance to address these issues today and will ping you with a summary when they've been addressed.

katyhuff:
@jeanconn : Thank you for your comments! They were very helpful for improving clarity in my submission.

1) I have fixed the sentence concerning temperature feedback of reactivity to include the essential "to" that makes the sentence readable.
2) I have clarified what PyRK is the first of in this case. It is the first open source tool for reactor kinetics (as far as I know). It is also probably the first to implement an object-oriented lumped capacitance model to support generic reactor prototyping. However, that second point is less important than the first, so I have emphasized it.

katyhuff:
@wfspotz : Thank you for your close reading. I appreciate that you caught a few of my copy-paste issues and pointing out I was missing a part of the paper I'd thought about adding (results!) Regarding your specific comments: 

1) I had gone back and forth about whether to include numerical results in this paper, which is focused more on the design of the code and is so limited in length. Your comment (pointing out a residual comment in the siminfo section about output) convinced me to include an example case that was sufficient minimal as to be understood by a non-nuclear-engineer, but sufficiently interesting to demonstrate the way the code works. I’ll add a brief illustration of that minimal case tomorrow.

2) 1+j+k is, indeed, what I meant. Good catch! Fixed.

3) First sentence of the THSystem section, and indeed the rest of the section, has been edited significantly to fix editing errors.

katyhuff:
@lederernc : Thank you for your review comments! I appreciate your close reading. 

1) You were right that the last sentence of Background Neutronics section did not make sense. I believe I have fixed it by rewording a bit and changing the verb tense.

2) You said the background sections for Reactivity and The PRKE need a lot of work, so I went through them to improve clarity, sentence simplicity, and grammar. I will go over those sections again tomorrow morning with fresh eyes. You also stated that rho and p are presented inconsistently. I think this is largely because p was previously undefined - an oversight. I have added a definition of p for clarity.

3) You said the equation with \alpha_i appears to be missing a summation or product symbol. You were correct! It’s a sum over all components, i. Fixed.

4) You said Equation 5 appears to use three different lambdas and the two lower case ones are not distiguishable as they are in equations 10 and 14. That is true. It is unfortunately necessary to talk about three different lambdas in this particular set of equations. These are the canonical ways to refer to that data, so we’re somewhat stuck with lots of lambdas. However, I have made these more clear with subscripts that are now consistent everywhere (and by virtue of being subscripts should no longer be confusable with exponents). The variable definitions now indicate that lambda_{FP,k} represents fission product decay from group k and lambda_{d,j} is used for indicating delayed precursor decay from precursor group d. Neither of these should be easily confused with capital \Lambda. I think this is thus much improved. Thank you for pointing it out.

5) You pointed out that the PyRK Implementation section on Neutronics contains duplication in the first and last paragraphs. Thank you! That is correct, and was an editing error. It has been fixed to summarize the key facts without redundancy.

6) You pointed out that the first sentence in the THSystem paragraph does not read well in English. That’s quite true. Thanks. I have fixed it to be more clear.

7) You also pointed out that there was a lot of clean up needed in the markup of that section, e.g. gls{PBFHR}, gls{TRISO}, LaTeX throughout, and that the word radiation is misspelled before the equations for Q. Thanks for pointing this out! I believe I have now fixed all misspellings and all instances of my latex draft leaking into my rst submission!

wfspotz:
@katyhuff: Your description of your approach to showing results sounds exactly like what I was looking for.

katyhuff:
I finally got a chance to put in the minimal example. I hope it's somewhat clear. Sorry for the delay @wfspotz !

wfspotz:
@katyhuff: I think the minimal example helps a lot.

I noticed that a caption reads "The reactivity insertion that can drives ..."

katyhuff:
nice catch @wfspotz . All fixed. I think that covers all reviewer suggestions @jaberg !

PR #163, opened 2015-06-16
Include Bibtex Bibliographies

michaelpacer:
This should be a pull request that allows local .bib files to be included in the document as external citation repositories. 

In the .rst file, include a bibliography(e.g., `myfile.bib`) at the top by including:
```
:bibliography: myfile
```
Do not include the `.bib` extension and rename the file to not include and characters that need to be escaped or have spaces, as this can cause issues for bibtex properly finding the appropriate file.  

To use a citation from this bibliography, use the `:cite:` role followed by the citation-key for the citation as described in the .bib file. The citation-key will need to be surrounded by backticks ( \`citation-key\` ) to be recognized properly.


katyhuff:
I think this is a really helpful and useful contribution, but it's super polluted with other commits. I can't bring myself to add this history in this way. In any case, it should be Pull Requested against the master branch so that it's available in future years.

PR #168, opened 2015-06-23
Adds bibderwalt to the excludes list to avoid server build conflicts

michaelpacer:


katyhuff:
gonna have to restart the server. let me know if it doesn't get fixed.

michaelpacer:
Interestingly, it builds the right paper again, but doesn't seem to update to the most recent commit.

An easier fix might be to just change the name of the `bibderwalt` file to finish with `vanderwalt`(that worked earlier for repo's build), but I'm confused as to why what's happening is happening.

michaelpacer:
Ok — it's back to thinking that bibderwalt is a valid submission again — not sure why unless the excludes catch isn't working.

michaelpacer:
Aha — I think I know what the problem is (though I'm not sure). When I try to run the make file to make the entire proceedings, then it properly excludes the `bibderwalt` folder (as it should), but I can still individually use the make_paper shell script to build the `bibderwalt` paper. My guess is that when files are individually compiled, that it runs that script and does not rebuild the entire conference proceedings (understandable) but then that code doesn't exclude all of the things listed in `conf.py`. Is the server code stored in a separate repo? I think that the issue might have to do with that and not the scipy_proceedings repo per se.

katyhuff:
yup. I fixed and restarted the server. Unfortunately, you're now getting a build error, but at lest it's the right paper. 
```
:299: (ERROR/3) Too many autonumbered footnote references: only 4 corresponding footnotes available.
```

michaelpacer:
Interesting — so I got the same error when I ran the makefile locally, fixed it, pushed the fix, and now the server doesn't seem to recognize the newer commit. 

I hope restarting the server isn't necessary again…I'm sorry to have caused so much trouble :(. 

Still, my guess is the next time the build server rebuilds the entire proceedings it'll pick up the right thing.

katyhuff:
@michaelpacer - try refreshing your browser.? your paper seems to be building now. (note that you have to press the sync button to make the server rebuild your paper).

image

katyhuff:
(also, note that this thing you keep mentioning about building the entire proceedings isn't a thing. It happens once, right before we publish... it never happens on the server. the server always builds the papers individually. there is definitely a script that does them all individually all at once, but it's not building them together as a single thing... hope that's clarifying. )

michaelpacer:
I had been pressing the sync button, and syncing doesn't help. It's still
linking to the pdf of is the `terraforming the desert` bibtex example (not
my actual paper). It's bizarre since it failed to build it before but with
was an error specific to the mike_pacer tex file, but now it builds but
returns the pdf that it's supposed to never make.

michaelpacer:
Ah — sorry about that. I just assumed that the uniform times of all the
papers suggested that unless someone went in and manually rebuilt their
paper that it was just occasionally rebuilding them all in accordance so as
to keep them up to date. The only way it seemed to get the `conf.py` to run
when doing it locally is to run the make file that makes the entire
proceedings, so I just assumed the server was relying on that to build all
of the papers, and then if anyone clicked the sync button then it would run
it individually on their paper according to which one had been built
according to the global make file.

I probably got this misunderstanding because the server's build code is
separate from the rest of the repo, and so what applies to my local build
won't apply to the server's build.

katyhuff:
@michaelpacer I think you have a browser side caching error. Please go to this url and **refresh your browser** : http://nipy.bic.berkeley.edu:5000/download/12 .

katyhuff:
>  I just assumed that the uniform times of all the
papers suggested that unless someone went in and manually rebuilt their
paper that it was just occasionally rebuilding them all in accordance so as
to keep them up to date. 

That's from when I have to restart the server. ;)

michaelpacer:
Interesting — I had hard refreshed the page on chrome (⌘ + ⇧ + R), and thus
I had thought that *was* refreshing the browser.

just opened safari and you're right, it loaded just fine. Weird that
sometimes the hard refresh seems to work and sometimes it doesn't (e.g., it
did work earlier today — at least that is my understanding).

Thanks for putting up with my mistakes; fortunately they continue to be
correctable once I understand what's going on!

Issue #144, opened 2015-06-15
Setting last author as corresponding author?


adamrp:
Currently, whatever author is listed first is the corresponding author. However, we would like to use the *last* author specified as the corresponding author. Will that be possible?

katyhuff:
Let me look into it.

katyhuff:
Ok, I think I can fix this, but not today. If you're willing, please go ahead and make your PR and I'll try to address this issue before the proceedings are published.

adamrp:
Great! I will make the submission now, and fix this later if possible. Thanks again @katyhuff!

jackiekazil:
I was going to ask something similar. Is it possible to get both authors to be corresponding authors? (My colleague is awesome, but I don't want either of us to be a single point of failure.)

katyhuff:
@jackiekazil good idea. I'm sure it's possible, but I'm putting out too many fires this week, so if you wait on me it'll have to wait until shortly before publication. However... if any intrepid pythonista wants to look into it..  here's the offending line of python:  https://github.com/scipy-conference/scipy_proceedings/blob/master/publisher/writer/__init__.py#L127

jackiekazil:
No worries. If I wasn't doing polishing / updates of Mesa code for Scipy I
would totally dig into the code.

katyhuff:
@jackiekazil asked in #133 :+1: 

> Is it possible to include two email addresses or set a group email address as oppose to having this go to one person?

@jackiekazil :  Which of these functionalities do you want? I'm only going to have time to do one of them before the proceedings happen.

jackiekazil:
I would two emails might be the easier of the two. However, we did a quick hack in the paper, where we listed both emails under the first author. So, if that works for you all, I would say that I am satisfied.

katyhuff:
ahh, I see what you're saying. Great!! I'll still try to resolve this issue
at some point soon to help with @adamrp and the Qiita team.

On Fri, Jul 3, 2015 at 4:21 PM, Jackie Kazil <notifications@github.com>
wrote:

> I would two emails might be the easier of the two. However, we did a quick
> hack in the paper, where we listed both emails under the first author. So,
> if that works for you all, I would say that I am satisfied.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/scipy-conference/scipy_proceedings/issues/144#issuecomment-118437109>
> .
>



-- 
http://katyhuff.github.com

[...]

PR #171, opened 2015-07-07
Footmarks

katyhuff:
This allows for the following options to any (and, indeed, many) to be added to any author definitions in the paper header:

```
:corresponding: True
:equal: True
```

- these are optional, 
- the default is to have the first author be the corresponding author 
- and the default is to have no indication of equal authorship

However, if you want two authors to have a footnote indicating that they contributed equally to the paper, include the `:equal: True` tag to each of their area of the header. 

If you want one or more people other than the first to be the corresponding author, put `:corresponding: True:` in their area of the header.

It would be great if @stefanv could review this. Note that it seeks to solve #144 desired by @adamrp, @jackiekazil, and @philippjfr .

### DIFF
--- a/papers/00_vanderwalt/00_vanderwalt.rst
+++ b/papers/00_vanderwalt/00_vanderwalt.rst
@@ -2,6 +2,7 @@
 :email: jj@rome.it
 :institution: Senate House, S.P.Q.R.
 :institution: Egyptian Embassy, S.P.Q.R.
+:corresponding: True
 
 :author: Mark Anthony
 :email: mark37@rome.it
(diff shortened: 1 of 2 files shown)


philippjfr:
This is great! Thanks for implementing this so quickly.

stefanv:
@katyhuff Looks good apart from the equal auth template not being used, and the first few characters of the __init__ file being deleted.  Also, it may be a bit cleaner to simply use ``:corresponding:`` instead of ``:corresponding: True``.  Perhaps we can also make the `:equal:` field a bit more descriptive: `equal-contributor` or somesuch.  I guess I wonder what it says about the other authors on the paper if you only mark a few as equal contributors, but perhaps we can just leave that up to paper authors.

stefanv:
I made a PR with the ``:corresponding:``-without-value suggestion.

katyhuff:
Fixed the part where I forgot to use the equal author template!

stefanv:
Looks good to me; let me know how you feel about the other recommendations--happy to merge.

katyhuff:
Whoops @stefanv , I just didn't read your comment all the way through and missed the other recommendation!

It's now :equal-contributor: rather than :equal:
Doesn't require True.

PR #173, opened 2015-07-12
A few edits to intro

kjordahl:
@katyhuff Here are my edits to the intro. Sorry it took me so long. I lost the commit history, but branching directly off yours had a lot of extraneous commits.

Happy to comment on changes here.

### DIFF
--- a/papers/00_intro/intro.rst
+++ b/papers/00_intro/intro.rst
@@ -27,11 +27,11 @@ SciPy 2015, the fourteenth annual Scientific Computing with Python
 conference, was held July 6 |--| 12th in Austin, Texas. This community, focused
 on the use and development of open source scientific software in Python, has
 grown enormously since its origins at the first SciPy conference in 2002 at the
-California Institute of Technology. This growth is owed in no small part to the
+California Institute of Technology. This growth is due in no small part to the
 enormous growth in support from industrial, government, and academic partners.
-However, the success of this community is a testament, first and foremost to
-the individual attendees, whether returning for their 14th year or attending
-for the first time, who provide the science, enthusiasm, and inspiration on
+The success of this community is a testament to the individual
+attendees, whether returning for their 14th year or attending for the
+first time, who provide the science, enthusiasm, and inspiration on
 which the community thrives.
 
 |noindent|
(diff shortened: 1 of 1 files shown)


stefanv:
I guess the intro needs to be at least partially rewritten for 2015 from

https://github.com/scipy-conference/scipy_proceedings_2014/blob/master/papers/00_intro/intro.rst

?

kjordahl:
@stefanv Mmm, yes. Sorry, I missed how much of this had come from the previous one. I'll take another crack at it in this PR.

katyhuff:
Yeah, sorry about that. My fault.

katyhuff:
So, this is now the final blocking issue. As soon as this is ready, we'll be able to publish the proceedings. A couple things about that. 

- I've reworded a few things, in my own words, that provide some reflection on the conference. Sorry for the copy paste nature of the previous template. @kjordahl , it would be great if you could look that over and see if it is ok with you.
- Also, it would be great if we could add some more specific data. Apparently the proceedings are one of the few places where we record, for posterity, statistics about the conference attendance and demographics. It would be nice if those statistics were added to the proceedings before they're published. Ideally, it would have the actual number of registrants, the number of folks attending the tutorials, the percentages of women and/or minorities, the number of talks, and maybe even stuff like the number of scheduled bofs.

Thoughts on that?

kjordahl:
@katyhuff Thanks! I've added a few edits and filled in some numbers. The number I have for registrations is 604, but I left it at "over 600" because I'm not positive that is the final true number. I was similarly vague for tutorials and sprints - we should have a firm number for tutorials, because they  were paid registrations, but for sprints at some point we had 215 (IIRC) signed up, but I don't think we had that many actually attend.

There are still two placeholders for numbers: number of countries and % new attendees. I believe the numbers are ~15 and ~60%, but I will double-check these tomorrow.

I didn't put it in, but there were 14 BOFs on the final schedule, plus a few informal ones that happened spontaneously.

Demographics are tougher; we don't have real data on this. In previous years we have counted feminine names by hand, I believe. It's worth doing, but I don't know if it needs to be in the proceedings. Thoughts?

kjordahl:
^ Updated with number of BOFs

katyhuff:
Excellent!!! Thanks so much Kelsey. That's enough to go with I think. Thanks!! I should have a chance to merge this intro and do the final once-over sometime this weekend. I realized last night that I also need to add links to the talks, now that they're all up. Hopefully @jaberg and I can even get it all up on the arXiv, like EuroSciPy does...

PR #184, opened 2016-05-23
Labelchanged

rainwoodman:
I need this fix to compile my paper. Please consider merging into 2016 branch. Thanks.

### DIFF
--- a/publisher/build_paper.py
+++ b/publisher/build_paper.py
@@ -103,6 +103,23 @@ def rst2tex(in_path, out_path):
 
 def tex2pdf(out_path):
 
+    # Sometimes Latex want us to rebuild, because labels have changed.
+    # but we will only try at most 5 times.
+    for i in range(5):
+        out, success = tex2pdf_singlepass(out_path)
+        if not success:
+            # building failed, bail.
+            break
+        if "Latex Warning: Label(s) may have changed. Rerun" in out:
+            # Latex tells us to rebuild.
+            continue
+        else:
+            # Latex has reached to an equilibrium, we are done
+            break
+    return out
+
+def tex2pdf_singlepass(out_path):
+
     import subprocess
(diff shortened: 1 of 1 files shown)


rainwoodman:
@stefanv -- This is more pressing than the package name stuff..

publisher/build_paper.py
```diff
@@ -144,25 +155,16 @@ def tex2pdf(out_path):
...
                 )
-        out, err = run.communicate()
+        out1, err = run.communicate()
 
         if err:
             print("Error compiling BiBTeX")
-            return out
+            return out1, False
 
+    if "Label(s) may have changed." in out:
```
stefanv: Why use out and out1 separately here?  Should we not check the second run to see if labels changed?

rainwoodman: If bibtex has died, False would terminate the run, and out1, the error message is returned.

If bibtex has succeeded, we do not care its output. Whether to rerun is encoded in the output of pdflatex.

rainwoodman: The second / third / ... runs will be issued by another call to tex2pdf_singlepass -- which will check the output again.

stefanv:
I've read this patch again, and I don't think the logic is correct.  When a label change is detected, the build should return a failure, upon which you want to retry.

rainwoodman:
I don't think the logic is incorrect; but the variable name was poorly chosen. 

The boolean indicates whether to retry. IF label has changed we need to retry. Otherwise we can break the loop. 

I have renamed the variable name from 'success' to 'retry'. Does this make more sense to you now?

stefanv:
Thanks, Yu, I merged a slightly modified version.

rainwoodman:
Thanks! As a reminder, could you merge this to 2016 branch if haven't done so?

stefanv:
I did both at the same time.

PR #190, opened 2016-05-25
Proposal: Change default BibTeX style from alpha to alphaurl.

Chroxvi:
This is a proposal to change the default BibTeX style used in the SciPy proceedings to `alphaurl`.

The `alpha` BibTeX style currently used in the SciPy proceedings does not support the `doi` and `url` BibTeX fields. The [suggested review criteria](https://github.com/scipy-conference/scipy_proceedings/blob/master/review_criteria.md) as well as major publishers (e.g. [Elsevier](https://www.elsevier.com/journals/neurocomputing/0925-2312/guide-for-authors) and [IEEE](http://www.ieee.org/documents/ieeecitationref.pdf)) encourage the use of the DOI. Thus, the BibTeX style used in the SciPy proceedings should be changed to one that supports the `doi` and `url` fields.

The `alphaurl` BibTeX style from the [urlbst](https://www.ctan.org/pkg/urlbst) project is a minimum modification of the `alpha` style to support the `url` and `doi` fields. 

An example of a reference using the `alpha` style:
opa14_alpha

The same reference using the `alphaurl` style with a DOI:
opa14_alphaurl

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,11 @@ examples of how to:
 ## Requirements
 
  - IEEETran (often packaged as ``texlive-publishers``, or download from
-   [CTAN](http://www.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/) LaTeX
+   [CTAN](http://www.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/)) LaTeX
    class
  - AMSmath LaTeX classes (included in most LaTeX distributions)
+ - alphaurl (often packaged as ``texlive-bibtex-extra``, or download from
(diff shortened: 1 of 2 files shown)


stefanv:
I am +1 on this change, but it may require a change on the build bot, so I'm going to wait for @katyhuff to acknowledge before I merge.

katyhuff:
I am also +1 on this change! @sarostru or @sbenthall may actually be running the buildbot from their user accounts on zibi... I'll double check.

katyhuff:
We're good to go.

PR #193, opened 2016-05-27
Paper: Launching Python on HPC

rainwoodman:


mrocklin:
Generally this paper appears to be well laid out, well written, and addresses a longstanding problem in using Python on common HPC systems.

*Disclaimer: I have a slight conflict of interest.  I am affiliated with Continuum Analytics, which has a vested interest in the `conda` technology that is used and suggested within this paper.*

### Some slight areas for improvement

*  Consistently capitalize `python`
*  Redesign Figures 3, 4.  These appear to be the main performance result of this paper and yet take some time to understand.  It might be worth considering stacked bar charts (although the logarithmic scale makes this difficult) or breaking out the different timings into side-by-side plots.  Generally I would consult someone with more visualization expertise (not me) for advice here.  These are nice results and there is probably a way to convey them more effectively in an image using modern techniques.
*  Assess the current costs and how they might be reduced.  I may have missed this, but why do `import scipy` costs jump to minutes long under more cores?  Would turning of gzip compression on the tarballs reduce or increase costs?  What are the next bottlenecks to consider in future work?
*  Perform a direct comparison with previous work or at least mention and compare their stated quantified results with the results shown here.

### Two conda questions

 (and again, note disclaimer above)

1.  Did you run into any problems with hard links?  In doing something similar on HDFS clusters we've had to use the `--copy` command when packaging up conda environments without hard-links.
2.  Thoughts on just using conda for all tiers?  At the end of the day conda just moves around binary tarballs.  Merging these tarballs into a single environment is fairly straightforward.  This would require some effort to shield the user from the pain of making conda packages, but presumably you're doing something like this already.

rainwoodman:
@mrocklin Thanks for the comments.

I will address the rest gradually in the text Here are my thoughts over conda:

1. At NERSC the user defined environments and conda root env are on different drives, so hard link isn't really an option (unless we clone a user defined environment)
I've thought about problems with symlinks, but haven't run into yet. I would imagine if we ask the tar command to follow symlinks, then there should be no problems, regardless how the environment is created.

2. It seems to be more complicated because we do not have write access to the root environment. Does conda provide a command that creates a partial/patch environment image from source code files based on the current environment? Then we can probably directly switch to use conda instead our silly bash scripts that are based on pip -- they work but they are a maintenance burden.

mrocklin:
I believe that a conda environment is just a directory of binaries.  If this is true then "patching" would just be copying the contents of one directory into another.

bcdarwin:
The paper suggests an interesting approach to reducing Python startup overhead on HPC systems which takes advantage of the Python interpreter's configurability to provide convenient run-time usage mechanisms.

I agree with the first reviewer that benchmarking or pointing to existing benchmarks for some existing solutions would be quite useful (even potentially quantifying, e.g., the build/deployment time for containers, if your contention is that ease-of-use is the main factor here).

In para 3 of "Previous solutions", you seem to suggest that Scalable Python falls into the class of "compile everything into one binary", but that doesn't seem to be true (they provide a patched interpreter which uses MPI when `import`ing).  In general, this section could make it clearer as to the different categories you've identified of existing solutions and which previous attempts fall into which -- but you might already know that judging from the "Clean this up!!!" in that section ...

Since the papers are supposed to be accessible to a varied audience, perhaps define 'rank' (which has a rather strange definition in MPI) the first time it's used rather than requiring the reader to read down to "1,000 instances or MPI ranks".

The second (and to some extent the first) paragraph of your conclusion doesn't really add much; it could perhaps restate what you see as the principal differences of this approach.

A couple minor typographic issues: 'interpretive' -> 'interpreted' in intro/conclusion; 'staticly' -> 'statically'.

michaelpacer:
Wanted to let you know that I was assigned to review this paper. 

*Disclaimer, I have a slight conflict of interest as I am a member of **BIDS** where both of the authors are positioned, and will soon be officially employed by BIDS. However, I am unaffiliated with anything directly related to this project, so I do not believe this would affect my judgment.*

This is a well-written and well designed paper. 

The major issues I have with it are regarding the intended audience, which I'm assuming was only supposed to be those who already have successfully deployed python applications on HPC servers. From that perspective, the paper does not use space to define concepts that would be basic to someone with appropriate experience.

If this is meant to also be an introduction to the techniques described I think that a decent amount of refactoring may be in order. As mentioned by @bcdarwin, basic concepts from parallel computing and message passing (e.g., "ranks" and their relation to task scheduling and network size) are not even given cursory definitions. The same goes with the notions of (shared vs. local) file system and load. If you are aiming for a general audience, please let me know as I will go into much greater detail about many of the parts of the paper that could use a more thorough introduction. I get the impression that you aren't, in which case it is not useful for me to do so.

However, I don't think that this needs to be directed at a general audience. Explaining the basic concepts of parallel task scheduling, high performance computing, file systems, &c. could require much more than 8 pages, even if you only focused on python and python-mpi. More importantly, someone probably should not presume that they can learn the entirety of what is needed to deploy python applications on a peta-scale system from a single 8-page paper. This is especially the case for an 8-page paper that is not intended to be only an introduction to these ideas, but to describe a nice new solution to a long-standing problem.  It's also worth noting that it would be unlikely that a novice would be *capable* of implementing any application that used any MPI technologies (including `python-mpi-bcast`) without guidance, and I imagine the authors would not encourage novices to do so given the shared nature of these resources. And, to be fair, a real novice would be unlikely to have unassisted/unguided access to a hpc system for which `python-mpi-bcast` would be relvant. 

That said, your paper is fairly readable – I would imagine – even for someone not well versed in implementing software on massively parallel systems. Rather than be expected to expand your paper to meet a broad audience, you could consider giving the audience that cannot follow your work an avenue to learn how to do so. Currently, you do not provide any recommended resources as good introduction to the ideas basic to understanding your work. This would include: what are ranks and how do they relate to nodes (and task scheduling), meta-data requests versus data requests, when local replication of data is useful or harmful, what kinds of information would be included and received in a file-system request, when is an operation a file-system request and what is it when it is not, why does flooding the file-system with requests halt the system (i.e., what load is and how it works), what is the job manager node, what is the standard application launch process, why does a long PYTHONPATH negatively impact launch performance, what is static versus dynamic linking and how does it interact with any MPI based solution, &c. ?

Some additional questions:

* Does this work with non-conda based workflows? What would that look like?

* Could you provide a heuristic criterion (or a boundary case) for moving something from Tier 3 to Tier 2? What are the relevant packaging steps and are there ways to develop software that will make that transition easier? Are there any things that can be expected to never go into Tier 2?

* Did why are there erratic jumps in the curves for the benchmarks? Relatedly, did you only benchmark once? Given your knowledge that distributing python applications can be highly variable in the time they take to run, wouldn't it be more appropriate to sample many times and provide a distribution of the run-times across various system loads rather giving a point estimate? You could monitor the system load and directly state the conditions in which `python-mpi-bcast` fails to be robust.

* The symbol you should use for describing your `x` should be `×` as it's more properly a "cross" than an "ex".

* Rather than `bcast` being a circle, if it is the composition of `bcast/MPI_Bcast` and `bcast/tar_xzvf` then you might want to use `⁕`, which suggests the composition of `×` and `+`. 

* Why was the reduction in operations between Python 2 and Python 3 not reflected in performance improvements later on? If it does not matter why doesn't it matter? If it would have mattered but `python-mpi-bcast` made it not matter, why is that? Why didn't you benchmark with Python 3 as well (i.e., if they were different before, in what way does `python-mpi-bcast` treat them the same)?

* Given that file-system load is a stochastic process, could you not also expect there to be an interaction between computational operations you are running and the runtimes for the solution you provide? Are all of the broadcasting and file-system operations guaranteed to occur at the beginning of the process, and if they are not, wouldn't a test suite that takes into account computation related delays (rather than just file-system and import related delays) be more appropriate for hpc applications?

Have your two figures span the columns on a single page, it will be easier to read them.

"which will does not" → "which does not" or "which will not"

wfspotz:
I like this paper very much. It is clearly organized and written, and as I consider it against the SciPy review guidelines, it fulfills the requirements admirably. I have the following short list of comments:

* The uncompiled length that I printed was over 9 pages long, and that does not include the figures. The paper is supposed to be 8 pages long, including figures, so a little bit of work is required to get it down to the length restriction. I would suggest shortening the "Previous Solutions" and "Conclusions" sections (the first two paragraphs of the conclusion are unnecessary), and making a pass through the paper looking for redundant wording. The authors sometimes repeat themselves, and this is a good place to help shorten the paper. The figure captions can be shortened considerably as well.
* The term "stable" is used in the "Three-tiers of bundles" section, and it would have been helpful to me if this would have been defined to refer to frequency of changes in the code.
* The "Tier 3 components" section, second paragraph, contains the phrase "which will does not," which I do not think the authors intended.
* In the "Benchmark and Performance" section, the third and fourth paragraphs employ the term "capability" but I think "capacity" is what the authors intend.
* In the first paragraph of this section, the authors say that "Python 3 and Python 2 are treated the same by `python-mpi-bcast`" which I take as a reference to the fact that `python-mpi-bcast` is compiled C-code run under the shell prior to invoking a Python script. However, the authors show in this paper that Python 3 employs a more efficient import system, and so we would expect the import phase to be twice as fast. Is it still? If so, we don't need a figure, but a sentence to this effect would be warranted.
* As the lead developer of PyTrilinos, a more recent, and better, reference is

> William F. Spotz, "PyTrilinos: Recent advances in the Python interface to Trilinos," Scientific Programming, v. 20, 2012, pp. 311-325.

@michaelpacer, as this is a conference paper, I do not think it is intended to provide an introduction to general techniques. Since it is over the length limit already, there is little hope of it becoming that. I do hope that a full length journal article with more detail is in the future.

sarostru:

rainwoodman:
This is indeed extremely nice of you! Thanks.

wfspotz:
Looks good to me. I accept the changes for merging.

rainwoodman:
@wfspotz thanks!

rainwoodman:
Thanks for the reviews! Here is a summary of our responses:

The current PDF version is 7 pages.


Several reviewers commented on interaction with conda.  The implementation here is a proof of idea, and we simply picked the easiest tools at our disposal (bash scripts). There is no reason why these strategies cannot be reimplemented as an extension to conda. We added a pointer sentence regarding this at the end of the paper.

Several reviewers commented on the figures. We tried to enhance the figures with suggests from @michaelpacer and @mrocklin: colors are added, and symbols are tweaked. It is unfortunate the figure is in log scale on both axes and a stacked bar chart requires a linear scale in y, which cannot reasonably present the dynamic range of the data, and splitting will hide the correlation.

We added a reference to a standard parallel programming text book for the basic concepts in HPC. We agree with @michaelpacer that a novice user is likely unable to write MPI applications, thus benefit from our methods. Unfortunately, developing an tightly coupled distributed MPI application requires a slightly different programming mindset. It is a barrier that we cannot address in this short paper.

The Previous solution section has been reorganized to give it a better flow.  @bcdarwin Also the reference to ScalablePython is corrected to mention it falls into the same user space import hook caching catagory.

We do not perform additional full system benchmarks with Python 3 as the reduction of local file operations PEP 302 is not a main focus of this work. We do have reasons to expect it to be quite similar, though faster to Python 2 numbers (which are already quite fast). This is now clearly noted in the text. 

As for build time of reproducible containers, we added the build time of an example package that we use (pfft-python, 10 minutes). We also mention that tarring an anaconda to from a Tier 1 bundle takes more than 5 minutes. 

We would like to include Figure 3 and 4 as one single panel, but apparently restructured text does not allow including two images in one figure object. Given that the currently length of paper is below the limit, we have not invested to workaround this limit of restructured text.

Erratic jumps: performing a controlled experiment on a HPC system is beyond the amount of resource in our reach. At NERSC there are many metrics but typically measured much sparser than the run time of our test scripts. We are not aware of any easy way of tracking the metrics in BlueWaters. Practical users of HPC systems shall be rather familiar with the probabilistic model we described in the text for explaining these jumps.

Contamination from payload: There is an explicit barrier between bcast and launching the interpreter.  Therefore the payload application does not affect the time of bcast; the correlation has to be due to a shared status of the system.

Head to head comparison with previous solutions: We do not include a head to head speed comparison with previous solution because this is not the main point of this work -- every approach can be made very fast. (We never claimed python-mpi-bcast is the fastest) Our main point is that launching applications on HPC can be done very simply (probably much simpler than one had thought) and in a way that interacts naturally with the developing cycle.

bcdarwin:
Looks good, +1 to merge.  Congrats!

sarostru:
@michaelpacer @mrocklin have you had a chance to review the updates?

mrocklin:
+1

sbenthall:
Ping @michaelpacer can you please respond?

michaelpacer:
Looks good to me  (it did before mod the formatting stuff was fixed and the audience issue was decided upon). Good to go; Well done!

PR #198, opened 2016-05-30
Paper: UConnRCMPy: Python-based data analysis for Rapid Compression Machines

bryanwweber:
Please let me know if I should remove the text files etc. that are only necessary to generate the figures.


dschult:
Hello Bryan,
My name is Dan Schult and I am a reviewer for this submission. I have never done a Github-based review before so please bear with me as we make our way through this process. I am quite excited to see how it works because the potential is really nice. :)

The manuscript looks really good already. My comments are relatively minor. I will try making may of them via line-comments, though I guess that will create lots of emails. I'm guessing that it will also be good to keep the comments separate from each other so we can keep track of which have been addressed. At first I thought I would be making Issues and PRs to your personal branch, and I guess I could do that. But it seems set up for me to make comments and then you address them via further commits to this PR.  Let me know if there are better ways I could be giving feedback.

dschult:
My main comment is that it is not consistently clear which parts of the process are done directly by the user and which are done by the software. Also, for the parts done by the user, which involve interacting with the RCM directly and which by controlling the RCM through the Jupyter notebook interface?  The best presentation of the user workflow is in the "Usage Example" section (which is appropriate). Perhaps you can change to an "active voice" from the passive case in that section to make it clear who is doing what. 

One example is: "From the runs at this condition, it is determined that the reference experiment..." Is that determination done by the software and reported somehow? Does the user make this determination by looking at the figures?

At the end of this paragraph, "the first non-reactive experiment is added to the Condition" could be: "we then add the first non-reactive experiment to the Condition".  In this case, I'm pretty sure that the user must type the next python command, but using an active tense makes it explicit.  

Also, it is not clear whether the non-reactive experiment was performed after processing the results from the previous step, or were all experiments performed before coming to the computer.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+the pathways present in the model. There are several methods to measure the ignition delay at
+engine-relevant conditions, including shock tubes and rapid compression machines (RCMs).
+
+An RCM is typically designed with one or two pistons that rapidly compress a homogeneous fuel and
+oxidizer mixture inside a reaction chamber. After the end of compression (EOC), the piston(s) is
+(are) locked in place, creating a constant volume reaction chamber. The primary diagnostic in most
+RCM experiments is the pressure measured as a function of time in the reaction chamber. This
+pressure trace is then processed to extract the ignition delay.
+
+In this work, the design and operation of a software package to process the pressure data collected
```
dschult: "In this work" sounds funny...   "In this paper", "In this presentation", ... ??

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+As mentioned previously, the primary diagnostic on the RCM is the reaction chamber pressure,
+measured by a dynamic pressure transducer (separate from the static transducer used to measure
+|P0|). The dynamic transducer outputs a charge signal that is converted to a voltage signal by a
+charge amplifier. This system measures changes in pressure in the reaction chamber (as opposed to
+the absolute pressure measured by the static transducer) and as such, has a nominal output of 0 V
+prior to the start of compression. In addition, the output range of 0 V to 10 V is set by the
+operator to correspond to a particular pressure range by setting a "scale factor". Typical values
+for the scale factor range between 10 bar/V and 100 bar/V.
+
+The voltage output from the charge amplifier is digitized by a hardware DAQ and recorded into a
```
dschult: DAQ should be written out when first used with (DAQ) trailing.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+factor from the charge amplifier. However, as can be seen in Fig. :ref:`raw-voltage`\ b there is a
+small offset in the initial voltage relative to the nominal value of 0 V. To correct for this
+offset, it can be subtracted from the voltage trace
+
+.. math::
+    :label: corrected-pressure-trace
+
+    P(t) = F \cdot \left[\overline{V}(t) - \overline{V}(0)\right] + P_0
+
+where :math:`\overline{V}(0)` is the initial voltage of the filtered and smoothed signal. The result
```
dschult: Since smoothing treats the initial values specially, would it be better to offset the trace at, say, the 50th point rather than this initial point? Of course, you'd need to pick a point before any action has started and after the smoothing "end-effects" are gone.  [no correction needed... just something to think about]

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+and shortly after the EOC can be excluded to avoid the peak in the derivative around the EOC.
+
+In the current version of UConnRCMPy :cite:`uconnrcmpy`, the first derivative of the experimental
+pressure trace is computed by a second-order forward differencing method. The derivative is then
+smoothed by the moving average algorithm with a width of 151 points. This value for the moving
+average window was chosen empirically.
+
+For some conditions, the reactants may undergo two distinct stages of ignition. These cases can be
+distinguished by a pair of peaks in the first time derivative of the pressure. For some two-stage
+ignition cases, the first-stage pressure rise (and consequently the peak in the derivative) are
```
dschult: "are" at end of line should be "is"?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+
+where :math:`c_v` is the specific heat at constant volume of the mixture, :math:`v` is the specific
+volume, :math:`u_k` and :math:`Y_k` are the specific internal energy and mass fraction of the
+species :math:`k`, and :math:`t` is time. For a constant-area piston, the rate of change of the
+volume is equal to the piston velocity. In UConnRCMPy, Eq. :ref:`first-law` is integrated by Cantera
+:cite:`cantera`.
+
+In Cantera, intensive thermodynamic information about the system is stored in an instance of the
+``Solution`` class. The ``Solution`` classes used in this study model simple, compressible systems
+and require two independent intensive properties, plus the composition, to fix the state. In
```
dschult: What is an "intensive property"? If that is a technical term, could you elaborate? If it is std english term then can you be more specific about what properties are needed?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+state information and evaluate thermodynamic properties. The difference between the ``Reservoir``
+and the ``IdealGasReactor`` is simply that the state (i.e., the pressure, temperature, and chemical
+composition) of the ``Solution`` in a ``Reservoir`` is fixed.
+
+Integrating Eq. :ref:`first-law` requires knowledge of the volume of the reaction chamber as a
+function of time. To calculate the volume as a function of time, it is assumed that there is a core
+of gas in the reaction chamber that undergoes an isentropic compression :cite:`Lee1998`.
+Furthermore, it is assumed that there is negligible reactant consumption during the compression
+stroke. Then, a Cantera ``Solution`` object is initialized at the initial temperature, pressure, and
+composition of the reaction chamber.
```
dschult: Does the user initialize directly, or does UConnRCMPy initialize these values?  Also you use the word "Then" to start the sentence. What does this come after?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+appropriate velocity when passed a time.
+
+The ``IdealGasReactor`` is installed into an instance of ``ReactorNet`` from Cantera
+:cite:`cantera`. The ``ReactorNet`` implements the interface to the solver CVODES. CVODES is an
+adaptive-time-stepping solver, distributed as part of the SUNDIALS suite :cite:`Hindmarsh2005`. As
+the solver steps towards the end time of the simulation, the state of the system is stored on each
+integrator time step, producing simulated pressure, volume, and temperature traces. Finally, the EOC
+temperature is recorded as the simulated temperature at the EOC.
+
+Two simulations are conducted using this procedure. In the first, the multiplier for all the
```
dschult: "are conducted" is not as clear as "UConnRCMPy conducts two simulations", or some other wording describing who/what conducts two simulations.  Perhaps a clause "for each ``Condition``" or something to indicate how often these two are conducted.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+reaction rates is set to zero, to simulate a constant composition (non-reactive) process. In the
+second, the reactions are allowed to proceed as normal. The temperature traces from the two
+simulations are compared to each other and the inclusion of the reactions does not change |TC|,
```
dschult: Here you give the result of the simulations (does not change |TC|). But that isn't guaranteed. Who checks? Does the software give a warning if they are not the same?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+validating the assumption of adiabatic, constant composition compression. Nonetheless, when
+conducting simulations to compare a kinetic model to experimental results, it is important to
```
dschult: This was confusing to me. I had to read multiple times. I thought we were simulating to find TC. But now we are comparing a kinetic model to experiments.  I think this is a separate thought -- a warning that if we do wish to make such comparisons then we better include stuff. If I've understood correctly, maybe this sentence should be its own paragraph? Or maybe rewording it can make it sufficiently clear.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+
+As can be seen in Fig. :ref:`ign-delay-def`, the pressure decreases after the EOC due to heat
+transfer from the higher temperature reactants to the reaction chamber walls. This process is
+specific to the machine that carried out the experiments, and to the conditions under which the
+experiment was conducted. Therefore, the rate of pressure decrease should be modeled and included
+in simulations that compare predicted ignition delays to the experimental values.
+
+To conduct this modeling, a non-reactive experiment is conducted, where |O2| in the oxidizer is
+replaced with |N2| to maintain a similar specific heat ratio but suppress the oxidation reactions
+that lead to ignition. The pressure trace from this non-reactive experiment closely matches that
```
dschult: closely matches  -->  should closely match       (to make clear that this is not a report of a result, but rather an expectation of the setup)

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+
+One step up from the ``VoltageTrace`` is the ``ExperimentalPressureTrace`` class. This class takes
+a ``VoltageTrace`` in the ``__init__`` method and processes it into a pressure trace, given the
+multiplication factor and the initial pressure. This class also contains methods to compute the
+derivative of the experimental pressure trace, as discussed previously.
+
+All of the information about a particular experiment is stored in the ``Experiment`` class. When
+initialized, the ``Experiment`` expects an instance of the ``pathlib.Path`` class; if none is
+provided, it prompts the user to enter a file name that is expected to be in the current working
+directory. Then a ``VoltageTrace`` is created, followed by an ``ExperimentalPressureTrace``. The
```
dschult: What file is this referring to? I'm guessing an output file from the RCM, but it isn't clear.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+storing the experiment.
+
+Two plots are optionally created each time a reactive experiment is added to the ``Condition``
+(plotting is controlled by passing a boolean argument ``plotting`` to the ``Condition`` when it is
+initialized). The plots use Matplotlib :cite:`Hunter2007`. The first plot is a cumulative plot of
+the pressure traces of each of the experiments that are added to the ``Condition``. The second plot
+is an individual plot for each experiment showing the pressure trace and the time derivative of the
+pressure trace.
+
+In general, for a set of experiments at a given condition, all of the reactive cases are run first.
```
dschult: by "are run first" do you mean processed by UConnRCMPy first, before any non-reactive data is processed? Or do you mean that the RCM is set up and data is collected for many reactive runs first and then the non-reactive data is collected.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+The experiment chosen as the reference experiment (i.e., the one whose ignition delay and |TC| are
+reported) for a given condition is the one whose overall ignition delay is closest to the mean
+overall ignition delay among the experiments at a given condition. Once the reference experiment is
+selected, non-reactive experiments are run at the same initial conditions as the reference
```
dschult: replace passive "are run" by active "The ``Condition`` class initiates non-reactive experiments at the same ..."

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+experiment. Non-reactive experiments are added to the ``Condition`` by the same ``add_experiment``
+method and UConnRCMPy automatically determines whether the experiment is reactive or non-reactive.
```
dschult: What happens if a non-reactive experiment is added before the reference experiment is chosen?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+the pressure traces of each of the experiments that are added to the ``Condition``. The second plot
+is an individual plot for each experiment showing the pressure trace and the time derivative of the
+pressure trace.
+
+In general, for a set of experiments at a given condition, all of the reactive cases are run first.
+The experiment chosen as the reference experiment (i.e., the one whose ignition delay and |TC| are
+reported) for a given condition is the one whose overall ignition delay is closest to the mean
+overall ignition delay among the experiments at a given condition. Once the reference experiment is
+selected, non-reactive experiments are run at the same initial conditions as the reference
+experiment. Non-reactive experiments are added to the ``Condition`` by the same ``add_experiment``
```
dschult: "are added"   --> active voice: "The user adds non-reactive experiments to ..."

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+from the default), as shown in Fig. :ref:`all-runs`. It also generates one figure per experiment.
+The figures showing each experiment look similar to Fig. :ref:`ign-delay-def`, but the non-reactive
+trace is not plotted and the EOC and ignition delays are not labeled.
+
+.. figure:: figures/all-runs.png
+
+    All of the runs at the example experimental condition. The legend shows the date and time the
+    experiment was conducted. :label:`all-runs`
+
+From the runs at this condition, it is determined that the reference experiment is the run that took
```
dschult: who determines?  Who dds to .yaml?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+experiment. Non-reactive experiments are added to the ``Condition`` by the same ``add_experiment``
+method and UConnRCMPy automatically determines whether the experiment is reactive or non-reactive.
+Adding a non-reactive experiment creates a figure comparing the pressure trace of the non-reactive
+experiment with the reference reactive experiment.
+
+When the user is satisfied with the agreement of the reactive and non-reactive pressure traces, the
+creation of the volume trace is triggered by running the ``create_volume_trace`` method of the
+``Condition``. This function goes through the process of converting the reactive pressure trace
+(before the EOC) and the non-reactive pressure trace (after the EOC) to a volume trace. The actual
+computation of the volume trace (as described previously) is done by the ``VolumeFromPressure``
```
dschult: "previously" is not very precise. Can you say "(as described in *Finding the EOC*)"?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+class. This class expects a pressure trace, initial temperature, and initial volume. First, the
+volume trace of the reactive (pre-EOC) portion is generated using the pre-EOC pressure trace, the
+experimental initial temperature, and an initial volume of :math:`V_0 = 1.0\,\text{m}^3`, as discussed
+previously.A temperature trace is also constructed for the pre-EOC pressure trace using the
```
dschult: similar "previously" comment

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+
+.. code:: python
+
+    class AltCondition(Condition):
+        def add_experiment(self, file_name=None):
+            exp = AltExperiment(file_name)
+            # Omit the plotting code...
+
+Then, the ``AltExperiment`` overloads the ``parse_file_name`` method of the ``Experiment`` class to
+parse the alternate format. The user must make sure to return the expected values as defined in the
```
dschult: The user doesn't return anything... :)   "The user must ensure the new ``parse_file_name`` method returns the expected values..."

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+flexibility in the data format and naming conventions.
+
+Conclusions and Future Work
+---------------------------
+
+UConnRCMPy provides a framework to enable consistent analysis of RCM data. Because it is open source
+and extensible, UConnRCMPy can help to ensure that RCM data in the community can be analyzed in a
+reproducible manner; in addition, it can be easily modified and used for data in any format. In this
+sense, UConnRCMPy can be used more generally to process any RCM experiments where the ignition delay
+is the primary output. Future plans for UConnRCMPy include the development of a robust test suite to
```
dschult: Perhaps "Future plans" should start a new paragraph.

dschult:
This comment may not be reasonable to influence this paper, but it suggests yet another possible method to find EOC times.  You could fit the initial portion of the data to an exponential and the middle portion of the data as a line. Then the EOC time could be the time where the exponential crosses the line. I suspect this would give slightly smaller EOC times than the max of pressure.  But the max of pressure is unlikely to be at EOC. The system will take some (short) time before the loss to walls overcomes the gain from recent compression. I don't know what the relative time-scales are so hard for me to make a guess at which is more accurate.

suchow:
Here's some advice from Proselint (http://github.com/amperser/Proselint), a Python-based linter for prose, with seven false alarms removed:

```
689:24: redundancy.garner Redundancy. Use 'plans' instead of 'Future plans'.
221:46: weasel_words.very Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be. Found once elsewhere.
```

bryanwweber:
@dschult Thanks for the comments! I'll wait for comments from @katyhuff before responding so that I can batch up the comments and send them to my coauthor for review.

katyhuff:
Will send comments ASAP. Sorry for the delay!

On Fri, Jul 1, 2016 at 6:14 AM, Bryan W. Weber <notifications@github.com>
wrote:

> @dschult <https://github.com/dschult> Thanks for the comments! I'll wait
> for comments from @katyhuff <https://github.com/katyhuff> before
> responding so that I can batch up the comments and send them to my coauthor
> for review.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/scipy-conference/scipy_proceedings/pull/198#issuecomment-229921840>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAYCq7Wu8eBC_NFilqWVP3QdoH-c-UrBks5qRPacgaJpZM4IpwYp>
> .
>



-- 
http://katyhuff.github.com

astrofrog:
@bryanwweber - I'm one of the other reviewers on this paper, and I'll now be going through and leaving some comments - sorry for the delay!

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+
+This generates a figure showing all of the experiments together (the axis limits have been adjusted
+from the default), as shown in Fig. :ref:`all-runs`. It also generates one figure per experiment.
+The figures showing each experiment look similar to Fig. :ref:`ign-delay-def`, but the non-reactive
+trace is not plotted and the EOC and ignition delays are not labeled.
+
+.. figure:: figures/all-runs.png
+
+    All of the runs at the example experimental condition. The legend shows the date and time the
+    experiment was conducted. :label:`all-runs`
```
astrofrog: I'm confused by the legend in the plot - I'm assuming what comes after the month is the time, not the year? :) If so, could you use a different format to remove the ambiguity?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+
+.. code:: python
+
+    from uconnrcmpy import Condition
+    from pathlib import Path
+    %matplotlib
+
+    cond_00_in_02_mm = Condition()
+    cond_00_in_02_mm.add_experiment(Path(
+        '00_in_02_mm_373K-1285t-100x-19-Jul-15-1620.txt'))
```
astrofrog: Just out of curiosity, why is the ``Path`` class needed here? Is it not possible to pass the path as a string directly?

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+    All of the runs at the example experimental condition. The legend shows the date and time the
+    experiment was conducted. :label:`all-runs`
+
+From the runs at this condition, it is determined that the reference experiment is the run that took
+place at 16:33. The filename of this run is added to the ``volume-trace.yaml`` file:
+
+.. code:: yaml
+
+    reacfile: >
+      00_in_02_mm_373K-1282t-100x-19-Jul-15-1633.txt
```
astrofrog: How does the package know about the YAML file? Is it a file that **has** to have this name and be in the current working directory?

astrofrog:
Apart from minor comments above, my only substantial comment is that I actually found it quite hard until the end of the paper to understand what UConnRCMPy actually does - the examples which start on page 7 are good, but until then it's not clear what the package provides. I think it would be beneficial near the start to give a high-level overview of the main features of UConnRCMPy, maybe formatted as a bullet point list.

bryanwweber:
@katyhuff OK no problem thank you for the update!

@astrofrog Thanks for the comments! As I mentioned before, I'm planning to roll all the comments into one response for my coauthor's sake, so I'll reply once I have all the comments.

katyhuff:
Again, I'm so sorry for the delay. 

## Written Quality

The prose is written in professional english and clearly communicates the narrative. That said, I do agree with @astrofrog that a lot of the punch isn't delivered until the end. Namely, UConnRCMPy delivers a tool to 'process' the pressure data from RCMs, however the goal of the process (to extract ignition delay) is buried somewhat between pages 1 and 7. This broader purpose is mentioned on page 1, but is subsequently somewhat lost in the weeds until the final demonstration section.

## Technical Content

The technical content appears sound and describes the application of python to a domain specific problem. The equations are well presented and clearly defined. I had a few questions about the generality of the implementation (specifically, the need to fine tune for certain situations where the method fails and the apparent hard coding of the filter width to 21 points) which are noted inline. 

## Novelty and Impact

The package appears to fill a gap in the existing array of tools available to the authors. That said, additional literature review discussing the place of this work in the context of other similar and previous work would improve the reader's understanding of its novelty and impact.

## Verifiability

The software is available online under an MIT license and software dependencies are noted where necessary (e.g. Cantera is cited, SUNDIALS is cited, matplotlib is cited, etc.). Software snippets are informative.ces should be identified (e.g., with citation to a persistent DOI). Analysis is clearly described in the usage example, though referencing a full input file would improve repeatability. 


### Figures

- Figure 1 is very busy for its size. Perhaps its size can be increased to cross both columns?

- Figures 5 and 6 have lines which overlap so closely, the meaning is somewhat lost. Is there a more helpful way to plot these results (e.g. plot the difference between the two lines as an additional overlay on the plot?)

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+
+Background
+----------
+
+The RCMs at the University of Connecticut have been described extensively elsewhere
+:cite:`Das2012,Mittal2007a`, and will be summarized here for reference. The RCMs use a single piston
+that is pneumatically accelerated and hydraulically decelerated. In a typical experiment, the
+reaction chamber is first evacuated to an absolute pressure near 1 Torr, measured by a high-accuracy
+static pressure transducer. Next, the reactants are filled in to the desired initial pressure
+(|P0|), and a low dead volume valve on the reaction chamber is closed. Compression is triggered by a
```
katyhuff: For folks who are not combustion engineers, it may be helpful to be clear about the meaning and function of this low dead volume valve.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+:math:`2^{14}-1`. The cutoff frequency, window type, and filter order were determined empirically,
+based on Fig. :ref:`frequency`. Methods to select a cutoff frequency that optimizes the
+signal-to-noise ratio are currently being investigated.
+
+.. figure:: figures/frequency.png
+
+    Power spectral density profiles of the original, filtered, and filtered and smoothed signals,
+    showing the peaks of noise above 10 kHz. :label:`frequency`
+
+After filtering, the signal is smoothed by a moving average filter with a width of 21 points. It is
```
katyhuff: Is this moving average filter width tuned to the specific case you are running at UConn? If it is hard-coded in the algorithm rather than an option to specify in the input file, it will need to be made clear why this value is the right one for all problems that the package might be used for.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+The ``Simulation`` class sets up the simulation in Cantera and controls the running simulation.
+Importantly, the maximum time step is set to be the time step used in the volume trace, so that the
+simulation does not take steps larger than the resolution of the velocity. Larger time steps may
+result in incorrect calculation of the state if the velocity is not properly applied to the reactor.
+As the simulation runs, the solution time, temperature, pressure, and simulated volume are appended
+to lists that are converted to NumPy arrays :cite:`vanderWalt2011` when the simulation finishes.
+Once the simulation finishes, the derivative is computed using second order Lagrange polynomials, as
+suggested by Chapra and Canale :cite:`Chapra2010` because the time step is not constant in the
+simulation. Finally, |TC| and the overall ignition delay (if a reactive simulation was run) are sent
+to the system clipboard to be pasted into a spreadsheet. The first stage ignition delay must be
```
katyhuff: Rather than pasting into a spreadsheet, have you considered directly sending the output to a csv? Pasting into a spreadsheet seems error prone.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+found manually because determining peaks in the derivative is currently unreliable, as mentioned
```
katyhuff: This unreliability of the algorithm seems to defeat the purpose of the package. Perhaps I'm misunderstanding.  How often is this determination unreliable in the simulation case?

bryanwweber:
@katyhuff Thanks for the comments! I'm working on the response right now, but I have a question about this sentence:
> Software snippets are informative.ces should be identified (e.g., with citation to a persistent DOI).

I'm not sure what ces is, can you clarify? If it should be "Sources", can you clarify which sources? Thanks!

katyhuff:
Whoops. That was a copy paste error from the review criteria template.
Delete the part that says "ces should be identified (e.g., with citation to
a persistent DOI)."  sorry!

papers/bryan_weber/bryan_weber.rst
```diff
@@ -0,0 +1,708 @@
...
+engine-relevant conditions, including shock tubes and rapid compression machines (RCMs).
+
+An RCM is typically designed with one or two pistons that rapidly compress a homogeneous fuel and
+oxidizer mixture inside a reaction chamber. After the end of compression (EOC), the piston(s) is
+(are) locked in place, creating a constant volume reaction chamber. The primary diagnostic in most
+RCM experiments is the pressure measured as a function of time in the reaction chamber. This
+pressure trace is then processed to extract the ignition delay.
+
+In this work, the design and operation of a software package to process the pressure data collected
+from RCMs is described. This package, called UConnRCMPy :cite:`uconnrcmpy`, is designed to enable
```
bryanwweber: Changed to "our package"

papers/bryan_weber/bryan_weber.rst
bryanwweber: Changed to "In this paper", also in a few other places where a similar usage was intended

papers/bryan_weber/bryan_weber.rst
bryanwweber: Done.

papers/bryan_weber/bryan_weber.rst
bryanwweber: Assuming that the noise is approximately normally distributed, all of the points before the action
starts are equivalent in terms of setting the offset. Thus, it shouldn't matter if we pick the 50th
point or the first point where the convolution is fully applied. Since the initial point is set
equal to the first point where the convolution is fully applied, choosing the initial point is
equivalent to choosing any other point before the compression starts. We have clarified the text at
this location:

> **where :math:`\overline{V}(0)` is the initial voltage of the filtered and smoothed signal. Assuming
the noise in the signal has an equal probability of being above or below the mean voltage, choosing
the initial point (i.e., :math:`\overline{V}(0)`) to set the voltage offset is equivalent to
choosing any other point prior to the start of compression. The result is a vector of pressure
values that must be further processed to determine the time of the EOC and the ignition delay.**

papers/bryan_weber/bryan_weber.rst
bryanwweber: Replaced the parentheses with commas, "are" is the correct tense now.

papers/bryan_weber/bryan_weber.rst
bryanwweber: "Intensive property" here is used in the thermodynamic sense, intensive meaning not dependent on
system size, and property being any path independent quantity associated with the system. This is
standard usage in thermodynamics. That paragraph has been corrected to read:

> **In Cantera, intensive thermodynamic information about the system is stored in an instance of
the ``Solution`` class. The ``Solution`` classes used in this study model simple, compressible
systems and require two independent properties, plus the composition, to fix the state. The
two properties must be intensive (i.e., not depending on system size), and are typically
chosen from the pressure, temperature, and density.**

papers/bryan_weber/bryan_weber.rst
bryanwweber: The "Then" was a leftover from a previous draft and has been deleted. The paragraph has been
clarified as to the user's role vs. the parts implemented in the software.

papers/bryan_weber/bryan_weber.rst
bryanwweber: We have modified the text to read as follows:

> **Two simulations can be triggered by the user that utilize this procedure. In the first, the
multiplier for all the reaction rates is set to zero, to simulate a constant composition
(non-reactive) process. In the second, the reactions are allowed to proceed as normal. Only the
non-reactive simulation is necessary to determine |TC|.**

> **When a reactive simulation is run, the user must compare the temperature traces from the two
simulations to verify that the inclusion of the reactions does not change |TC|, validating the
assumption of adiabatic, constant composition compression. Although including reactions during the
compression stroke does not affect the value of |TC|, it does allow for the buildup of a small pool
of radicals that can affect processes after the EOC :cite:`Mittal2008`. Thus, it is critical to
include reactions during the compression stroke when conducting simulations to compare a kinetic
model to experimental results.**

papers/bryan_weber/bryan_weber.rst
bryanwweber: The user has to check this. The software doesn't give a warning in the current version. This has
been clarified and an issue has been opened: https://github.com/bryanwweber/UConnRCMPy/issues/4

papers/bryan_weber/bryan_weber.rst
bryanwweber: This was indeed a new thought, and has been moved to a new paragraph. See the response to [L339](https://github.com/scipy-conference/scipy_proceedings/pull/198#discussion_r69908765).

papers/bryan_weber/bryan_weber.rst
bryanwweber: Done. Also corrected an instance of the same thing earlier in the paragraph.

papers/bryan_weber/bryan_weber.rst
bryanwweber: Yes, it refers to the output file from LabView containing the measured voltage trace. This has been
clarified.

papers/bryan_weber/bryan_weber.rst
bryanwweber: It is actually meant in both of these senses; the reactive experiments are conducted first, before a
non-reactive experiment is conducted and each reactive experiment is processed by UConnRCMPy as soon
as it is collected. This has been clarified in the text as follows:

> **In general, for a given condition, the user will conduct and process all of the reactive
experiments before conducting any non-reactive experiments. Then, the user chooses one of the
reactive experiments as the reference experiment for the condition (i.e., the one whose ignition
delay(s) and |TC| are reported) by inspection of the data in the spreadsheet. The reference
experiment is defined as the experimental run whose overall ignition delay is closest to the mean
overall ignition delay among the experiments at a given condition. To select the reference
experiment, the user puts the file name of the reference experiment into a YAML format file called
``volume-trace.yaml`` with the key ``reacfile``.**

papers/bryan_weber/bryan_weber.rst
bryanwweber: Done.

papers/bryan_weber/bryan_weber.rst
bryanwweber: In the current version of UConnRCMPy, nothing specifically happens. Usually, either the
`volume-trace.yaml` file or the experimental data file referenced by the `reacfile` key will not
exist and a `FileNotFound` exception is thrown. This has been clarified in the text.

papers/bryan_weber/bryan_weber.rst
bryanwweber: Done.

papers/bryan_weber/bryan_weber.rst
bryanwweber: The user in all of these cases. This has been clarified in the text.

papers/bryan_weber/bryan_weber.rst
bryanwweber: Done. Also added explicit sections to other cases of "previously".

papers/bryan_weber/bryan_weber.rst
bryanwweber: Done.

papers/bryan_weber/bryan_weber.rst
bryanwweber: Done.

papers/bryan_weber/bryan_weber.rst
bryanwweber: Done.

papers/bryan_weber/bryan_weber.rst
bryanwweber: We have used the `Path` class to use some of the convenience features in that class, such as the
`resolve` method. At the moment, the `add_experiment()` method expects either an instance of `Path`
or no argument at all, although there is no particular reason it couldn't be modified to accept a
string. An issue has been opened to track this: https://github.com/bryanwweber/UConnRCMPy/issues/7

papers/bryan_weber/bryan_weber.rst
bryanwweber: We have used the `Path` class to use some of the convenience features in that class, such as the
`resolve` method. At the moment, the `add_experiment()` method expects either an instance of `Path` or no argument at all, although there is no particular reason it couldn't be modified to accept a string. An issue has been opened to track this: https://github.com/bryanwweber/UConnRCMPy/issues/7

papers/bryan_weber/bryan_weber.rst
bryanwweber: Yes, the `volume-trace.yaml` file must have exactly that name and must be located in the current
working directory. We have clarified in the text after the code block.

papers/bryan_weber/bryan_weber.rst
bryanwweber: We have removed the "low dead volume" phrase in the interest of saving space on a full explanation.

papers/bryan_weber/bryan_weber.rst
bryanwweber: The moving average filter width, along with the low pass filter's cutoff frequency, filter order,
and window type, were all determined empirically. It is likely that the selection of the filter
width can be optimized using similar methods as selecting a cutoff frequency, as mentioned in the
previous paragraph. We have modified the text to clarify that this is on the agenda for improvement.

papers/bryan_weber/bryan_weber.rst
bryanwweber: We prefer copy/pasting to a spreadsheet so that the user can interact with the data immediately. At
least for Excel (not sure about LibreOffice or other software), modifying files that are open
does not reload the modified file, which limits interactivity. In any case, the analysis can be
easily repeated if the value gets lost since the computation of TC takes less than a minute in all
the cases we have seen.

papers/bryan_weber/bryan_weber.rst
bryanwweber: The algorithm reliably detects first stage ignition about half of the time. However, in the worst
case it is only unreliable for one third of the output information (i.e., it only has trouble with
the first stage ignition delay, but the overall delay and TC are the more important parameters).
Therefore, detection of the first-stage ignition in the simulations is disabled until the algorithm
can be improved. Fortunately, most cases don't have a first stage ignition, so this does not
strongly affect the usability of the package. Nonetheless, this is high on our list of desired
improvements for the package.

bryanwweber:
> Hello Bryan, My name is Dan Schult and I am a reviewer for this submission. I have never done a Github-based review before so please bear with me as we make our way through this process. I am quite excited to see how it works because the potential is really nice. :)

> The manuscript looks really good already. My comments are relatively minor. I will try making may of them via line-comments, though I guess that will create lots of emails. I'm guessing that it will also be good to keep the comments separate from each other so we can keep track of which have been addressed. At first I thought I would be making Issues and PRs to your personal branch, and I guess I could do that. But it seems set up for me to make comments and then you address them via further commits to this PR. Let me know if there are better ways I could be giving feedback.

@dschult Hi Dan! Me neither, but I think this will work quite well. Thank you for your comments, they are very insightful to see the gaps in my assumptions!

> My main comment is that it is not consistently clear which parts of the process are done directly by the user and which are done by the software

We have clarified these specific comments where they are mentioned in the inline comments.

> Also, for the parts done by the user, which involve interacting with the RCM directly and which by controlling the RCM through the Jupyter notebook interface? The best presentation of the user workflow is in the "Usage Example" section (which is appropriate). Perhaps you can change to an "active voice" from the passive case in that section to make it clear who is doing what.

The interaction with the RCM is not controlled by the Jupyter notebook (yet). The notebook is only used to process the data after an experiment is conducted. We've clarified where the user is interacting with the RCM and where the data is being processed. We've also gone through the manuscript and changed to active voice wherever it helped the flow of the text and clarified who the actor was.

> One example is: "From the runs at this condition, it is determined that the reference experiment..." Is that determination done by the software and reported somehow? Does the user make this determination by looking at the figures?

> At the end of this paragraph, "the first non-reactive experiment is added to the Condition" could be: "we then add the first non-reactive experiment to the Condition". In this case, I'm pretty sure that the user must type the next python command, but using an active tense makes it explicit.

The response to these examples is in the inline comments above. To answer your other questions, the user makes the determination of the reference case after examining the data that have been copied to a spreadsheet. We have clarified this in the text.

> Also, it is not clear whether the non-reactive experiment was performed after processing the results from the previous step, or were all experiments performed before coming to the computer.

Each run is processed as soon as it is conducted, so all of the reactive experiments are processed before any non-reactive experiments are conducted.

> This comment may not be reasonable to influence this paper, but it suggests yet another possible method to find EOC times. You could fit the initial portion of the data to an exponential and the middle portion of the data as a line. Then the EOC time could be the time where the exponential crosses the line. I suspect this would give slightly smaller EOC times than the max of pressure. But the max of pressure is unlikely to be at EOC. The system will take some (short) time before the loss to walls overcomes the gain from recent compression. I don't know what the relative time-scales are so hard for me to make a guess at which is more accurate.

Thank you for the suggestion. At present, the largest uncertainty in determining the EOC time is the noise in the pressure trace, which may introduce up to about ±2 ms uncertainty in the EOC time, as judged by overlap in the plotted compression strokes. This method could certainly help reduce that uncertainty, and we've opened an issue to investigate this further:
https://github.com/bryanwweber/UConnRCMPy/issues/5

bryanwweber:
> Here's some advice from Proselint (http://github.com/amperser/Proselint), a Python-based linter for prose, with seven false alarms removed:

>     689:24: redundancy.garner Redundancy. Use 'plans' instead of 'Future plans'.
>     221:46: weasel_words.very Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be. Found once elsewhere.

@suchow Thank you for the advice. We prefer how the text reads as written, so we have opted not to incorporate the changes.

bryanwweber:
> Apart from minor comments above, my only substantial comment is that I actually found it quite hard until the end of the paper to understand what UConnRCMPy actually does - the examples which start on page 7 are good, but until then it's not clear what the package provides. I think it would be beneficial near the start to give a high-level overview of the main features of UConnRCMPy, maybe formatted as a bullet point list.

@astrofrog Thank you for the comments. We have responded to the inline comments above. In addition, we have added a few paragraphs in the introduction to clarify the features of UConnRCMPy and the purpose of this paper:

> **In this paper, the design and operation of a software package to process the pressure data collected from RCMs is described. Our package, called UConnRCMPy :cite:`uconnrcmpy`, is designed analyze the data acquired from the RCM at the University of Connecticut (UConn). Despite the initial focus on data from the UConn RCM, the package is designed to be extensible so that it can be used for data in different formats while providing a consistent interface to the user.**

> **Recognizing that reproducible research is an important goal for the scientific community :cite:`NatureEds2016`, and that the code used to process experimental data is an important part of reproducing research, the primary goal of UConnRCMPy is to enable consistent, reproducible analysis of RCM data. Thus, UConnRCMPy offers offers all of the features required to process standard RCM data including:**

> * **Filtering and smoothing the raw voltage generated by the pressure transducer**
> * **Converting the voltage trace into a pressure trace using settings recorded from the RCM**
> * **Processing the pressure trace to determine parameters of interest in reporting the experiments, including the ignition delay and machine-specific effects on the experiment**
> * **Conducting simulations utilizing the experimental information to calculate the temperature during the experiment**

> **Previous software used to analyze RCM data has generally been undocumented and untested code specific to the researcher conducting the experiments. Moreover, the software typically used to estimate the temperature in the experiments is difficult to integrate with the data processing code. To the best of the authors' knowledge, UConnRCMPy is the first package for analysis of standard RCM data to be presented in detail in the literature, and it tightly integrates the temperature estimation routine into the workflow, reducing errors and inefficiencies.**

> **This paper serves to describe some of the important aspects of RCM data processing, particularly the choices that the operator must make that are rarely documented. In addition, as a complement to the in-source documentation, this paper documents the design choices, interface, and flexibility of UConnRCMPy.**

bryanwweber:
> Again, I'm so sorry for the delay. I think this is paper is in good shape and looks like a good addition to the proceedings. I had only minor comments.

@katyhuff Thank you for the insight! We have responded to your inline comments above.

> The prose is written in professional english and clearly communicates the narrative. That said, I do agree with @astrofrog that a lot of the punch isn't delivered until the end. Namely, UConnRCMPy delivers a tool to 'process' the pressure data from RCMs, however the goal of the process (to extract ignition delay) is buried somewhat between pages 1 and 7. This broader purpose is mentioned on page 1, but is subsequently somewhat lost in the weeds until the final demonstration section.

Please see the response to @astrofrog. We have expanded and made more explicit the discussion of the use of this package, so that the purpose doesn't get lost in the discussion.

> The technical content appears sound and describes the application of python to a domain specific problem. The equations are well presented and clearly defined. I had a few questions about the generality of the implementation (specifically, the need to fine tune for certain situations where the method fails and the apparent hard coding of the filter width to 21 points) which are noted inline.

We have responded to these comments inline.

> The package appears to fill a gap in the existing array of tools available to the authors. That said, additional literature review discussing the place of this work in the context of other similar and previous work would improve the reader's understanding of its novelty and impact.

We have added a paragraph to the introduction, please see the response to @astrofrog.

> The software is available online under an MIT license and software dependencies are noted where necessary (e.g. Cantera is cited, SUNDIALS is cited, matplotlib is cited, etc.). Software snippets are informative. Analysis is clearly described in the usage example, though referencing a full input file would improve repeatability.

We have added discussion of the `species.cti` input file, including a citation. In addition, the third YAML code snippet is a minimal, but complete, example of a `volume-trace.yaml` file. We have clarified this in the text.

> Figure 1 is very busy for its size. Perhaps its size can be increased to cross both columns?

We have increased the size of the figure to cover both columns.

> Figures 5 and 6 have lines which overlap so closely, the meaning is somewhat lost. Is there a more helpful way to plot these results (e.g. plot the difference between the two lines as an additional overlay on the plot?)

For Figures 5 and 6, the purpose is to show how much the lines overlap, so we haven't made any changes to the figures. We have clarified the caption of Figure 6 to note that the desired result is that the lines overlap. Note that Figure 5 is removed from the revised version for space considerations.

bryanwweber:
@katyhuff @dschult @astrofrog One big change I made was to rearrange the Usage Example and the Implementation sections, because I had to eliminate duplicated ideas to fit in 8 pages. This made the Implementation section somewhat shorter, so we get to the Usage section faster. Hopefully that will help the comment about losing the point of the work. Thank you to all for the comments!

katyhuff:
Looks good to me -- I'm satisfied by these changes.

papers/bryan_weber/bryan_weber.rst
```diff
@@ -687,7 +687,7 @@ method to determine the optimal cutoff frequency in the filtering algorithm.
 Acknowledgements
 ----------------
 
-This material is based on work supported by the National Science Foundation under Grant No.
+This material is based on paper supported by the National Science Foundation under Grant No.
```
dschult: this line (690) should probably be left as "work" instead of "paper"

dschult: Ignore that last comment as it was changed in a later commit :)

dschult:
This looks good to me too. Very nice. :)

I accept this paper to be merged

papers/bryan_weber/bryan_weber.rst
bryanwweber: Find/replace all mistake! Thanks for the detailed notes 😄

sarostru:
Looks like this one is about ready to go, this post may not be relevant but letting everyone know about the timeline change.

astrofrog:
@bryanwweber - the changes look good to me.

I accept this paper to be merged

PR #199, opened 2016-05-30
Paper: Fitting Human Decision Making Models using Python (Alejandro Weinstein et al.)

aweinstein:


papers/alejandro_weinstein/alejandro_weinstein.rst
```diff
@@ -33,16 +33,139 @@ A topic of interest in experimental psychology and cognitive neuroscience is to
...
+
+In this work we present a Python program able to fit experimental data to a RL model. The fitting is based on a maximum likelihood approach [Cas02]_. We present simulation and experimental data results. 
+
+A Decision Making Model
+-----------------------
+
+In this section we present the model used in this work to describe how an agent (either an artificial one or a human) learn to interact with an environment. The setup assumes that at the discrete time :math:`t` the agent selects action :math:`a_t` from the set :math:`\mathcal{A}=\{1, \ldots, n\}`. After executing that action the agent gets a reward :math:`r_t \in \mathbb{R}`, according to the properties of the environment. Typically these properties are stochastic and are defined in terms of probabilities conditioned by the action. This sequence repeats :math:`T` times. The objective of the agent is to take actions to maximize the total reward
```
sbenthall: how an agent...learn 
(number agreement)

sbenthall:
Hello, just letting you know I've been assigned to review this paper.

sbenthall:
This paper clearly explains reinforcement learning, specifically the Q-learning paradigm, and an implementation of it using Scientific Python. Nice use of data visualizations. This is an informative paper.

I have one major question about the paper, whose answer I think would greatly improve the scientific interest of the paper. Given that reinforcement learning here is used as a model of _human decision making_, how well does that model fit the data, relative to other competing models?

To be clear, the paper presents how the parameters of a reinforcement learning model can be fit to empirical decision-making data. The paper would be strong if the reinforcement learning model were compared with some other model--perhaps a null hypothesis of some kind?

Then you could look at the likelihood of the data given the reinforcement learning and null hypotheses. The likelihood ratio would give you the Bayesian evidence in favor of using the reinforcement learning model in the first place.

As is, I don't know from the paper whether reinforcement learning is actually a good model of human behavior. You could have fit any model to the data!

sbenthall:
Figure 4 is a very interesting graphic. It appears that there are very different ranges of values of the estimated alpha and beta for the 80% context and 20% context.

Can you provide some kind of interpretation of this result? If I understand the model correctly, a low alpha and high beta mean that the learner is not updating Q-values quickly from experience (low alpha) and that the learned Q-values are having little impact on probabilistic behavior (high beta). Isn't it striking that many of the subjects in the 80% condition don't seem to be learning at all?

This seems like a good opportunity to talk about what you can learn from fitting reinforcement learning models to human decision-making. Please do!

aweinstein:
Thank you @sbenthall  for reviewing the paper. You bring very good points. 

Regarding your first comment, it's well known that this model is not always able to describe the behavior of some subjects. Typically we know when that happens because the optimization of the likelihood function hit a bound.  We'll add a comment about that in the paper. In addition, it is posible to compute confidence intervals of the estimates. We'll add that to the paper also.

Regarding your second point, we'll add some comments about figure 4.

michaelpacer:
Hi, just letting you know that I was assigned to review your paper. 

I like this — it's going good places. I'm determined to help you get to those good places :). 

I did my first pass through, and now will need to compile my comments. In the meantime here are straightforward things to fix that relates to figures. I know figure changes can take a while so I wanted to get you that info as soon as I could. 

Fig 1:

1. Other figures are in sans serif, change this.
2. What is the arrow? If it is the "win" or "loss" response say so explicitly
3. If all the delays are 1s long, don't use 1000 mS (and if you did its ms not mS).
4. Use a dotted line in the part of the arrow to indicate a self-paced arbitrarily long period of time (if there is an upper limit on the amount of time say so in the caption)

Fig 2: 

1. Don't call them "reward context" and "punishment context" when these are the cues/contexts that alter the expected reward value. Instead use "winsome" and "losesome" or "likely gain" and "likely loss" or "happy-making" and "sad-making" (not really suggesting the last pair… but just trying to give ideas about the range of solutions  you could turn to).
2. The top of this figure is hard to read when printed grey scale.
2. I know based on what you've said that the top part indicates whether a particular trial was a success or a failure, but it is not clear from the way it is currently displayed. It needs a legend as well as a caption to map the semantics of your paper to the semantics of the figure. 
2. This format of a figure would not generalise easily to other more complicated RL games and blends information across the figure modalities. Specifically, you cannot identify what $r_t$ is for any trial without aligning the win/loss record with the action taken. This makes it exceptionally hard to see how the data relate to the Q-learning policy you've been discussing.
3. A much stronger and more useful figure would include a plot of the Q function for that cue to illustrate what the expected reward would be after each timestep (at for a one or a few values of $α_c$ and $β_c$). To clarify why: it will allow people to see more directly the connection between the contextual Q-policy estimation method and the results you are presenting. In particular, it should show the model learning a unique feature of your problem: all of the actions are effectively Bernoulli wagers with the same probability (this did not need to be the case). 
4. Font is too small on the axes
5. x-axis scales change between the two graphs, don't do that, its confusing at best.
6. Include all three context cues, you have the space.

Fig 2 suggestion: Because you have a lot of convenient structure that you can leverage in displaying your data, I would recommend different manner of conveying this info. Here's a way of changing it that addresses a number of the above points. 

First, given that you have the space, I would allow these time-course graphs to span the columns. Then, rather than plotting actions as line graphs, keep the general structure you have, but plot the expected reward(Q function) for each action as a line graph, and I would move the win/loss records from the top to the bottom and  either be labeled with the action choice (likely to be too small) or (because you essentially are asking for a continuous wager rather than a true categorical decision) you could allow the length of the bar representing the action be analogous to the size of the wager.

It would have been nice if you had investigated a wider variety of learning algorithms, or at least discussed what the options were. As it is, you seem to have 

Fig 3:

1. Does not print well on greyscale
2. Why did you cut off the parameter boundaries where you did? In the upper left hand corner it looks like you might have a second mode hiding, it would be worth noting what that mode is, and why it manages to have a high log-likelihood. E.g., it looks like it is the case with a relatively low $\alpha$ but high β meaning that a hard maximisation strategy that doesn't learn all that much may predict your decision sequences fairly well. Given that the MLE parameters with that characteristic show up in your estimates of the human data, it is worth investigating whether bimodality holds more generally, or only under certain cues, or if the estimate is unstable, &c..
3. Why is it α and β, I thought it was supposed to be $α_c$ and $β_c$? did you assume in your model that they are all going to be identical? If you did, you should say so more prominently and justify why it is appropriate. This matters less because of the structure of your problem (uniform iid cue sampling) but in the more general case, the distribution could differ strongly for the different contexts.
4. Is this figure from a single sample of 360 trials? What would it look like if there were many fewer trials? 
5. Do the likelihood graphs for other parameter settings look qualitatively similar? If so, it would be useful to say so, if not, please describe in what way they do not. 
6. Include your plotting code; it is useful to see how your plots directly relate to the quantities used throughout the rest of your code

By the way, in estimating your likelihood, on the one hand it seems like you are taking the entire history of the data that you have observed and just calculating parameters for the total likelihood. However, the presumption of your model is that (even if we simplify $\alpha_c$ and $\beta_c$ to be constant across contexts as α and β) they will be constant over the course of learning (NB: you have a typo in your eq (3) in which you index $\alpha_c$ and $\beta_c$ as $\alpha_t$ and $\beta_t$). That means that you should be able to take a on-line estimate of $\alpha$ and $\beta$ and you should not find the MLE changing too dramatically (at least after enough data has accumulated).

It would have been nice if you had discussed other methods of parameter estimation and justified the stability of a MLE estimator in this case. (This is especially true since it looks like it might have a multimodal loglikelihood). 

Fig 4. (this one is the most problematic)

1. I don't understand what this figure actually does for the paper…
2. You don't have a legend so its impossible to identify the colour of any of the dots
2. Your caption and content conflict with each other, you suggest on the one hand that you are estimating $\hat{\alpha}$ and $\hat{\beta}$ for a *subject* which suggests based on your previous analysis that you are assuming each *subject* has the same α and β for each of the cues. But then based on your caption (and the colouring of the dots) you seem to be fitting them separately for each of the cues. Which is it? Why would you presume that people have individual differences that covary with the conditions that you have developed? If they are to covary, do you think that those differences are general enough to be useful in the analysis of human decision making (i.e., that they are not artefacts of the particular structure of your environment)? 
4. Do not include only two of the conditions, it is misleading. It is especially misleading if the main point of the graph is to indicate the general distribution of the parameters 
5. If these are parameter estimates from the same participants under different cues, having some way of linking the parameter estimates derived for the same participant, even if it's just a line connecting them (or a triangle once you include the neutral condition). 

Were it not for this being the only instance in which you are modelling human data I would consider cutting this figure entirely. However, given the title of your paper, and its general focus I think it cannot be cut. Instead, I would focus on explaining how you arrive at the graph (i.e., give more code examples that lead you through estimating from people's data to generating a graph like this).

More generally to make the human decision making data more relevant to your reported work it would help to explain how you might try to estimate these parameters from people directly, or discuss how to interpret them as they apply to more complicated cases (e.g., where actions aren't implicitly on a continuum, where different contexts actually have different kinds of pay-off structures).

Fig 5: 

1.  increase font size (especially for the word "choice")
2. why does it say Cue 0 at the top? Either use standard nomenclature or no title, letting the caption cover it
3. Right now this doesn't seem to add all that much to the paper. What insight are we intended to draw from this.
4. I think a grid like this of similar graphs that plotted the Q functions like I suggested in 2), but to have this plotted for different values of αs and βs would be significantly more interesting than an array of data that don't really fit with the rest of what you are saying.

Some notes on the math:
Your probability function should be including (either as parameters or as conditions) not only $\alpha_c$ and $\beta_c$ but also $Q_0(•,•)$. Right now your derivation is not accurate. Also I think you want to estimate the conditional probability (since you aren't explicitly modelling the distribution of the cues themselves).  

Some notes on your discussions of modelling: 

You need to separate out the two kinds of models that you are working with in this paper: the decision making model and the inference model for fitting parameters to your decision making model. Right now their exposition (and consequently their identities) are muddled. 

In the first case, you are describing a decision making model, based on the reinforcement learning model paradigm, specifically using Q learning to determine your decision policy, which you have implicitly taken to be stochastic. This has several different parts that need to be separated from one another. E.g., though eq (1) is relevant to eq (2), but they have extremely different roles in the context of the model. eq (1) is essentially a rolling estimate of the value of different actions in different contexts. eq (2) is a generative distribution for randomly selecting actions (or determining the probability of an action given your parameters, Q-history, past actions, and past rewards). You may use this in retrospect as just being part of your likelihood, but you then need to be more explicit about how you are doing so. I mention this because I worry that right now you are not exactly doing this correctly. 

The problem arises in how you use this probability function in your parameter learning model, which I think is correct in the code, but is not what you have written down. Basically you don't have 
$$P(a_t, c_t) = \frac{e^{\beta_{c_t}Q_t(a,c_t)}}{\sum_{i=1}^ne^{\beta_{c_t}Q_t(a_i,c_t)}}$$ 
but rather
 $$P(a_t| c_t,{a_1,\ldots,a_{t-1}},Q_0(\cdot,\cdot),\alpha_{\cdot},\beta_\cdot) = \frac{e^{\beta_{c_t}Q_t(a,c_t)}}{\sum_{i=1}^ne^{\beta_{c_t}Q_t(a_i,c_t)}}.$$

NB: if you didn't want to condition but did want to calculate the joint probability I think you would need an additional summand in the normalisation term (specifically to account for all of the different values that $c_t$ can take on in actuality). 

This matters because it draws attention to the fact that this probability distribution is dynamically updated between time-steps, and (more importantly) carries strong dependencies across the time-steps. and yet your $\mathcal{L}(\alpha_c,\beta_c) = \prod_{t=1}^TP(a_t,c_t)$ equation implicitly treats them as independent. It might even be good to denote the probability function as $P_t(a_t,c_t)$ to make it even more explicit that it is changing

It looks like in your actual code, you are generating the incremental log-likelihood on the basis of the current Q-value at each time-step, so I think your computational work should be fine. You should spend extra time going through your first few sections, because these kind of details will jump out at you. I want to take a look at the newer version of the section introducing the parameter estimation model once you have a chance to edit these things. Right now it is probably the least clear of your mathematical sections. 

Note on theory: You are relying on sophisticated implicit arguments relating rational models of decision making/reinforcement learning to computational systems and to human beings. At the least, you should cite some theoretical justification in the literature discussing computational models of human behavior. E.g., Marr's(1982) levels of analysis or Anderson's(1990) rational analysis. There is more recent work along these lines, but you've got some citations in that vein so I don't have any specific things to suggest. Those citations won't be enough, you should actually include at least a paragraph explaining why doing any of this makes sense.

Small note: explain what temporal discounting is, and in particular why you aren't using it in your reward function?

Small note: You should cut all references to neuroscience. Why? Rhetorically: it does not appear in your paper anywhere after the intro, adds nothing substantial without the reader doing independent reading elsewhere, and distracts from your point. Conceptually: you already have two inferential layers between the human behaviour, your (stochastic) decision model, and parameter estimation model. Being able to correctly navigate all of the probabilistic care needed to properly take those inferential steps into account (i.e., actually having some uncertainty in the parameters of the decision model, if not in the form of the decision model itself) is exceedingly difficult.  

I have more comments, but this is probably more than enough for now. I hope this helps.

suchow:
Here's some advice from Proselint (http://github.com/amperser/Proselint), a Python-based linter for prose, with five false alarms removed:

```
261:388: weasel_words.very Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be.
```

darribas:
Hi there,

I've had a run through the paper and, without being an expert in the topic at all, I think it explains the problem, the approach, and the solution in an intuitive way. I agree with @sbenthall that it'd make a stronger case comparing this approach with other standard ones in the literature. But I also think, being a paper for a Python conference, the paper rightly focuses on the implementation tricks and details.

I only have a couple of cosmetic changes I'd suggest:

* The first code snippet, where `ContextualBandit` is spelled out is partially chopped off by the figure on its right. I think it's just a matter of making the comment link shorter and breaking it into two.
* This might seem snob but, after seeing last year's [talk by Nathaniel](https://www.youtube.com/watch?v=xAoljeRJ3lU), the paper would be more effective using `viridis` instead of `Jet` for Figure 3 ;-)
* Finally, this is mostly out of pure curiosity. Could you please give a sense in the paper of how computationally intensitve these models are? For instance, on a standard laptop, how long does the artificial agent take to run? And fitting the model with the experiment's data?

Once these are corrected, the paper is good to go AFAIC.

sarostru:

aweinstein:
@sbenthall I think that the current version of the paper address the two points you made:

- I added standard errors to the parameter estimations. We now have an idea of how well does that model fit the data.

- I added some comments related to figure 4.

I'll work on the comments of the other two reviewers during the day.

aweinstein:
@darribas I think that the current version of the paper addresses the three points you made:

- Fixed code snippet that was chopped off.

- Use viridis in figure 3. I let you decide if now looks better :)

- Added info about execution time.

sbenthall:
Thanks for these changes, @aweinstein . I recommend this paper for acceptance.

sarostru:
@darribas @sbenthall Based on the comments I'm adding the ready label.

@aweinstein next up we will prepare for the final version, so I final proofread by yourself and @sbenthall and I will help with updating the proceedings build from draft to final.

aweinstein:
@sarostru @sbenthall  Thanks. I just proofread the paper. I found that one image was over-running the text. I just commit a change to fix this.

Thank you all for reviewing the paper.

PR #200, opened 2016-05-30
Paper: The Climate Modelling Toolkit

JoyMonteiro:
This is a paper supporting our poster in the earth sciences mini symposium titled **Towards CliMT 1.0**


cbcunc:
I accept this paper to be merged.

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+.. class:: keywords
+
+   Climate Modelling, Hierarchical Models 
+
+Introduction
+============
+In an influential essay, Isaac Held made the case for studying the climate
+system using a hierarchy of models, in a manner similar to the hierarchy of model organisms used by
+evolutionary biologists :cite:`heldModelHierarchy2005`. The essay argued that such a hierarchy would
+not only help in our understanding of the climate system, but would also help in interpreting results obtained from
```
sarostru: I think a diagram showing what climate understanding hierarchy looks like would be quite appropriate here.  Later you could have another diagram showing the climate toolkit structures and how they fit into such a hierarchical approach.

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+(cloud-scale) simulations of
+the entire atmosphere. One of the primary reasons, we believe, is that significant effort is required to build
+models which represent even the basic features of the atmospheric general circulation.
+Existing frameworks to develop such models like the Flexible Modelling System
+(FMS, ``http://www.gfdl.noaa.gov/fms``) and the MITgcm (``http://mitgcm.org/``) are typically
+written in Fortran and the effort to set up a model beyond those already provided as examples can,
+in our experience, be quite discouraging for new users who lack a strong background in Fortran and
+programming.
+
+In this paper we introduce the Climate Modelling Toolkit (CliMT), which attempts to reduce this
```
sarostru: I think this introduction could use a paragraph that gives the reader some notions of the details of why CliMT is better than the previous ones, beyond the programming language difference, e.g.:
 - Is it more modular, has better abstractions?
 - What type of numerical schemes does it support?  Quick summary of the class of methods supported.
 - What are the performance aims and how the approach will scale across larger compute infrastructures?

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+evolutionary biologists :cite:`heldModelHierarchy2005`. The essay argued that such a hierarchy would
+not only help in our understanding of the climate system, but would also help in interpreting results obtained from
+more complex models and even aid in improving them.
+
+Over the past few decades, efforts have been made to develop such a hierarchy of models (a
+significant fraction of which has been, not surprisingly, by Held's students and collaborators)
+which have had a positive impact on our understanding of the climate system and the general
+circulation of the atmosphere
+:cite:`hoskinsKaroly1981,heldSuarez1994,neelinQuasiequilibrium2000,schneiderTropopause2004,friersonMoistGrayGCM2006,marshallAquaplanet2007,caballeroAxisymmetricRadConv2008,mitchellTitanMethane2009`.
+Note that we restrict our focus here only on numerical models of the climate system, excluding many
```
sarostru: For non-specialists a quick explanation here of what a "numerical model of the climate system" consists of would be helpful, e.g.  3D models, 2D models, grid-based, finite-volume, non-hydrostatic or whatever properties you are thinking of?

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+interpreting the results of the simulation easier. Given that model initialization,
+execution and data analysis can be performed from within a single IPython notebook, this makes model
+results and the resulting scientific results reproducible as well. CliMT also enables users to study the effects 
+of changing physical parameterizations and dynamical cores on the simulated climate,
+something that is difficult to do in more elaborate GCMs.
+
+Architecture
+============
+
+In CliMT, each component (encapsulated in the ``Component`` class) provides tendency terms to the main
```
sarostru: what is a tendency term?

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+
+The initialization of the components and the execution of the resulting model is handled in the same
+script, which makes the parameters and assumptions underlying the model explicit. This makes
+interpreting the results of the simulation easier. Given that model initialization,
+execution and data analysis can be performed from within a single IPython notebook, this makes model
+results and the resulting scientific results reproducible as well. CliMT also enables users to study the effects 
+of changing physical parameterizations and dynamical cores on the simulated climate,
+something that is difficult to do in more elaborate GCMs.
+
+Architecture
```
sarostru: General comment, I think a broad sweep introduction to what a system using these looks like and what kinds of simulations it would do is necessary before diving into the details of components.

For instance, I'm assuming there is some kind of 3D or 2D grid and that somehow components are like contributions into a given cell, but I really have no idea if that is right, and the later examples that have a single 'ocean' component make that hard to understand.

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+other classes being used internally by these two classes.
+
+``Component``
+-------------
+
+A ``Component`` class is the fundamental abstraction in CliMT. It encapsulates the behavior of a
+component that takes certain inputs and provides certain tendencies as output. Each ``Component``
+object has (among others) the following members which are specified by the developer:
+
+- ``Prognostic``
```
sarostru: An inline short summary of what these terms mean is needed.

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+    # Main timestepping loop
+    for i in range(1000):
+        # The following code adds a uniform
+        # 1 K/day cooling rate to 
+        # the internally-computed tendencies
+        dT= np.array([[-1./86400.*kwargs['dt']*\
+                2.*np.ones(rad.nlev)]]).transpose()
+
+        fed.step(Inc={'T':dT})
+
```
sarostru: Some results plots from one or both of your examples would be nice, lots of page room left

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+characteristic time scales are very different from each other without increasing the computational
+load during the simulation. Notice also the external tendency term ``dT`` passed on to ``fed`` in the
+``step`` method.
+The output fields are again accessed by treating ``fed`` as a
+dictionary.
+
+Monitoring fields and I/O
+=========================
+
+CliMT also provides for online monitoring of the simulated fields. Currently, upto four fields
```
sarostru: upto is two words

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+integrator does not multiply the tendency terms by ``dt``, as is normally the case.
+
+Current Development: towards CliMT 1.0
+======================================
+
+Moving forward, we hope fulfill the vision of using CliMT as a full fledged moist idealized GCM. As
+a first step, we have integrated a dynamical core adapted from the Global Forecast System (GFS).
+Together with this, we have added a new Held-Suarez module which provides the Held-Suarez forcing
+terms for a 3-d atmosphere. A working example of the benchmark is now available from a development
+fork (``https://github.com/JoyMonteiro/CliMT``).
```
sarostru: Could you point at the exact example code?  I looked around but didn't find the held-suarez example

papers/joy_monteiro/towards_climt_1.0.rst
```diff
@@ -0,0 +1,377 @@
...
+
+Many changes were incorporated enroute this integration. The dynamical core is the first component
+of CliMT that interfaces with the Fortran library using Cython and the ISO_C_Binding module
+introduced in Fortran 2003. This will be used as a template to eventually move all components to
+a Cython interface: ``f2py`` does not seem to be actively developed anymore, and currently cannot interface with code
```
sarostru: A related question I had regarding the fortran libraries, where do they come from?  I didn't quite understand that, and how do you get updates from the source of the routines when they are updated?

In the introduction you say that there are well tested fortran libraries, but if you take a point in time copy of them, then you are now on the hook for maintaining them and I didn't see any testing infrastructure in the repo.  Testing is an active process, how do you test your code?

sarostru:
Thanks this is a nice summary of your library and I enjoyed reading through, the idea of doing global climate modelling in an ipython notebook is very attractive.

I think as is, it is very close to being documentation for the software so I recommended a some improvements to help with comprehension, especially for readers not already versed in using climate modelling tools.  

One other aspect that I didn't fully appreciate was to what extent does using CliMT empower a user to use all of the different libraries in the python ecosystem?  For instance, how do users mix and match components written in python with other components using fortran?

JoyMonteiro:
Thanks, Scott; this is very helpful. Will start working on the edits.

papers/joy_monteiro/towards_climt_1.0.rst
taldcroft: :+1:, I had the same idea in reading this.

taldcroft:
I read the paper and went to make comments and found that @sarostru had already written down almost exactly what I was thinking!  So we are on the same page that this is a nice paper where more context for non-climate specialists would be very helpful.  Following on the ideas already expressed:

- You show some toy examples, but can you show some results from a full-fledged simulation done using CliMT?
- Is this package at the point of being able to do production climate work?  Perhaps that is the goal for the 1.0 release that you mention.  It would be good to specifically compare what is now, and what will be, possible with CliMT versus established codes like FMS and MITgcm.
- Do you see any fundamental limitations with your approach?
- Echoing what @sarostru mentioned, I was very unclear on how the different models get stitched together in this framework.  This is particularly relevant for model components that are coupled.  Presumably the ocean and atmosphere are coupled so I wonder how that gets done in this framework.
- Relatedly, the discussion of how the internal integrators of different components get exposed (i.e. some Fortran model code) could use some work.  Getting back to the example of coupled components, how do you do the integration?
- Have there been any non-developer users of the code?  Maybe it is still early.

JoyMonteiro:
Thanks, Tom. Some of the simulations using the 3-d dynamical core are working, and I plan to use them as demos during my poster session. There is a standard test suite for dynamical cores that is used by major modelling groups around the world, and we are in the process of implementing all of them (2 out of three are working!) and will demo them.

One of the reasons I don't mention these developments in the paper is that they are on my fork which is under heavy development right now, and the code is broken because we are rewriting/designing some internals. We don't expect the API to change, however, and this was the reason we wanted to show simple examples. 

I must mention that the radiative convective equilibrium example in the paper, though it looks simple, is a fundamental model in climate science, and that script can be used for research work without much modification. In some ways, it shows how easy it is to use CliMT for non-trivial simulations.

Thanks for your suggestions, I will try and address all comments in the next draft!

Joy

papers/joy_monteiro/towards_climt_1.0.rst
wfspotz: Especially if the python syntax to obtain a plot is non-trivial

wfspotz:
A pronunciation guide would be helpful. Do you prefer "Klimt" or "Kly-em-tee"?

wfspotz:
How extensive is the built-in python documentation? Can I quickly find a list of available modules with definitions, a list of valid field names with definitions, etc?

wfspotz:
The line continuation character `\` is not necessary within parentheticals, and is more readable if you omit them.

wfspotz:
Did you mean "Stratospheric" in comments in the Federation code example?

wfspotz:
Laymen may not be aware that vertical levels are usually measured in pressure rather than spatially. A short comment to this effect could be helpful.

papers/joy_monteiro/towards_climt_1.0.rst
JoyMonteiro: the meaning of tendency and leapfrog have been added. I have omitted the filtering part since it is common knowledge that both have to be used together, and it just makes the prose verbose.

papers/joy_monteiro/towards_climt_1.0.rst
JoyMonteiro: I had added some plots. @wfspotz obtaining plots simply involves sending a list of fields to be 'monitored' during instantiation of the model. I have made this clearer now.

papers/joy_monteiro/towards_climt_1.0.rst
JoyMonteiro: They are a point in time copy, I have now made it explicit in the manuscript as well. As we have mentioned in the conclusion, testing in CliMT has been quite informal, and a formal testing framework is high on the priority list for CliMT 1.0. 

The more sophisticated Fortran code that has been used in CliMT is taken from climate models whose code base has been frozen (like the Community Atmosphere Model v.3, frozen around 2004), so no updates are expected upstream as well.

JoyMonteiro:
Hello, I have committed a new version which hopefully address most comments!

sarostru:
@JoyMonteiro You've addressed all of my comments and the changes look great.

I accept this paper to be merged.

sarostru:
Need to let the other reviewers respond, before putting the ready tag, sorry about that.  Seems like you are in good shape, posting a timeline change notification on all the PRs.

wfspotz:
This looks good to me and I accept the changes.

taldcroft:
I accept this paper to be merged.

PR #202, opened 2016-05-30
rmsf example

orbeckst:
Example with figure that shows how to do basic MDAnalysis while showing off integration with NumPy and abstractions. Plus a nice segue into the Analysis chapter.


orbeckst:
sorry, that was premature. closing

PR #204, opened 2016-05-30
Paper: ObjArray: Better Integrating Python Objects with NumPy Arrays

Justin-Fisher:
Paper submission for SciPy 2016 proceedings.


mrocklin:
### General feedback

This paper identifies a nice missing feature set of nd-arrays of objects and a natural extension of the numpy syntax to homogeneous arrays of objects.  It does a good job of laying out the competing objectives and the difficulties in implementing such a library.

However, when discussing the design of the library it flows more like a list of features and choices made rather than a principled and tested approach.  Some specific areas for improvement are listed below:

Should probably strip out informal speech.  This includesi all  mention of my, I, myself as well as phrases like "So really, ..." or "as far as I can tell"

Perhaps provide a brief outline towards the beginning of the functionality added and the functionality not added before discussing each piece separately

Does this support broadcasting?

It would be good to have a running example that was reproducible and illustrative of the virtues and challenges of this project.  This example would ideally be robust enough to motivate benchmarking and complex enough to compare and contrast API with other alternatives, like recarrays or nested lists.  These performance evaluations and API comparisons would give this paper much more substance.

The source code for this project is not public.  Attempts to view the source code result in a message "You do not have permission to access this location."  The link points only to a single .py file, which raises software engineering questions of "does this have tests?"  "is it published on PyPI or some other channel?"  "Is it a proper software package with documentation?"  Given that this is a submission about a software library rather than a scientific experiment these questions are possibly fairly important.  

This paper provides no references.

papers/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
...s of Numeric Python (NumPy) arrays, including speed, a vast library of powerful functions, and flexible indexing.  Before now, it was quite cumbersome for Python coders to enjoy both sorts of benefits at the same time.  Python's objects offer excellent OOP, but could not easily be used with NumPy.  NumPy's structured record arrays offer NumPy virtues but only a pale approximation of OOP.  NumPy also allows arrays of Python objects, but these provide only fancy indexing, not straightforward access to OOP attributes or methods, nor direct access to the vast library of fast NumPy functions.  Instead, coders who want to enjoy both OOP and NumPy have needed to write many explicit loops/mappings/comprehensions to copy arrayed objects' attributes into NumPy arrays and vice versa.  My free package, ObjArray, offers a solution to this problem, providing NumPy users clean natural access to OOP attributes and methods--for example, ``people[ people.age > 18 ].height`` does exactly what you would intuitively expect it should do, namely return a NumPy array of the height attributes of all the people over age 18--all while providing full access to the speed, power, and flexible indexing of NumPy.
```
kynan: Avoid unscientific phrases like "before now" or "my free package"

papers/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+1.  Goals.
+----------
+The overall goal of my free package ObjArray is to better integrate Python's object-oriented programming (OOP) features with Numeric Python (NumPy).  Such integration would be suitable for some, but not all, Python projects.  Python's OOP may not be suitable for projects involving extremely large data sets because Python's objects are not especially efficient in terms of memory storage or processing speed.  Conversely, very small Python projects may have little need for NumPy's speed (though other features may still be useful).  Very many Python projects fall between these extremes.  These "medium-sized" projects are large enough to need, in some places, to draw upon fast powerful NumPy functions, but are small enough that memory considerations permit using numerous Python objects and speed considerations allow that many non-time-sensitive operations can be written in an OOP format that is often easier for many coders to write, read, understand, and debug.  These medium-sized projects could be well served by a package like ObjArray that makes it much simpler to take full advantage of the strengths of both OOP and NumPy.
```
kynan: This is a rather non-traditional opening for a paper: better call it "Introduction" and start with an introduction of the *problem* your package is addressing and then make the case for how it does so.

I also don't think I agree with your notion of "medium-sized" projects. It's not about size, it's about the problem you're trying to solve. You're package gives users a higher-level access to OO methods on collections of objects in an efficient manner, so why not argue from that angle?

papers/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+To allow this, the class ObjArray has a special ``__get__`` method that catches all failed attempts to retrieve an attribute from an ObjArray itself, and instead redirects those attempts to retrieve that attribute from the various objects in the ObjArray and to package those attribute values together in a new array.  If OA is an ObjArray, then ``values = OA.x`` will return a new array whose initial dimensions match OA and whose contents match the corresponding ``.x`` attribute-instances for the objects in OA, including whatever dimensions those instances have.  For example, if OA is a 5x5 ObjArray, and each ``.x`` instance is a 3x3 array of integers, then ``OA.x`` would return a 5x5x3x3 array of integers.   The shape and dtype of attribute instances is automatically determined by sampling a member of OA, or you can manually specify these with keyword arguments, via the more verbose ``OA.read_attr(attr_name, shape=None, dtype=None)``.
```
kynan: Should this be `__getattribute__` instead of `__get__`? Or do you use the descriptor protocol? This does not become clear.

What happens if there are "conflicts" i.e. attributes which exist in both the ObjArray and its containing objects? Which takes precedence?

papers/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+For these reasons, I regretfully opted to make the dot idiom for ad hoc writing require some form of prior declaration that an attribute name is to be treated as a writeable attribute of arrayed objects, rather than as a writeable attribute of the ObjArray itself.  This prior declaration can be done explicitly, but more often is implicit--e.g., ad hoc reading an attribute implicitly declares it to be eligible for later writing.  Explicit and implicit declarations both cause a factory to create a new property for the class ObjArray (or for whatever subclass of this you create and employ) whose ``__set__`` method intercepts attempts to set ``OA.x`` and redirects and broadcasts these to the ``.x`` attributes of the arrayed objects (except when ``OA.__dict__`` already contains an entry for 'x', in which case that is modified).  In cases where the dot idiom is unavailable--because an attribute hasn't been declared yet, or because the attribute name is one like 'shape' for which NumPy has already dictated a meaning--you can always use the more verbose ``OA.write_attr('shape')``.
```
kynan: How does explicit declaration look like? An example would be useful.

The explanation of attribute access triggering the creation of a property would very much help in understanding how reads work. Consider moving it to 3.1.

papers/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+Coupled buffers experience no slowdown at all for NumPy operations on the buffers, which are usually the operations for which speed is most crucial (goal N1).  Coupling may cause slight slowdown accessing or changing attributes from the object side though, because these operations are redirected to the relevant portion of the buffer.  This slowdown might be significant by NumPy standards, but not by the standards usually applicable for Python code that operates on objects one at a time.
```
kynan: How is this implemented? Does coupling also create a property for that attribute on the ObjArray returning the coupled buffer directly?

papers/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+The ideal destination for my ObjArray package would be to have a future version of it be adopted as an optional subpackage within NumPy itself (e.g., as ``numpy.objarray``).  I suppose it could be incorporated pretty much as is, and would likely be fairly useful to many users already as a much more convenient way to make good use of NumPy arrays with dtype=object.  It might be sensible, though, to have this undergo public discussion first, and/or for developers who better understand the under-the-hood workings of NumPy to optimize my inner loops in C.
```
kynan: Distributing objarray with NumPy would put the maintenance burden on the NumPy maintainers. Until the package is rigorously tested and very popular this seems unlikely.

papers/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+Still, my proof of concept already works well enough to be tremendously useful in generating code that is simple, clean, fast, intuitive and beautiful.  This demonstrates that the rewards to be gained through my "dark magic" very likely will be worth the costs.  Of course, that's what practitioners of dark magic always say.
```
kynan: Rather than an off-the-cuff funny remark it would be great to have some "hard evidence" that the package actually performs. Some benchmarks would be great or at least some real use case beyond the trivial people class mentioned throughout the article.

kynan:
This paper lays out 8 goals for working with objects stored in NumPy arrays, argues how NumPy's record arrays and arrays of objects fall short in achieving those and describes the implementation of a Python package ObjArray which achieves all 8 goals.

It certainly seems an interesting and reasonably thought through approach to what is probably a niche problem. The implementation becomes mostly clear from the text (some points for improvement highlighted in comments).

However I would like to second @mrocklin's concerns. This paper is not yet in a state suitable for publication. The following needs to be addressed (also see my inline comments on the diff):

1. The source code and the example on the website linked in the paper is not accessible. I consider this crucial for accepting this paper for the SciPy proceedings.

2. There is no substantial evidence the package actually delivers what is promised in the paper. Neither a compelling use case (the paper only uses a trivial person class throughout) or full example program, nor any benchmarks are provided.

3. No references are cited.

4. The writing style is very informal and "unscientific". Avoid the first person (me, I) and colloquial phrases and expressions like "nicely", "coders", "crisp clean abstraction", "excellent", "sadly", "pale approximation", "soup up", "smattering", "happily", "beautiful simplicity", "diabolical genius", "glass half empty/full", "regretfully", "dark magic", "as far as I can tell" etc. Many arguments in the text are rather weak and off-the-cuff.

5. The opening is very "non-traditional" and fails to adequate motivate the need for this package. The conclusions (luridly titled "The perils of dark magic and a vision of the future") is also rather weak (see 2.)

6. The pull request must be fixed to not include the output directory.

Another issue not addressed by the paper (more for my personal curiosity):

Given that ObjArrays afaiu are not typed, how are heterogeneous arrays dealt with i.e. those containing objects of different types? How are errors handled? I can immediately think of 2 classes of errors:

* missing values: some objects contain attribute x, others do not
* shape mismatch: different objects yield data of nonuniform shape for attribute x

suchow:
Here's some advice from Proselint (http://github.com/amperser/Proselint), a Python-based linter for prose, with five false alarms removed:

```
60:168: cliches.write_good 'reinvent the wheel' is a cliché.
1:18: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
22:4: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
86:4: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
112:849: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
118:292: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
120:479: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
122:347: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
124:731: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
124:880: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
90:638: typography.symbols.multiplication_symbol Use the multiplication symbol ×, not the letter x.
90:690: typography.symbols.multiplication_symbol Use the multiplication symbol ×, not the letter x.
90:594: typography.symbols.multiplication_symbol Use the multiplication symbol ×, not the letter x. Found once elsewhere.
92:216: typography.symbols.sentence_spacing More than two spaces after the period; use 1 or 2.
93:-6: typography.symbols.sentence_spacing More than two spaces after the period; use 1 or 2.
90:715: typography.symbols.sentence_spacing More than two spaces after the period; use 1 or 2. Found 5 times elsewhere.
134:378: uncomparables.misc Comparison of an uncomparable: 'quite possible ' is not comparable.
20:424: weasel_words.very Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be. Found 6 times elsewhere.
```

davclark:
First off - my apologies for the late review. I have just recently moved and this review assignment came as I was wrapping up my old job, driving across the country, etc. Given my late feedback, I adopted the process of taking notes independently, but then reading other reviewers comments to streamline my own feedback.

## Summary

The paper presents a new class using numpy as a backend, promising to efficiently store arrays of objects, with full access to object oriented programmin, and an easy transition from working with single instances of objects to arrays. Unlike recarray, this extension makes non-trivial departures from the in-memory storage logic of structured arrays.

The figure is very helpful, but more concrete explication is needed.

## Detailed feedback

In short, I am in strong agreement with other authors. I would in particular point out that the one code example in the intro is not hard to implement with record arrays or with pandas:

```
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({'a': [1,2], 'b': [3,4]})

In [3]: df[df.a == 1].b
Out[3]:
0    3
Name: b, dtype: int64

In [4]: import numpy as np

In [5]: ra = np.rec.fromarrays([[1,2],[3,4]], names='height,width')

In [6]: ra[ra.height == 2].width
Out[6]: array([4])
```

More generally, the implementation appears to be a hack that transforms a record store into a complicated column store. To this end, pandas is probably a better choice of back-end, and moreover, it is surprising that pandas is not more thoroughly considered.

More comments are provided inline with the prose.

output/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+1.1. Object-Oriented Goals.
+===========================
+	**O1.  Familiar Python objects in familiar class hierarchies.** A good integration of Python OOP and NumPy should make it easy to take code that began as pure Python OOP and to expose the attributes of arrayed Python objects for use in fast NumPy functions.  In particular, we'd want object arrays to contain familiar Python objects, situated in familiar class hierarchies.  For example, if A is a list, tuple, or array of ordinary Python objects, a simple command like ``OA = ObjArray( A )`` should suffice to create an array of those objects with easy access to the other features listed below. 
+
+	**O2.  Flexible object size.**  Different objects in the same class may contain data of different sizes (e.g., strings of different lengths), and some may possess attributes or methods that others lack.  This flexibility is a distinguishing feature of Python's OOP and is appreciated by many coders.  NumPy itself does not generally allow this sort of flexibility, but our integration of OOP with NumPy should.
```
davclark: It is entirely unclear how ObjArray improves on object pointers vs. recarrays or pandas.

output/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+	**O2.  Flexible object size.**  Different objects in the same class may contain data of different sizes (e.g., strings of different lengths), and some may possess attributes or methods that others lack.  This flexibility is a distinguishing feature of Python's OOP and is appreciated by many coders.  NumPy itself does not generally allow this sort of flexibility, but our integration of OOP with NumPy should. 
+
+	**O3.  Attributes of arrayed objects should be easily accessible by familiar dot idiom.**  If ``Adam`` is an individual object, then all Python coders know that ``Adam.height`` probably returns Adam's height attribute.  We should support parallel behavior for arrays of objects.  For example, if ``people`` is an array of person objects, then ``people.height`` should return an array the same shape as ``people``, containing each person's ``.height`` attribute.
```
davclark: This is a non-point, as dot-notation attribute access is already available. Either require something stronger, or drop this point.

output/justin_fisher/justin_fisher.rst
```diff
@@ -0,0 +1,141 @@
...
+
+	**O3.  Attributes of arrayed objects should be easily accessible by familiar dot idiom.**  If ``Adam`` is an individual object, then all Python coders know that ``Adam.height`` probably returns Adam's height attribute.  We should support parallel behavior for arrays of objects.  For example, if ``people`` is an array of person objects, then ``people.height`` should return an array the same shape as ``people``, containing each person's ``.height`` attribute. 
+
+	**O4.  Methods of arrayed objects should be easily callable by familiar dot idiom.**  Similarly, for individual objects, we all know that ``Adam.grow(1)`` will call Adam's ``.grow`` method, with Adam being passed in as an implicit "self" argument, and 1 being passed as the explicit argument.  We should support parallel behavior for arrays of objects, so ``people.grow(1)`` should call each person's grow method on herself with argument 1, and ``people.grow(people.height)`` should call each person's grow method on herself with her own height as the argument.
```
davclark: The only situation in which pandas.DataFrame.apply falls short here is when a method would modify attributes of the object. Again - I think you could build your functionality around existing idioms in Pandas much more straightforwardly than with numpy!

sarostru:

kynan:
None of the concerns raised by the reviewers @mrocklin @davclark  and myself have been addressed. In addition this pull request has basic formal errors (includes the output directory) and is not suitable for merging. There was no response from the author @Justin-Fisher and it is unclear whether the major concerns outlined would be addressed before the final deadline. I can therefore *not* recommend this paper for inclusion in the SciPy 2016 proceedings.

sarostru:
@mrocklin @kynan @davclark Thanks for taking the time to review this submission and providing a lot of useful feedback, based on the comments above and the lack of response from @Justin-Fisher I am closing this PR.

Thanks for submitting @Justin-Fisher , hopefully you can make use of this feedback for your future work.  I hope you enjoyed the conference, but please in future out of consideration to the time of our volunteer reviewers; if circumstances change and you know you won't have time to finish your submission, then a note to that effect in the comments or closing the PR earlier yourself would be appreciated.

Thanks and best of luck in future,
Scott (@sarostru) and Sebastian (@sbenthall)

Justin-Fisher:
Hi folks, I'm very sorry not to have known that your comments were being posted here.  I was watching for them in my email inbox (which is where the referee comments on all my other submissions over the years have arrived), and was sad not have received any!  Fortunately, I emailed the organizers yesterday to check if my comments had somehow fallen through the cracks, and they pointed me to the right place.  At their invitation, I have revised the paper in light of your thoughtful comments, for which I am thankful!  I have received useless comments on more paper submissions than I can count over the years, but was delighted that yours were completely the opposite of that and instead made many points that were completely on target!  So thank you for the time you spent, and sorry again not to have been here to reply sooner.  

I will put as many responses inline above as I can, and will respond to other comments below.  I don't seem to be able to upload the revised paper into the pull request (perhaps because the pull request was closed) but it looks like I can attach it to this message (though only in .pdf form, not .rst).  If you get a chance to take a look at it, I'd appreciate it.

[Fisher_objarray_paper.pdf](https://github.com/scipy-conference/scipy_proceedings/files/392346/Fisher_objarray_paper.pdf)

output/justin_fisher/justin_fisher.rst
Justin-Fisher: No, I think they differ in many other circumstances as well.  Pandas’ apply calls the very same method for all arrayed objects, whereas ObjArray calls each object’s own method.  Since different objects may have different methods (due to being members of different subclasses, or as part of Python’s ubiquitous duck-typing), this is not the same thing at all.  Pandas’ apply passes only arrayed attributes of objects to the method, and not a Python object itself, so there is no way for the given method to access other attributes or methods of the relevant object. I think there are also limits to Panda's broadcasting of additional arguments.  Also, strangely, Pandas' apply calls the given function twice on the first row of data and only once on the rest of the data, which also is often not what you would want, unless your methods don't change anything.  Also, even if there weren’t all these important differences, OA.grow(1) is simpler than and hence preferable to OA.apply(‘grow’,args = (1,) ) anyway.

**In the new discussion of Pandas (section 2.3), I now note that apply provides a partial approximation of O4.**

output/justin_fisher/justin_fisher.rst
Justin-Fisher: **I have now clarified O3 and O4  to say that attributes and methods should be _collectively_ accessible.**  Dot notation is available in some data-structures, but not others (e.g., nested lists of objects, or ndarrays with dtype=object), and the ones that provide it fall short of other listed goals.

output/justin_fisher/justin_fisher.rst
Justin-Fisher: Recarrays and Pandas do not allow flexible object size -- instead every object within one of those arrays must have the same byte-size.  ObjArray employs object pointers.
**I added a note in the new introduction emphasizing that different existing approaches meet different goals, but none meets all.  I also note in the introduction that ObjArray will build on NumPy arrays with dtype=object.**

papers/justin_fisher/justin_fisher.rst
Justin-Fisher: **I added a new section with benchmarking (Section 5).**

papers/justin_fisher/justin_fisher.rst
Justin-Fisher: You may be right.  Pandas does use magic methods with much the same dot idiom syntax as I am proposing though (albeit with a much different underlying data structure), and NumPy uses magic methods to allow dot-idiom access to record arrays, so there is some hope that this syntax might be commonplace enough for NumPy to consider the natural extension to ndarrays of dtype=object.  **This paragraph now mentions these syntactic precedents**

Despite these admitted obstacles, I don’t think it hurts to mention this as an ideal future for the project.  NumPy does include ndarray’s with dtype=object, and ObjArray facilitates the usefulness of those arrays in ways that would be relevant in virtually every instance where arrays of dtype=object would be of use, so this would be quite natural to include with them.

Alternatively, you might think that it’s odd that NumPy includes dtype=object as an option at all, when these arrays won’t support the vast majority of NumPy functions, and when these manage data retrieval so differently from all other ndarrays (following pointers to data, rather than directly returning contents from the array’s memory block).  So maybe I’m lucky that NumPy offers this functionality at all.  But given that they have committed themselves to branching out at least a little way in this odd direction, maybe it isn’t that much to hope that they might make the natural extension to make this odd branch much more useful.

papers/justin_fisher/justin_fisher.rst
Justin-Fisher: To a decent approximation, yes.  Technically, it stores a link to the coupled buffer as the array’s __dict__ entry for that attribute name, and if necessary it will create the same descriptor in the class ObjArray that likely will already have been created by ad hoc reading or explicit declaration.  This descriptor’s __get__ method looks first for this __dict__ entry before attempting expensive ad hoc reading.  Since Python forces descriptors to be associated with whole classes, not particular members, the only way for me to allow some ObjArrays to have an attribute coupled even while others don’t is for me to use a one-size-fits-all descriptor and count on something else (I use __dict__ entries that will always be preempted by the descriptor) to distinguish between the arrays that are coupled and the arrays that aren’t.

**This has now been clarified and spelled out in more detail.**

Justin-Fisher: (pleae pardon its auto-correcting my double-underlines into boldface)

papers/justin_fisher/justin_fisher.rst
Justin-Fisher: **An example is now included, using the arguments keyword in the initial creation of an ObjArray.**  

**I have now clarified the reading section, and ensured the two fit together.**  I didn’t end up moving the bit about property-creation up to the section about reading though, because it really has very little effect on reading.  For the most part, the behavior of ad hoc reading is well captured by the blanket _ _getattr_ _, and the extra role that the created property plays is very minor.

papers/justin_fisher/justin_fisher.rst
Justin-Fisher: I use both `__get__` and `__getattr__`.  In order to handle attempts to get attributes that had not been declared in advance, I use `__getattr__` to catch each failed attempt to read an attribute from an ObjArray and to redirect those down to read that attribute from the arrayed objects.  Once this failsafe gets triggered for a new attribute name, or whenever an attribute is explicitly declared, a new descriptor is created so that, from then on, attempts to retrieve that attribute from an ObjArray will be handled directly by this new descriptor’s `__get__` method, rather than falling through to `__getattr__`.  The descriptor is needed for ad hoc writing, for reasons described in that section, but it also provides minor improvements for reading (being triggered slightly more quickly than `__getattr__` would be, and quickly redirecting to a coupled array when one is available, thereby avoiding the costs of an unnecessary ad hoc reading).

**Anyway, you’re right that `__getattr__` is probably a bit more clear to coders experienced with magic methods, albeit a bit harder to read and pronounce.  I’ve now changed to `__getattr__` and `__setattr__` where those are technically more appropriate.**

papers/justin_fisher/justin_fisher.rst
Justin-Fisher: **The introduction has now been lengthened and rewritten.**  

**The new introduction eliminates the discussion of project size.**  It probably goes without saying that Python objects consume too much memory space and processor time to be used profligately in very large projects, which was the main point I had been trying to acknowledge in that passage.

papers/justin_fisher/justin_fisher.rst
Justin-Fisher: **I have made many changes to wording throughout to make it sound more 'scientific'** (often at cost of brevity, readability or clarity, but you do what you have to do).

Justin-Fisher:
I could make inline responses to comments that were associated with particular lines, but not to others.  Here are a few responses to other points people brought up.

First off, sorry there was a **permissions error on the source-code download.  That has been fixed for some time now, and it downloads fine from my computers,** so please let me know if anyone still has trouble downloading it.

**I have added stock-references for NumPy, SciPy, and Pandas.  Are there other references you think I should include?**

**The new introduction does several things y'all suggested, including providing a brief description of ObjArray's starting point (numpy arrays of dtype object) and the functionality added.**

**Yes, @mrocklin, ObjArray does support broadcasting (both of new values in write operations, and of arguments in method calling), and I have now added emphasis on that in several places.**

**I have now added significant discussion of Pandas, both explaining why it falls short of the goals I lay out (this is done in section 2.2) and describing how ObjArray attributes can be coupled with Pandas dataframe columns to allow programmers to use full-fledged OOP in concert with Pandas (section 4.2), which also does a little to address @kynan's concern that ObjArray might serve a small niche.**  Thanks @davclark for forcing me to look more closely at Pandas.

**@kynan asked how ObjArray handles missing data or mismatches in shape.  This has been clarified in the paper, and I'll type a longer answer here.**  

Verbose calls allow you to specify a desired shape and dtype with keyword arguments, e.g., `OA.read_attr(‘x’, shape=(2,), dtype=np.int16)` will ad hoc read the .x attribute from the objects arrayed in OA and broadcast them to the given shape and cast them all as the given dtype.  Dot-idiom calls, like `OA.x` will instead sample the .x attribute of an object in OA to determine what shape and dtype to use.  (The sampled object is whichever is first in a flattened version of OA.)

In the case of missing values, ad hoc reading will cause an error from trying to read the .x attribute from an object that lacks a .x attribute.  `OA.read_attr` allows you to pass a default_value keyword argument which instead replaces missing values with the given default value.

In the case of shape mismatches, broadcasting will occur where possible (e.g., if we expected a 2D value, a 1D value will be broadcast to fill it) and otherwise an error will occur (e.g., if an object’s .x attribute is 3D whereas the specified/sampled shape is just 2D, there is no way to broadcast 3D down to 2D so an error will occur).  

Another sort of mismatch you don’t mention is mismatching dtype.  This will work wherever NumPy can do the casting itself, and otherwise will cause an error.

(Ad hoc reading basically uses ndindex to loop over indices in OA and sets output[i] = OA[i].x.  This automatically yields the various behaviors mentioned above, generating errors when there is no OA[i].x, and automatically recasting and broadcasting OA[i].x to fit into output[i] wherever possible.  In the case where a default_value is given, the loop catches failed attribute retrievals and uses the default_value instead.)

sarostru:
@mrocklin @davclark @kynan There was an unfortunate miscommunication with @Justin-Fisher so I've reopened the PR, if you are able to take a look at his updates and see if your comments are addressed, then we can try and finish this review properly.

sarostru:
Something is wrong with this build on the build server, so hopefully people can build locally.  

Sorry about this, I've spent an hour debugging with no success and my laptop battery is about to die :(
I opened an issue for it, https://github.com/scipy-conference/scipy_proceedings/issues/229 @stefanv @katyhuff @sbenthall

davclark:
I am confused. I don't see any new commits - just the PDF uploaded into the comment above. What are you trying to build @sarostru?

davclark:
On that point, @Justin-Fisher it would be very useful to have the diff between the previous and current paper. I had a look [here](https://github.com/Justin-Fisher/scipy_proceedings/network) and honestly have no idea where the new commits are.

If you commit your changes, it is trivial to get the diffs I'd like to see! One of the good reasons to use git for this kind of writing!

davclark:
And while I'm asking for things:

1. Are you willing to share your benchmarking code?
2. Your links for "Source Code" and "Sample Usage" point to the same place [here](http://www.justin-fisher.com/research/objarray/)

I'd *strongly* suggest hosting your code in a more standard location / manner if you're interested in gaining collaborators or even users!

output/justin_fisher/justin_fisher.rst
davclark: No argument regarding simplicity / clarity.

But, what are you talking about with this?

> Also, strangely, Pandas' apply calls the given function twice on the first row of data...

Please have a look at [this notebook](https://gist.github.com/davclark/228bc2a73afbd6a13b12bed0682a736a) and tell me if you get different results. You may be using a borked pandas. Or are you calling functions some other way?

Justin-Fisher: I was trusting the Pandas documentation which says the following:

> In the current implementation apply calls func twice on the first column/row to decide whether it can take a fast or slow code path. This can lead to unexpected behavior if func has side-effects, as they will take effect twice for the first column/row.

I'll check to see if this is actually correct when I get home.

davclark: Well, I just checked and I also only get one function call on the DataFrame (before I was using a Series actually). Maybe the docs are out of date, or they only do that sometimes... there are many possible code-paths in apply now.

This is one of those, "remember, pandas is still 0.XX!" moments.

Justin-Fisher:
Either due to my own incompetence or the bug that @sarostru mentioned above, I haven't been able to figure out how to include the updated .rst into the pull request.  A .pdf generated from the new .rst is attached to my message half-a-dozen or so up from this one, but I understand it would be preferable to have a straightforward diff comparison of the .rst files.  I'll struggle with it more when I get home tonight.

The benchmarking source is now included in the zip file alongside the source code and example usage:  [here](http://justin-fisher.com/research/objarray/objarray.zip).  I'll try to set up a more standard hosting soon.

output/justin_fisher/justin_fisher.rst
Justin-Fisher: Yep, I've tested it now and my results match yours, and not the documentation, so I agree that this is probably just an instance where Pandas documentation is out of date.  I'm eliminating the line on that from my paper -- no need to disseminate out-of-date info on Pandas.  Thanks @davclark for catching that!

Justin-Fisher: Oh, and while we're on Pandas, is there any straightforward way to make 3D or higher-D dataframes in Pandas?  My attempts to do so have all generated errors, but that may just be because I don't know how to use Pandas right, or it may be a restriction built into Pandas, which fits with their docs describing DataFrame as being 2D.  The solutions I've found searching online all just stack extra layers side-by-side in a 2D table, which isn't nearly as convenient as being able to actually index across three or more dimensions.  Insofar as Pandas disallows, or makes it difficult, to have higher-dimensional arrays, that's another big advantage of ObjArray, which works fine for arbitrarily many dimensions.

Justin-Fisher:
It looks like I managed to update the pull request now, so you should be able to view the current .rst here and see differences from the earlier one.  I'm pretty sure my incompetence was at least partly to blame (making changes to the wrong branch of my fork).  Sorry!

output/justin_fisher/justin_fisher.rst
davclark: You probably want this: http://xarray.pydata.org/en/stable/

I can't find the reference, but my recollection is that pandas was dropping some related functionality and authors were encouraging users to use xarray for 3+D data.

The one exception that may remain to this would be panel data, which (to me) still makes sense to be in pandas (it's still an econometric-ish / experimental sampling idea that is a logical extension of tables): http://pandas.pydata.org/pandas-docs/stable/dsintro.html#panel

sarostru:
Ok thanks, the build I was talking about was on [zibi](http://zibi.bids.berkeley.edu:5000/) where the automated pdfs are built.  The problem seems to have resolved itself over the weekend, so maybe Justin's updates resolved the issue.

davclark:
@mrocklin @kynan I did some initial prodding to bring this up to snuff. It's not clear to me if you're intending to re-review or what. It seem that @Justin-Fisher has made substantial improvements and was mostly struggling with proper use of GitHub for paper submission. Even if you want to say, "sorry, too late," it'd be good to know what you're thinking!

I don't expect I can accept on my own, right @sarostru / @sbenthall ?

davclark:
Also, @Justin-Fisher, I forgot that I wrote something relevant a long time ago:

http://nbviewer.jupyter.org/github/davclark/pandasORM/blob/master/pandasORM.ipynb

Might be worth comparing benchmark something that simple vs. the complexity of what you're doing! Obviously your code does a lot more towards programmer efficiency in an OOP paradigm, but you could use the basic ideas in that notebook to re-implement/extend ObjArray in pandas.

You might also profitably look at:

https://bitbucket.org/zzzeek/calchipan/

cc @zzzeek - you may think the arguments in this SciPy paper provide an interesting dialectic for calchipan

sbenthall:
I'll take on reviewing this paper in order to give the process some clarity.

I don't think your citations are rendering properly. You could revisit the syntax from the example.

Otherwise, looks ready to me.

Justin-Fisher:
@davclark I'm not sure I totally follow the point of your pORM.  I see how you're embedding a Pandas dataframe as the .data attribute inside a larger class and setting up magic methods in that larger class to access information from the embedded dataframe.   What I'm not seeing is how that would gain mileage towards the goals of this paper.  A Pandas dataframe already achieves many of these goals, especially the Numpy goals N1-N4, and also dot-idiom attribute access (03), and a limited form (via dataframe.apply) of calling methods on an array of Pandas-objects (O4).  You don't need to embed dataframes inside a larger class to get that much, and embedding the dataframe inside the larger class is bound to slow down access to the dataframe as compared to accessing the unembedded version of the dataframe.  I do see how subclassing your larger class (or you might have been able just to subclass dataframe itself to get something simpler and faster?) could allow for a sort of hierarchical structuring of objects (e.g, defining your "amount" property on the subclass but not the superclass), but as far as I can tell, this would force all the objects in each particular dataframe to be treated as though belonging to the same subclass, rather than allowing a single dataframe to include objects from diverse subclasses, as ObjArray does.   So I guess I'm seeing this as an incremental improvement at best over just using Pandas itself, still offering no real hope of meeting (O1) using ordinary Pandas objects with their own methods and patterns of class inheritance, nor (O2) allowing for diverse duck-typed objects in the same array.  In the benchmarking, I do consider using Pandas dataframes' built-in attribute retrieval, and since your pORM's basically just wrap that inside an extra layer, they're bound to be slower, right?  So I feel like I'm already giving Pandas the best case possible in the benchmarking I have, and I'm not sure what would be gained by benchmarking some larger class with embedded dataframe.  Or am I missing something?

I have similar questions involving Crouching Alchemist Hidden Pandas.  I can definitely see how this would be useful for people familiar with SQL who want to use Pandas dataframes.  But again, this isn't a way of (O1) getting arrays of ordinary Python objects with their own methods and class inheritance, and I don't think it allows for (O2) much diversity between objects in the same array.  So I'm afraid I'm not seeing how this would get much closer to the goals of this paper than Pandas alone does.  

However, I could imagine someone wanting to wrap ObjArray inside an SQL-dialect, much as Calchipan wraps Pandas DataFrames, thereby allowing SQL-style queries of ordinary Python objects arranged in an ObjArray.   For speed reasons, I would probably want to set up coupled buffers for all attributes that would be accessed in this way, rather than waiting for ad hoc methods to iterate through a bunch of objects on the fly for each query.  Using the approach described in section 4.2 of the paper, you could have the various coupled buffers live together in the same DataFrame and then use Calchipan to query that dataframe.  This would allow you to deal with the objects as Python objects, or to access their attributes as NumPy arrays or Pandas Dataframes, or to query them SQL-style, with all the coordination between these different views of the same underlying data being done automatically behind the scenes.  **Section 4.2 now notes the potential for doing this.**

Justin-Fisher:
I think I figured out the reference bug -- didn't realize I needed to precede the square brackets with a space.  Now it seems to build just the same as the sample paper, as far as I can tell.  Thanks!

davclark:
@Justin-Fisher I don't want to wade too deep into a side conversation here - but just wanted to point out some other ideas that might be constructive. On the issue of having an extra function call or member access, I think in the use case we care about (large datasets, doing SIMD-style operations), that's not the issue - the issue is that I use iterrows! But anyway, that was mostly to potentially spark some thinking - not as a recommendation on how to best implement. In particular, thinking about what you're doing as a kind of ORM in the context of not-really-an-ORM and then also sql alchemy.

Given @sbenthall's chiming in, I'd like to run some of the code samples for sanity checking and otherwise, I agree that this has been dramatically improved in many ways and seems fit for inclusion.

mrocklin:
Are there code samples?  It's not clear to me what experiments were run.  

There appear to be two rst files.  I'm not sure which I should inspect.

Were these comments raised earlier addressed:

> It would be good to have a running example that was reproducible and illustrative of the virtues and challenges of this project. This example would ideally be robust enough to motivate benchmarking and complex enough to compare and contrast API with other alternatives, like recarrays or nested lists. These performance evaluations and API comparisons would give this paper much more substance.

> ... which raises software engineering questions of "does this have tests?" "is it published on PyPI or some other channel?" "Is it a proper software package with documentation?" Given that this is a submission about a software library rather than a scientific experiment these questions are possibly fairly important.

mrocklin:
> @mrocklin @kynan I did some initial prodding to bring this up to snuff. It's not clear to me if you're intending to re-review or what. It seem that @Justin-Fisher has made substantial improvements and was mostly struggling with proper use of GitHub for paper submission. Even if you want to say, "sorry, too late," it'd be good to know what you're thinking!

I'm willing to review this again, but I'm probably biased against at this point.  I probably won't want to go through a lengthy iterative process.  I'm not sure how acceptance works if we have conflicting votes.

davclark:
Given the commitments from others to engage in further (if minimal) review, I went and kicked the tires on ObjArray: https://gist.github.com/davclark/b6ab9ef9ad92cb74bc25c367e0752780

Code was obtained from the link in the paper.

This was copied from benchmark.py, and modified to provide more straightforward output of timings. The script as-is does not print timings, and it should print the analogous numbers to the ones you are putting in the paper. I'll note that the test script also requires interpretation, and should ideally indicate passing or failing of each test. Most of the hard work is done, and could be easily converted into a proper test suite using something like assert statements (and could then be run with py.test for nicer output).

My notebook was run against a recent conda install of python 3.5, and the referenced libraries.

I encountered serious issues with timings being dramatically different across runs, so restricted to single runs in my timing. Based on what I'm seeing, I am unsure of where the numbers in the paper are coming from. I do indeed see a non-trivial speed gain in ObjArray for method application, but idiomatic approaches using matrix math in pandas or record arrays are still *much* faster.

For reads, I see ObjArray as slower (even with more loops and allowing for cacheing). Writes behave funny, with the first write being slow and subsequent reads being faster than pandas and numpy. Of course, whether this matters depends on the kind of work you're doing - but for most of the data analytic codes I'm used to (where I expect you'd get users concerned with ease of use), there is not a lot of re-writing data.

In sum, while a lot of progress has been made, I have serious concerns about the validity of comparisons to other packages. Figuring out why I might be getting opposite patterns is a bit beyond me, but this should at a minimum be noted in the paper!

mrocklin:
If the paper is partially about performance then it should include reproducible and consistent performance numbers from experiments that are well described.  A note saying that things are screwy is, I think, below minimum.

Justin-Fisher:
@davclark  Oh sorry, I had commented out the code that did the earlier timing tests towards the bottom of the benchmarks.py file so I wouldn't have to wait for it to execute as I was probing around trying to figure out why Pandas' apply was so slow, and I forgot to uncomment before I uploaded.  It should be very straight-forward to uncomment segments of that code to run whichever tests you want, and I'll upload a version momentarily that defaults to running all of the timing tests again.

I think timeit's time format may vary from operating system to operating system, but at least on my windows machine it prints the outputs in seconds as floating point, and aside from a units conversion to fixed-point milliseconds, and a slightly more verbose labeling of the functions, those are precisely the numbers I included in the paper.  E.g., here is what it printed for me just now for the reading functions:

> Reading functions:
>   OA_read_coupled: 6.037043297674514e-07
>   DF_read: 5.43333896790496e-06
>   RA_read: 1.0262973606043474e-05
>   OA_read_ad_hoc: 1.068401511675642
>   DTO_ndindex_read: 0.9488756006858097
>   DTO_for_loops_read: 0.3548755161672048
>   NL_for_loops_read: 0.25443541571080175

It is normal for benchmark timings to differ across single runs, due to background processes that take more or less time sometimes interfering.  As noted in the paper, I adopted what I take to be the common practice (e.g., it is the practice that [timeit documentation](https://docs.python.org/3.5/library/timeit.html) suggests) of reporting the minimum time out of multiple trials (I used 10), as that minimum time surely includes whatever time is required for the process in question, and longer-than-minimum times presumably were made longer by outside interference, which hopefully was at a minimum at the minimum time.  At least on my computer, these minimum times remain highly consistent across repeated 10-trial runs.  (If you do 1- or 3-trial runs they do become much more variable, and 10-trial runs done on different days seem to differ more than 10-trial runs done back-to-back probably due to having different background processes running.  E.g., I see that a couple of the numbers I got today differ by a factor of about a third from the ones I included in the paper, but on repeated trials today, they are highly consistent again, just as the repeated trials on the day I recorded them were highly consistent, staying within several percent of each other on repeated runs, which is much less variation than the amount we see across approaches.  The smallest numbers remain perfectly identical across repeated runs even across days.  I think that is due to resolution limits of the time function timeit uses.)  

I'm not sure what you mean by "passing and failing each test."  Each tested piece of code "passes" in the sense that it accomplishes the described task.  (Though the benchmark file does include one function that really would fail, DF_call_raw, which I considered as I was trying to figure out why dataframe.apply was so slow, but that one isn't called in the benchmarking.)  The relevant comparisons from the benchmark trials are just the timing comparisons, not success vs failure.  As the paper notes, timing comparisons are only part of the story because some of the tested approaches (record arrays, dataframes) do not allow for use of ordinary Python objects, and not all the approaches use equally user-friendly syntax.  I suppose I could count approaches as "failing" insofar as they fall short of stated goals of the paper, e.g., record arrays and data arrays "fail" to satisfy (O1) allowing for the use of ordinary python objects in their arrays.  I do make clear in earlier sections how alternative approaches fall short of these goals, and mention that again in the benchmarking section, so I'm not sure what other "passing and failing" notes you'd want me to include?

Justin-Fisher:
@mrocklin Thanks for your at-least-grudging willingness to take another look at this.  In case you missed it in the long thread above, let me repeat my apology for not realizing that referee feedback would be posted here.  I've published my fair share of academic papers over the years, though all in philosophy and/or cognitive science, and I am accustomed to receiving referee feedback by email and didn't know to expect this submission to be any different.  One of the perils of inter-disciplinary work, I guess!  Anyway, I certainly didn't mean to be non-responsive to your earlier comments -- I just didn't know there was anything to respond to!  I'm very sorry if my misunderstanding about how the communication here would work contributed to your now feeling biased against the paper.  

I did appreciate your comments and made efforts to improve the paper in light of them.  I noted these  in the long thread above, but will recap here for your convenience.  I rewrote the introduction, including specifying the starting point for ObjArray (numpy arrays of dtype object) and the functionality that ObjArray adds to this.  I did better to highlight the ways in which ObjArray allows broadcasting.  I included more discussion of principled reasons to develop a project in the way I did.  I included benchmarking to compare against alternative approaches.  You had suggested that I come up with a running example that could be used for benchmarking.   I ended up leaving the running examples in the paper involving things like `people.age` as I think this does best at making it readily apparent to the reader what's likely to be an array with multiple objects in it, and what's likely to be an attribute of individual objects.  However, the new benchmarking section uses a slightly-less-toy of an example, a 1000x1000 array of objects each with three relevant attributes.  

Thanks again for your helpful comments and the time you've put into this!

davclark:
I did some thinking about how to give a [gist](https://gist.github.com/davclark/b6ab9ef9ad92cb74bc25c367e0752780) of what I'm seeing. I also realized that results may be highly skewed by my processor, which is a core m7, so the most recent version of the notebook was done via "Run All" to keep the CPU frequency up.

Doing things this way, I still get more of a multiplier from the slowest to fastest run even on simple reads. I have 4MB cache, so not everything would fit, and so I'd expect a cache miss across each run, but honestly this is getting well outside my expertise.

But more generally, the claims made in the paper seem to hold up approximately on my machine (i.e., ObjArray is faster under the benchmarks).

As you can tell, I am not great at benchmarking - but I'm trying to verify yours. As @mrocklin says, it should be easy for me - I should just run a script, read some prose that explains how you thought carefully about the benchmarks and we're good. For benchmarks AND tests, the operator should not have to know what to expect, the code should know what to expect. The operator can then question whether the code's expectations (i.e., in this case, YOUR expectations) are good ones.

mrocklin:
> @mrocklin Thanks for your at-least-grudging willingness to take another look at this. In case you missed it in the long thread above, let me repeat my apology for not realizing that referee feedback would be posted here. I've published my fair share of academic papers over the years, though all in philosophy and/or cognitive science, and I am accustomed to receiving referee feedback by email and didn't know to expect this submission to be any different. One of the perils of inter-disciplinary work, I guess! Anyway, I certainly didn't mean to be non-responsive to your earlier comments -- I just didn't know there was anything to respond to! I'm very sorry if my misunderstanding about how the communication here would work contributed to your now feeling biased against the paper.

I agree that interdisciplinary work is hard.  Standards and communication mechanisms in different communities differ.  It's important to stay on top of the published instructions on how to submit things and receive feedback.

> I did appreciate your comments and made efforts to improve the paper in light of them. I noted these in the long thread above, but will recap here for your convenience. I fixed the permissions error on the sourcecode download. I altered wording in many places to make it more 'scientific'. I rewrote the introduction, including specifying the starting point for ObjArray (numpy arrays of dtype object) and the functionality that ObjArray adds to this. I did better to highlight the ways in which ObjArray allows broadcasting. I included more discussion of principled reasons to develop a project in the way I did. I included benchmarking to compare against alternative approaches. You had suggested that I come up with a running example that could be used for benchmarking. I ended up leaving the running examples in the paper involving things like people.age as I think this does best at making it readily apparent to the reader what's likely to be an array with multiple objects in it, and what's likely to be an attribute of individual objects. However, the new benchmarking section uses a slightly-less-toy of an example, a 1000x1000 array of objects each with three relevant attributes.

This paper appears to be basing itself on the merits of a software project and so I'm applying the standards of that community.  It could be that my standards here aren't appropriate though; SciPy is composed of a variety of overlapping communities and that's where it gets its strength.  I'm entirely happy to be overruled if the organizers or other reviewer thinks that's appropriate.  Having said that, I'm going to go ahead and continue judging this as a software paper.

It is rare to see a paper about software that is 99% prose with a single table of a single numerical experiment.  The opinions of the author alone are rarely considered sufficient argument for an accepted paper.  For a paper like this, that emphasizes the value of a project through both increased reasoning and increased performance I would prefer to see a number of examples and experiments that show the scope of the problem that the software is trying to solve, both in terms of code readability and performance.  I'm not seeing that breadth here (insofar as I've skimmed again, please let me know if I'm likely missing something).  What I am seeing is a lot of discussion around the problem, claims that the solution is better, and results from a single experiment that is not clearly reproducible.  

It'd be good to see side-by-side comparisons of lots of problems with the relevant projects.  It would be good to see timings on these and the times in which some solutions are better than others.  It would be good to see cases where your project is *not* the best solution.  It would be good to bring up problems that are just plain hard and nothing works well.  It would be good for all of these experiments to be included in a reproducible way, either within the paper itself or in a linked-to repository.  It would be good (though not necessary) for the software to be easily available, perhaps through PyPI.  It would be good (though not necessary) for the project to have cited documentation and a test suite.  

Have you had a chance to check out the other accepted papers this year or in previous years?  The mix of explanation, numerical experiments, code examples, and quantifiable analysis on those experiments and examples might convey what I'm trying to convey more effectively.  I've found the most rewarding thing about interdisciplinary work to be the reading of other papers in that field, rather than the authoring of new papers.  There are some good reads in there.

As I mentioned I'm making some of these claims after only skimming your recent work.  If you think they're incorrect please let me know and I'll dive through things again.  If you think that they're correct but shouldn't be applied here that's totally fine, and we should just bump this up to an organizer or I'll happily step aside if I'm the odd reviewer out.

Justin-Fisher:
@davclark I'll look at your benchmarking stuff tonight, and thanks for sharing this.  If you get a chance, I'd be curious to see what your computer gives if you uncomment the `print_min_times_for_functions` lines at the bottom of your kicking-the-tires file, as those would provide straight-across comparisons to what I've been doing.

@mrocklin I did actually go through the SciPy 2015 proceedings before opting to submit this paper.  The most relevant point of comparison I had found there was [this paper](http://conference.scipy.org/proceedings/scipy2015/pdfs/scott_james.pdf) which provides a relation datastructure, which is basically a dictionary with magic methods to maintain a reverse-dictionary to allow for reverse lookups as well.  Like my project, that one was intended as a convenience for programmers.  As far as I can tell, my paper contains much the same things as theirs does, and goes into just as much depth as theirs does in terms of benchmarking comparisons to other data structures that could be made to provide similar functionality.  If anything, mine does significantly more because objarray provides more complex functionality than their comparatively simple relation class does, and my benchmarking covers more points of comparison.  (There are three tables in my paper, not just the one you mentioned, so it's possible that you didn't notice all of them on your quick read-through.)  Any ways in which my paper falls short in comparison to that paper are quite minor and should be easily corrected. 

I suppose you might think it was a mistake for SciPy to have published that paper in last year's proceedings.  And maybe you'd be right.  But the fact that that paper presumably went through multiple referees before publication suggests to me that maybe you're wrong, and that maybe you aren't really applying the standards that SciPy has used for papers like this.

Anyway, regarding the mix of prose versus numbers, it seems like you probably overlooked two of the three numerical tables in my paper, and regardless, the main claimed advantage for ObjArray is that it lets programmers write simple code using ordinary Python objects and to do vectorized operations on them using familiar OOP dot idioms.  This makes a real difference to programming simplicity, debugging and quality of life, but none of that is easy to quantify, hence the need for prose.  I mean, ideally one could quantify this by sitting down a bunch of programmers to do various tasks with the different approaches and measure how long it takes them to get bug-free solutions, or ask for Likert-scale ratings of how pleasant the programming experience was, but clearly such experiments should not be required in a paper like this.  I might be able to get some loose approximation to this, though, by providing code snippets which can illustrate the hard-to-quantify improvement you get from using dot-idioms.  I will consider including this, and also consider whether there are other good ways of specifying when, and when not, it would be good to use ObjArray.

davclark:
I have updated the gist.
Disclaimer:
The materials in this e-mail are private and may contain Protected Health Information. Please note that e-mail is not necessarily confidential or secure. Your use of e-mail constitutes your acknowledgment of these confidentiality and security limitations. If you are not the intended recipient, be advised that any unauthorized use, disclosure, copying, distribution, or the taking of any action in reliance on the contents of this information is strictly prohibited. If you have received this e-mail in error, please immediately notify the sender via telephone or return e-mail.

Justin-Fisher:
@davclark Thanks for updating it, and sorry again that I'd left those lines commented out in the earlier upload.  The numbers you got overall seem quite similar to the ones I got.  It looks like our computers run at similar speeds with neither being consistently faster.  Here I'll run through our numbers (attending just to order of magnitude and most significant digit) to be sure.

> Reading Functions
> ObjArray Coupled Buffer: 0.0006 ms (me) versus 0.0009 ms (you)
> Pandas Dataframe:  0.009 ms (me) versus 0.005 ms (you)
> NumPy Record Array: 0.02 ms (me) versus 0.01 ms (you)
> Dtype=Object loop via ndindex: 900 ms (me) vs 1000 ms (you)
> ObjArray Ad Hoc: 1000 ms (me) vs 1000 ms (you)

As far as I can tell, the above numbers came out very similar, with the orders of magnitude all the same and the most significant digits not all that different, especially on the slower readings.  The three fastest readings here are near instantaneous and pushing the resolution of timeit's time function so I wouldn't worry too much about the slightly larger percentage-wise differences there.  (Basically each of these is just returning a view of a specified slice of an array, so is very fast.)

> Writing Functions
> ObjArray Coupled Buffer: 0.7 ms (me) versus 1 ms (you)
> NumPy Record Array: 3 ms (me) versus 2 ms (you)
> Pandas DataFrame: 8 ms (me) versus 9 ms (you)
> ObjArray Ad Hoc: 700 ms (me) versus 800 ms (you)
> Dtype=Object, loop via ndindex:  800 ms (me) 1000 ms (you)

Again these numbers all match on orders of magnitude and are similar in most significant digit.  

> Method-Calling Benchmarks
> Dtype=object, loop via ndindex: 2 sec (me) vs 2 sec (you)
> ObjArray call with prebuffered method: 7 sec (me) vs 5 sec (you)
> ObjArray ad hoc:  8 sec (me) vs 7 sec (you)
> NumPy Record Array:  20 sec (me) vs 20 sec (you)
> Pandas DataFrame Apply:  50 sec (me) vs 50 sec (you)

As with the other tests, these numbers all match on order of magnitude and are very similar for most significant digits.  The ordinal comparisons work out entirely the same within each group.  There are some minor differences, but those are to be expected using different systems.  Overall your pattern of results is very much the same as what I reported.  I would take this to be a replication of my results on a different system.

---

The main remaining question in my mind about this is whether the results you got manually calling %timeit yourself somehow mismatch mine.  The first thing I'd note is that your trials were all using either just 1 or just 3 repetitions, whereas mine were using 10 repetitions, so yours should be much more susceptible to noise, e.g., from background processes, than mine were, especially on the near-instantaneous trials.  (To make yours do 10 repetitions, you could change your -r1 to -r10.)  

Second, when the tested code is near-instantaneous, I don't think it should be surprising that background interference might slow the execution of the code by a factor of 8, because basically any interference in the middle of something near-instantaneous will show up as a big slowdown by comparison.  I suppose some sort of caching could be involved too, though if it is, I think that would be some sort of caching built into Numpy and should affect all three of the near-instantaneous approaches equally.  The fact that the slowdown was biggest on the fastest one fits better with the hypothesis that it's just noise from background interference.  But you're right that it's an interesting question.  Since all three of these are blindingly fast, I think it's mostly just moot curiosity though.

The third thing I'd note is that your comparison between OA.product() and Dataframe DF.x * DF.y *DF.z is not at all a fair one.  OA.product() iterates through a million objects and calls a method for each of them, whereas the Pandas expression uses NumPy's vectorized multiplication function which of course is much faster.  Writing vectorized NumPy operations is equally available to coupled ObjArrays and coupled Objarrays are in fact a teensy bit faster at returning the .x, .y and .z arrays for NumPy to multiply -- that's what the reading benchmarks showed.  The point of the method-calling benchmark was to compare ObjArray's collective method-calling to the version of collective method-calling that Pandas offers in the form of 'apply'.  You do eventually test that, and your result there (53.8 secs) is very similar to my own (48.7 secs).

A fourth minor thing to note is that the relevant thing to pre-buffer in the Obj-array prebuffered calling is an array of the objects' methods themselves -- i.e., OA.product -- rather than an array of the attributes -- like OA.x, which you tested.  We're calling the .product method of each object, which then refers to that object's own attributes, so prebuffering the attributes together into an array shouldn't help with that.  Pre-buffering the methods to call does help because Python automatically interprets OA.product() first by interpreting OA.product and then calling the resulting thing's `__call__` method and pre-storing the value of OA.product eliminates the need to do that first step in a slow ad hoc fashion.  You'll see that the way I tested this was first to set `buffered_product_method = OA.product`, and then to time `buffered_product_method()`.

So, in summary, I don't think I see any cause for concern in the manual results you got.  Do you?  Thanks for delving into this!

Justin-Fisher:
In response to @mrocklin's suggestion, I have gathered together and lengthened my discussion of when it might be best to use objarray and when it might instead be best to use alternative approaches, in the new section 6, which I'll paste in here for your convenience:

> ObjArray will be useful only in cases where one has multiple objects that are usefully considered as Python objects, and whose attributes may be usefully arranged in NumPy arrays.  The most common reasons for it to be useful to arrange objects in arrays is to make use of fast NumPy functions or useful NumPy indexing.  One common reason for it to be useful to employ ordinary Python objects is that many programmers are already adept at OOP, and may even develop code with OOP and then later decide seek to integrate it with NumPy.  In addition, OOP allows flexibility for dealing with heterogeneous objects, especially ones that fall into a class hierarchy, using "duck-typing" to enable these diverse objects to be used together.  In cases where arrayed objects are instead quite homogeneous--e.g., the case in which each object has just the same three numerical attributes--there will be comparatively little reason to treat these as Python objects, rather than just as record objects or rows in a numerical array.
> 
> Ordinary Python objects typically consume more memory and processor time than would the same data in numerical arrays, so ObjArray also will be useful only in cases where space and time considerations allow this expense.  ObjArray's coupled buffers allow the full speed of numerical arrays for on-the-fly NumPy operations but incur a cost at initial coupling, which may be prohibitive.  Because altering or re-initiating a coupling is comparatively expensive, coupled buffers also may be unsuitable for arrays that will frequently be reordered, e.g., by sorting.  In these cases it may be preferable just to use numerical arrays.  
> 
> The preceding section revealed that collective method-calling in Python is especially slow in comparison to other collective array operations, regardless of whether one uses ObjArray or numerical arrays, so time considerations may weigh in favor of rewriting operations that might have been written with collective method-calling as a series of fast vectorized NumPy operations instead.  The more methods that will be vectorized in this way, the less reason there would be to retain the ability to view the data as Python objects, and eventually it may be preferable simply to view the data just as a numerical array.
> 
> These tradeoffs may be illustrated by considering different computational models of neural networks.  Some neural network models involve very large homogeneous networks, where each neuron has the same few numerical attributes, and each is fully connected to all neurons in the next layer.  This case would probably be best suited for implementation using just numerical arrays, because the large number of neurons involved will weigh in favor of preserving efficiency in memory space and processor time, and because the homogeneity will allow one-size-fits-all vectorized NumPy operations to operate upon whole arrays at once.
> 
> In contrast, other neural network models are much more heterogeneous, e.g., models that employ different classes and subclasses of neurons observed in living animals' nervous systems.  In heterogeneous neurons, some neurons possess methods or attributes (including sparse connections to other neurons) that most other neurons lack, so it would be very difficult to fit these efficiently in the same large numerical array, and it would be very difficult to write one-size-fits-all vectorized NumPy operations to operate upon them all.  This heterogeneity may be naturally represented using OOP, especially when the different neurons fall in a hierarchy of classes and subclasses.  Heterogeneous neural network models often involve fewer neurons and/or are used in circumstances where fine-grained correspondence to empirical observation is more important than running speed.  This would be a prime use case for ObjArray, which can easily accommodate diversity and hierarchical classing among neurons, and also can allow convenient dot-idiom notation for accessing common attributes or methods that are defined, perhaps in different ways, for each individual neuron. (Indeed the author's own work with the latter sort of neural network is part of what sparked the development of ObjArray.)

davclark:
@Justin-Fisher I feel I'm not getting through on this point, so I will say it as clearly as I can:

There are LARGE timing differences between runs in many of the benchmarks, which in the IPython displayed output for %timeit results in a warning that cacheing is probably occurring. So, the timings you report, while stable, are likely not representative of what one should expect for the standard use case: a single-run for most computation.

I understand the issue addressed by repeated runs. I also understand the issues created by repeated runs. I've run multiple 1-loop / 1-run tests and the timings I achieve that way are stable enough that I trust them. That is, I am pretty sure there is some cacheing or similar that is affecting the timings by at least an order of magnitude, and doing so in a way that is favorable towards ObjArray (i.e., that is skewed in your favor, though seemingly not in a way that would change the ordering of results).

davclark:
Moreover, if you think the tests are so small as to be highly unstable, why not use bigger data in those tests?

Justin-Fisher:
The issue isn't size.  You could make the arrays any size you want and the three near-instantaneous reading approaches will still be near-instantaneous.  The reason they're near-instantaneous is because these approaches already store the data in the form of a NumPy array, so to return the portion of the data that corresponds to a particular attribute is just to return a view of (a portion of) that array, and producing a NumPy view is a O(1) operation, not at all dependent on the size of the array.  (ObjArray is faster because it simply returns the coupled array; whereas the others need to slice a view out of a larger array which is a teensy bit slower but is still O(1).) There is no way to "use bigger data" to slow O(1) operations down.  I need to benchmark these near-instantaneous approaches to provide contrast to other approaches that instead store their data in some other format (e.g., as attributes within separate objects), and therefore need to iterate through all this data at O(N) in order to copy it to an array.  So that's why I include them, even though the small-fraction-of-a-millisecond differences between them would be virtually irrelevant to any realistic use scenario where other operations that are performed equally often will take much longer.

Huge speed differences across repetitions would worry me if they arose on any of the O(N) tasks, as that would suggest that on some repetitions it wasn't actually doing the iteration, and instead was just retrieving a cached result.  Since the main speed-cost is in doing iterations, it would definitely be worrisome if a benchmark test somehow managed to avoid doing an iteration.  But you got those speed differences only on the three O(1) near-instantaneous tasks that didn't involve iteration and that involve an amount of time that is clearly negligible in contrast to anything that involves iterating through all the objects.  These tasks are so fast that any interruption that happens to occur in the middle of them will significantly slow them percentagewise, so I'm not at all surprised that they end up being significantly slower percentagewise on some runs rather than others.  You didn't get any big speed differences across repetitions on the O(N) tasks that do involve iteration, so I'm not worried.

(Just to be clear, in my benchmarking, there were three O(1) tasks, the three near-instantenous reading tasks which were the three that popped up the caching warning for you.  All the other tasks are O(N) where N is the total number of objects, because each of them involves iterating through and doing something with each object.   Your kicking-the-tires notebook does not pop up the cache warning for any of these others, though to be fair, it looks like you may not have tested many of those, or at least you didn't include those tests in the file you posted.  If you do get the cache warning on any O(N) task, I'd definitely like to know about that.  But I really don't think you will.)

Justin-Fisher:
Hrm, I've looked into this a bit more, and it does seem that the slowest result on the O(1) tests is consistently the first of the ten repetitions, which certainly does fit a caching hypothesis.  I think this caching must be done by Python itself, perhaps just memory-caching relevant objects.  This sort of caching effect is not apparent for the O(N) trials, perhaps because these must iterate through all objects in the array before completing, which likely overflows the cache and/or consumes enough time that background noise ends up swamping whatever caching benefits might be at play in these slower trials. 

One option here would be for me just to report that this caching effect was observed.  Alternatively, I think I could try to gather the data in a way that thwarts caching, e.g., by tabulating one repetition of each task, then going back and tabulating another repetition of each, and so on, until I have 10 total repetitions for each tabulated.  By interspersing lots of other tasks in between the successive repetitions of one task, this would hopefully thwart any significant caching effects, or at any rate, whatever caching effects are still present would likely be ones that could show up in real applications too.  I'll let you know what I come up with.

sarostru:
Thanks @Justin-Fisher for your submission, it is indeed regrettable that a confusion about feedback mechanisms led you to miss the review period, however after discussing with @sbenthall we've concluded that we are too far past the deadline and the discussion around this paper isn't close to resolution so we will have to close this PR and not merge in the paper.

I do hope you submit again and take into consideration the number of ideas @mrocklin and @davclark  presented for how to better present the work to an interdisciplinary audience.

@davclark and @mrocklin, thanks for taking the time to re-review, it is greatly appreciated.

davclark:
@Justin-Fisher I wanted to say that you did indeed get things to a point where you exposed a lack of understanding on those fast O(1) timings. Regarding whether there is a cacheing issue or not, it actually makes sense that there would be a cacheing issue for these, as it's clearly something that would fit into CPU cache! Moreover, it really doesn't matter in the grand scheme of doing real computation - the main point is that under optimized conditions, getting the array pointer is "fast" compared to any compute you might do.

Anyway, I was in the midst of putting together a response, and I think we were getting close to sorting out my issues... and I think you're at a point where you should still submit this at some point. Maybe SciPy next year if nothing else. Perhaps by then you could have a more substantial example, such as the neural net idea you were talking about. As I think we've all agreed - in cases of data that fit cleanly into one table, your approach is likely less performant than using idiomatic pandas / numpy - and an object could be created with methods that operate on a "has-a" style dataframe. A more "schemaless" set of objects, however, could be really interesting.

I also ditched the 1-loop %timeit calls in my notebook and ran a few more things for your reference at the usual gist (I'll upload in a minute). Given how much time *I've* put into getting up to speed on what you're doing, I'm happy to benchmark things for you or talk a bit about how you could improve your test suite.

Regarding you cache question, the only thing I've seen so far is this call: `RA.x = new_vals`. No idea why that would happen.

Justin-Fisher:
Thanks for the feedback folks, especially to @kynan for carefully attending to how ObjArray's descriptors work, and to @davclark for kicking the tires and double-checking my benchmarks.  I was disappointed that @mrocklin skimmed the paper only enough to see just one of the three numerical tables it contained.  If you had actually read it, I think you would have seen that it actually did do much of the numerical stuff you were asking for.  Take care everyone, and perhaps I'll see you around.

kynan:
Apologies for missing the re-review.

@Justin-Fisher My impression from reading the paper was that you don't consider ObjArray an academic exercise but would want to get it out there for people to use. I therefore suggest engaging with the SciPy community, which you'll experience as very welcoming. For maximum exposure, consider releasing ObjArray on PyPI and / or put the source on GitHub (as already suggested by @mrocklin).

That will also put you in a better position for resubmitting it to SciPy next year or to another venue.

PR #211, opened 2016-05-31
Paper: datreant: persistent, Pythonic trees for heterogeneous data

dotsdl:
This is the proceedings paper submission for **datreant: persistent, Pythonic trees for heterogeneous data**.


suchow:
Here's some advice from Proselint (http://github.com/amperser/Proselint), a Python-based linter for prose:

```
529:140: garner.phrasal_adjectives.ly No hyphen is necessary in phrasal adjectives with an adverb ending in -ly.
37:29: needless_variants.misc Needless variant. 'extendable' is the preferred form.
195:7: typography.symbols.sentence_spacing More than two spaces after the period; use 1 or 2. Found 8 times elsewhere.
529:25: weasel_words.very Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be.
```

dfolch:
I'm a little late to the game, but wanted to provide my feedback based on the version posted a few weeks ago.  This project is very interesting. I can see the motivation behind this as I have used ```os.makedirs()``` as I traverse different simulation scenarios and store results. Having the corresponding metadata would also be great. I truly see the value of this package.

- The paper could benefit from a motivating example at the beginning. A "literature review" of other packages and other attempts at solving this problem would also help get the idea motivated.

- I am still not clear on few things...

 - When do directories get written to the hard drive?
 - Are all the intermediate directories (e.g., ```tallest```) also Treants?
 - In the ```Data``` example, is the pandas dataframe being written to the hard drive with each edit?

- The dependencies were not clear. For example, is ```map``` using multiprocessing?

- All of the pythonic syntax is great, but some is not entirely intuitive to me. I'm sure if I was using this on an actual project, and not just reading about it, the syntax would click faster.  It might be beneficial to developers of other packages, and those who want a deeper understanding of this one, to see a discussion of the motivation behind some of the syntax decisions. This would probably lead to a "gotchas" section [like pandas has](http://pandas.pydata.org/pandas-docs/stable/gotchas.html), for scenarios that lead to unexpected results.

Overall, I'm really happy to have been introduced to this, and I hope this catches on.

dotsdl:
Hi @suchow, thanks for the suggestions. Looking forward to your talk on proselint, and happy to see it put to work here. :D

dotsdl:
Thanks @dfolch for the very useful and detailed comments. Point-wise thoughts below:

> The paper could benefit from a motivating example at the beginning. A "literature review" of other packages and other attempts at solving this problem would also help get the idea motivated.

Agreed. The motivation for `datreant` originally was that nothing really like it appears to exist in the Python ecosystem, even though there do exist domain-specific packages that work with files stored in the filesystem as objects. The difference for `datreant` is that it's domain-agnostic, and goes a step further than, e.g. `pathlib` (which we use as a dependency) in making working with directories and files more object-oriented in ways computational researchers would generally find useful.

I'm going to expand the introduction to include some more discussion of the context `datreant` exists within, including how the problem it addresses is a bit orthogonal to those addressed by database solutions, specialized file formats, etc. Will this work, you think?

> When do directories get written to the hard drive?

Ah, yes, we should make this more clear. A `Tree` object is ultimately just a path, which may or may not point to an existing directory. We often make directories that may not yet exist by calling `Tree.makedirs()`, such as [here](https://github.com/datreant/scipy_proceedings/blame/2016/papers/david_dotson/david_dotson.rst#L129).

When creating new `Treant`s from a path, such as with ``Treant('the/oldest/tallest/sequoia')``, all intermediate directories are created if they don't exist, since a ``Treant`` requires a state file to exist at all.

> Are all the intermediate directories (e.g., tallest) also Treants?

They are not. We should find a succinct way of making this clearer. Treants are useful precisely because not every directory is itself a Treant, since being a Treant at all already marks the directory as being particularly interesting (and possessing useful metadata for filtering/grouping). A `Treant` is itself a subclass of a `Tree`, so all the filesystem manipulation bits of a `Treant` can be done by a `Tree`, which doesn't require planting a state file inside it to use.

> In the Data example, is the pandas dataframe being written to the hard drive with each edit?

`pandas` objects are written to the hard drive when using `__setitem__` syntax with the `Data` limb, such as with:

```python
t.data['sinusoid'] = pd.Series(np.sin(np.linspace(0, 8*np.pi, num=200)))
```
and read from disk when `__getitem__` syntax is used (or alternatively `Data.retrieve`):

```python
s = t.data['sinusoid']

# or instead
s = t.data.retrieve('sinusoid')
```
I think what we should do to make this more clear is to show what the directory structures of these Treants look like before and after storage. Although the section mentions in words at the end that the `Data` limb stores these objects within the Treant's tree, it doesn't *show* this. Do you agree, or should we do more?

**Note**: I just noticed that the example isn't clear in part because we stored `numpy` arrays in the loop but extracted `pandas` Series from that stored data. Fixing in next commit.

> The dependencies were not clear. For example, is map using multiprocessing?

Correct. We should explicitly talk dependencies throughout. We hew pretty close to using only standard library dependencies (backports of `pathlib` and `scandir` used for python2), as well as some lightweight dependencies (`asciitree`, `fuzzywuzzy`) for the `datreant.core` library. `datreant.data` has HDF5, numpy, and pandas dependencies, which some might feel make it heavyweight.

> All of the pythonic syntax is great, but some is not entirely intuitive to me. I'm sure if I was using this on an actual project, and not just reading about it, the syntax would click faster. It might be beneficial to developers of other packages, and those who want a deeper understanding of this one, to see a discussion of the motivation behind some of the syntax decisions. This would probably lead to a "gotchas" section like pandas has, for scenarios that lead to unexpected results.

I appreciate any help we can get with making syntax motivations clear. Is there a particular example that was especially hard to grok? This would help me to get a grasp of how this looks from the outside.

---

That was long. Thanks again! Will begin addressing these in new commits.

certik:
This looks like a really great project. I've written tons of Python scripts to operate on data calculated on HPC clusters to analyze it. It's a major pain to do by hand. This project seems like a good way to make this more comfortable to work with. Questions that are not clear to me:

* Does the project also offer some help with submitting molecular dynamics (MD) or other jobs to an HPC cluster?
* If I submit jobs by hand, do I first have to start a python shell and generate the Treants the first time, so that the json files get saved into each directory with the meta data? So I have to do this once, and then the post-processing scripts are using your library to operate on the data?
* Was the previous point done, so that the line `b = mds.discover(’NhaA_i2o_transitions’)` works in the MD example?
* Would you suggest for people to write Jupyter notebooks to do the post processing with your library? Or just have it as standalone scripts?
* Another issue is that our scratch file system (that holds the data files which can easily be several GB in size) is pruned often, so I save the data to a tape drive, and then load it when needed. I assume I can save it with the Treant json files, so that the post processing work out of the box, just the load from tape drive is needed.

Note: I have most of my heavy post processing code written in Fortran, but I can easily write Python scripts to call it from Python, so one can then hook it up with your library. 


The article itself is written nicely, I don't have any comments to that, besides what was said above. I think it should be published.

dfolch:
Motivation: I see what you mean about other stuff not existing for comparison. I guess I was thinking of motivating it with the current way people do this. In my case I make long file names and deep directory trees that try to encapsulate all the relevant setup decisions that went into each simulation. Then I parse the directories and file names when I come back later to analyze the mountain of results. This is generally okay in the moment, but when I put a project down for a couple months it is difficult to decode my "brilliant" naming structures. Also, my results analysis sometimes doesn't fit well with my naming structure, so it is a hassle to pull a few files from different directories, I think the tagging would help this.

Pandas read/write: So is it using ```pickle``` (or something like that?). Can I do something like ```t.data['sinusoid'].ix[0] = 999```? Would this write to the hard drive?

Syntax: The use of sets, tuples and lists for different types of queries is one example that is learnable, but not obvious to me.  The "motivation" part was relating to how to think about ```[]``` for example; should I be thinking more dictionary query or list slicing?  Were you thinking of pandas when designing this? Again, this is probably a function of me not actually using it.

Migration: Something else I just thought of. I often will move the simulation output from the cluster to my local machine for analysis or backup. Will the treants still work on the new machine?

dotsdl:
@certik thanks for the useful comments; point-wise responses below:

1. We don't address queueing systems at any level, but of course these objects could be manipulated by scripts that are part of a job submitted to a queue. My own use is along these lines by way of workflows written to work under the [Fireworks](https://pythonhosted.org/FireWorks/) framework, in which case individual workflow tasks are often written to operate on `Sim` objects that live in the filesystem. To make remote work easier, we recently merged a [`sync` method](http://datreant.readthedocs.io/en/develop/api_filesystem.html#datreant.core.Tree.sync) that uses `rsync` internally to allow syncing of trees with minimal effort, with `include` and `exclude` semantics. This will appear in the next release.

2. To work with Treants, they first must be created, yes. However, if you don't need a directory to have any metadata like `tags` or `categories`, then you could manipulate it as a `Tree` instead. The advantage of a `Treant` is that it has identifying characteristics beyond just its path, and it will persist these beyond any Python session it is manipulated with.

3. Correct. To keep the example brief, I left out how we created these `Sim` objects in the first place. I think this is still pretty true-to-life in terms of usage, though, since in practice it really is this easy to pick up and use `Sim`s/`Treant`s since they look the same as they did the last time they were used. I can make it clearer that we did this in the text.

4. I do both. I primarily use `datreant` by way of `mdsynthesis`, since my field is MD. The scripts/workflows I write do standard postprocessing steps that would be tedious to manage by hand, while I do my actual analysis and data work in Jupyter notebooks using the very same `Sim` objects. Often I do this while simulations/postprocessing are still executing in the background.

5. Since a `Treant` is defined solely by the presence of its state file, this should work just fine. I don't have a lot of recent experience with tape storage, but I recall that the only difference at the filesystem level is that you have to wait to access things (high latency)?

Let me know if I missed anything, or if there is anything you'd like me to make clearer in the main text based on these comments. Thanks!

dotsdl:
> Motivation: I see what you mean about other stuff not existing for comparison. I guess I was thinking of motivating it with the current way people do this. In my case I make long file names and deep directory trees that try to encapsulate all the relevant setup decisions that went into each simulation. Then I parse the directories and file names when I come back later to analyze the mountain of results. This is generally okay in the moment, but when I put a project down for a couple months it is difficult to decode my "brilliant" naming structures. Also, my results analysis sometimes doesn't fit well with my naming structure, so it is a hassle to pull a few files from different directories, I think the tagging would help this.

This is exactly the kind of pattern `datreant` seeks to make less painful, since many of us (myself included) use the filesystem hierarchy as a self-describing metadata scheme. I'll do my best to expand our introduction to talk about typical ways this is done without projecting too much. I think the biggest advantage of `Treant`s is that they are tolerant of *any* scheme one already uses, but they can greatly augment the utility (and sanity) of that scheme.

> Pandas read/write: So is it using pickle (or something like that?). Can I do something like t.data['sinusoid'].ix[0] = 999? Would this write to the hard drive?

Ah, I see the confusion. The convenience interface for typical data structures that `datreant.data` provides isn't that smart, and I think we do a better job explaining how it works [in the docs](http://datreantdata.readthedocs.io/en/develop/index.html) than we currently do in the manuscript. `pandas` and `numpy` objects are stored using HDF5, since this gives speed, compression, and the ability to subselect data "out-of-core", while other objects are pickled by this interface. 

The `datreant.data` module is meant primarily as a useful showcase for how to give `Treant`s perhaps complex but useful enhancements, but isn't necessary for using `Treant`s for data storage/retrieval more generally. I'll make this clearer.

> Syntax: The use of sets, tuples and lists for different types of queries is one example that is learnable, but not obvious to me. The "motivation" part was relating to how to think about [] for example; should I be thinking more dictionary query or list slicing? Were you thinking of pandas when designing this? Again, this is probably a function of me not actually using it.

Our so-called ["tag expressions"](http://datreant.readthedocs.io/en/latest/bundles.html#using-tag-expressions-to-select-members) are bit controversial (see datreant/datreant.core#65), so I understand completely. Short of implementing a mini-language for parsing tags (which could themselves be sentences, made of any characters), I went with using built-in Python data structures to encode intersections (lists), unions (tuples), and negation (sets). It's very hard-to-read, and I know of no other project that does such a thing, but it does work. They also allow one to directly work with methods (such as `Tags.fuzzy`) that return tuples of tags matching patterns.

I will add a bit more discussion motivating the design of this (admittedly strange) mechanism for filtering Treants on tags.

> Migration: Something else I just thought of. I often will move the simulation output from the cluster to my local machine for analysis or backup. Will the treants still work on the new machine?

Yes, they will work just fine when moved. `Treant`s don't internally store information on where they live, although subclasses defined in domain-specific packages (e.g. `mdsynthesis.Sim`) may store things like file paths that can break upon moving under certain conditions. One exception is that the [`Group`](http://datreant.readthedocs.io/en/latest/groups.html) is also susceptible to breakage, but that object may be removed in the future given that `Bundle` has mostly obseleted it.

I'll make this clearer, perhaps including an example of using our recently-merged [syncing](http://datreant.readthedocs.io/en/develop/trees.html#synchronizing-trees) functionality.

dotsdl:
@dfolch, @certik I have pushed a round of commits addressing your comments. Apologies for the delay. Please let me know if there remain confusing points or things I missed. Thanks!

sarostru:

dotsdl:
@dfolch, @certik is there anything additional to add?

dfolch:
Nothing more from me.

certik:
No, it looks good to me. Thanks!

dotsdl:
Awesome! Thanks everyone! I suppose we're ready for the [ready](https://github.com/scipy-conference/scipy_proceedings/labels/ready) label!

sbenthall:
Ready label assigned. thanks everyone!

PR #215, opened 2016-05-31
Paper: Validating Function Arguments in Python Signal Processing Applications (Pedersen et al.)

ppeder08:
Submission of the paper "Validating Function Arguments in Python Signal Processing Applications".


jakevdp:
Hi! I'm going to be reviewing your paper. I'll do some inline comments as I read through, and then a final comment at the end so you know I'm finished.

Carreau:
I'm reviewing too. I've notes on the printed versions I need to copy here.

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+function argument validation can be performed using method preconditions. In
+the following outdated example, the function argument, ``a``, of the function,
+``exemplify`` is validated to be a real scalar in the range :math:`[0;1]`:
+
+.. [#] See http://www.nongnu.org/pydbc/
+
+.. code-block:: python
+
+   import dbc
+   __metaclass__ = dbc.DBC
```
jakevdp: You might mention explicitly that this is Python 2.2, and this method of declaring a metaclass is long defunct (I believe?)

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+
+   def exemplify(a: float):
+       pass  # do something
+
+   exemplify(0.5)
+
+As for validation strategy, the following observations are made:
+
+1. The validation scheme of MyPy requires specifying a static Python type or a
+   union of static Python types. This is hardly surpricing for a static type
```
jakevdp: surpricing -> surprising

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+   checker.
+
+2. As mentioned, the syntax of annotations is given by PEP 3107, and the format
+   of the type hints is given by PEP 484 making the type hints very explicit
+   and readable although a less well-known feature of Python.
```
jakevdp: Perhaps mention relevant versions of Python here

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+numerical data are passed to any number of functions that inherently impose
+limitations upon that data. If such functions do not validate their arguments,
+these limitations may be violated without raising exceptions leading to
+potentially erroneous results. Thus, explicit validation may not only spare the
+user a lot of frustration by providing useful exceptions but may also prevent
+erroneous results and thereby ensure the credibility of works in scientific
+computing.
+
+The usage of explicit function argument validation could be considered
+"unpythonic" as it goes against dynamic typing and duck typing which relies on
```
Carreau: I would break the sentence here after `duck typing`, otherwise I read it as (duck typing rely on documentation) which is technically untrue, but I see what you mean.

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+validation strategy. For an early implementation, any application-driven data
+type will allow only a limited set of Python data types. This does, however,
+not mean that the application-driven data type is limited to a few Python data
+types. Rather, more Python data types may be added along the way as long as
+they provide the necessary attributes with the desired interpretation. Thus,
+effectively, the suggested validation strategy can be considered less strict
+than static type checking but more strict than duck type checking.
+
+The numerical trait of the Numtraits package has an interesting approach which
+is compatible with the concept of application-driven data types. The numerical
```
jakevdp: It strikes me that there is one case where this is not compatible: different matrix representations. I think a true application-driven approach would allow both sparse and dense representations, and I believe Numtraits would reject sparse forms. That's probably a minor point though...

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+capabilities of these solutions vary greatly from type, attribute, and value
+checks to fully customisable checks. Most of the solutions do, however, seem to
+have an interface which relates to the data model used by Python and therefore
+translates to Python check in a straighforward way.
+
+Unfortunately, there are a number of shortcomings with the existing validation
+strategies as implemented in the existing Python packages with validation
+capabilities; in particular in signal processing applications. Some of the
+existing solutions lack generality, some do not promote readability, and some
+are inconvenient to use. However, the primary issue is that the validation
```
Carreau: Even if you don't develop it that much further in the paper, I think speaking about _performance_ in the intro might be welcome. As you point out later, your library have a way to disable the checks globally.

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+.. code-block:: python
+
+   sequence = (0., 1.)
+   do_something(sequence)
+
+The issue is that a number of Python types represent sequences, and a number of
+Python types represent numbers. This could be accounted for in the example, but
+the point to stress is that the programmer should not have to know about every
+single Python type, nor should he or she have to explicitly list a large number
+of Python types for each validation check.
```
Carreau: I think here might be a good time to at least mention [`ABC`s](https://docs.python.org/3/library/collections.abc.html), that don't solve the problem in your case, and are underutilized, but are not irrelevant IMHO.

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+
+Decorators, as used by PyValid and PyContracts, are a well-known and general
+Python construct. However, it is not immediately apparent if something goes on
+"under the hood", and the pythonic approach is to specify the validation scheme
+of all function arguments in a single decorator call, both of which affect
+readability. Therefore, decorators are not suggested as a Python construct to
+use here.
+
+The suggested Python construct values explicit over implicit, does not look
+"magical", and promotes readability. The suggestion is to define a nested
```
jakevdp: I'd push back here regarding "magical". To my eye, the nested function pattern looks much more magical than, say, a decorator. A decorator is a common pattern that most reasonably experienced Python programmers could be expected to know. On the other hand, extracting keyword arguments from an embedded function using the calling scope feels a bit hackish. Perhaps you could motivate this suggestion further?

papers/patrick_pedersen/patrick_pedersen.rst
```diff
@@ -0,0 +1,900 @@
...
+           raise TypeError('Descriptive message.')
+
+       return a[0]**0.5
+
+Clearly, this approach to validation is along the lines of duck typing as it
+explicitly checks for the presence of the required attribute. In the example,
+``a`` may hold values of any type that defines the ``__getitem__`` attribute.
+Unlike with the first way to test for validity, the validation in the above
+example is not restrictive enough as already explained using the example with
+the dictionary. Furthermore, the power operation of ``**`` does not as such
```
Carreau: That seem untrue for pow, [`int.__pow__` and `float.__pow__`](https://docs.python.org/3.6/library/operator.html) exists. Yes you can't do it with a literal, but it still exists. You can change for `math.sqrt` if you like, or just footnote you are aware of `__pow__` but want to keep exemple simple. You can also switch to `0.5**a[0]` also.

jakevdp:
I've finished my inline comments. Overall, a very useful writeup! A couple global comments:

1. I think it would be useful to explicitly list in the abstract which current solutions you compare against.
2. I'd emphasize more in the abstract that the purpose of this paper is to:

    a. define the concept of *application-driven data-types*
    b. discuss current argument validation schemes in light of this definition
    c. discuss why you went the particular route you did in ``magni``

    In addtion, I think it would help to re-structure the paper a bit. For example, the discussion of application-driven data-types is central to the paper, yet appears only as a subsection under "Validation in Python at a glance"

2. The use of the phrase "the recommended solution" comes across as a bit heavy-handed. I think there is some room for people to disagree with your recommendations, and I'd like to see a more fleshed-out discussion of why you think the nested function approach is superior (currently it comes across as an assertion rather than a discussion).

Overall very nice writeup. I enjoyed reading it, and the background on various approaches to validation is extremely useful.

suchow:
Here's some advice from Proselint (http://github.com/amperser/Proselint), a Python-based linter for prose:

```
552:46: cliches.write_good 'Needless to say,' is a cliché.
162:11: weasel_words.very Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be. Found once elsewhere.
```

jakevdp:
Quick heads-up: the deadline for review is July 10th, but I'll be out-of-touch after July 6th. If you want my feedback/sign-off on this, it will have to happen before then.

dfolch:
Sorry for coming in late on this. This is some feedback from the version from a few weeks ago. This project addresses a very important topic. Personally, I think it is important for users to get meaningful feedback when they pass "incorrect" values to a function. Oftentimes the traceback comes up in some obscure place in the code making it difficult to figure out that the problem traces back to user input error; this is especially true for novice users.

- The paper does not mention the ```try``` and ```except``` approach, which seems to be the approach advocated by many python purists.

- Might be good to see the syntax of the error message returned to the user.  

- I encourage you not to use the ```from package import *``` syntax. This creates the sensation of "magic" that you were trying to avoid.

- Why was a semicolon used to separate a range instead of a colon?

- Some description of the dependencies would be good. matplotlib and scipy were especially surprising. 

- My only concern is the idea of defining a function each time a function is executed. This seems like an unnecessary performance hit. Performance testing could be done to see these costs. I see two alternative scenarios to avoid this while maintaining the spirit of the proposed approach.

 - The function of interest calls an outside function to run the validation. Besides not rebuilding the validation function each time, it also allows that validation function to be reused for multiple functions.
 - Build parallel functions, where the validation and substantive code is split into "user" and "base" functions.  The user function has all the validation; if everything passes, then the user function passes the parameters to the base function. It also allows the user who doesn't want the cost of the validation code to just call the base function directly.

Overall, I think this is an important issue. More voices discussing this topic will be a benefit for python.

Carreau:
I'm wondering if @ppeder08 receives notifications, opening the Pull request seem to be his only action on github ever. 

Do you think I should ping him by mail to see if he needs help with Github ?

Carreau:
> I'm wondering if @ppeder08 receives notifications, opening the Pull request seem to be his only action on github ever.
> 
> Do you think I should ping him by mail to see if he needs help with Github ?

Did mail authors. Contact author is unavailable until Monday 10, and will try to push address review as soon as possible.

sarostru:

ppeder08:
I apologise for being unavailable during the original period for revising the paper. Thank you, @jakevdp, @Carreau, @dfolch, and @suchow for your comments and suggestions. I am currently incorporating these in the paper, and I will summarise how, overall, everything has been addressed, when I am done.

ppeder08:
@jakevdp, I have incorporated all of your suggestions, and I make the following comments on a few of the suggestions:

- You are quite right that the concept of application-driven data types is central to the paper, and so I have restructured the paper by making this into a section and moving a paragraph from the previous section.

- I have removed the part about nested functions being less "magical". And as suggested, I have motivated the suggestion of nested functions without arguments further.

- Regarding "the recommended solution" coming across as a bit heavy-handed, my intention was to make a suggestion/proposal rather than recommendation. There can be good reasons for deviating from the suggestion such as avoiding the overhead of defining the same function over and over again. As noted under the previous bullet, I have, however, added a bit more on the suggestion of using the nested function.

ppeder08:
@Carreau, I have incorporated all of your suggestions, and I make the following comments on a few of the suggestions:

- I added a short sentence about performance in the introduction but would prefer not to put too much emphasis on this as I have deliberately chosen not to include discussions or considerations concerning this topic.

- Mentioning ABCs is indeed a good idea (and indeed an ABC is used in the reference implementation) as they are relevant, and I agree that they are underutilized.

ppeder08:
@dfolch, I have incorporated some of your suggestions, and I make the following comments on your suggestions:

- Regarding the `try` and `except` approach, I would argue that this is covered by the following "the most pythonic idiom is to clearly document what a function expects and then just try to use whatever gets passed to the function and either let exceptions propagate or catch attribute errors and raise other exceptions instead".

- I have rewritten the code examples to show the syntax of the error messages returned to the user.

- I have replaced the `from package import *` syntax with more explicit imports.

- An implementation could choose to use a comma, a semicolon, a colon, or possibly something else to separate a range. The reference implementation uses a semicolon as a matter of preference, but it could as well have used any of the other.

- I have added an elaboration of the dependencies of the reference implementation.

- I agree that there is a performance hit in defining a function each time a function is executed. Performance testing has not been done as I have deliberately chosen not to include discussion or considerations concerning this topic in order to focus on the main message.

  * I have added a brief note on having a global rather than nested function for validation to the paper.

  * I would discourage using parallel functions because this could cause confusion, and because I would rather define a single function and provide a way to disable validation when the performance overhead is deemed too costly. However, using parallel functions is, indeed, a possibility and anyone is free to implement a strategy which deviates from the suggestion.

Carreau:
Thanks @ppeder08 I'm just back from SciPy, and will have a look at your comments and re-read the paper.

Carreau:
> I added a short sentence about performance in the introduction but would prefer not to put too much emphasis on this as I have deliberately chosen not to include discussions or considerations concerning this topic.

That's fine, I think that's adding this sentence convey that you are aware, but it's not the focus of this article. 



> Mentioning ABCs is indeed a good idea (and indeed an ABC is used in the reference implementation) as they are relevant, and I agree that they are underutilized.

Thanks. 

> 	Merged upstream changes into current branch

I think you merged `master`, though your PR target the branch `2016`, which leads to extra-diffs.. 
I'm not sure if we should  try to unmerge. @stefanv ? @sbenthall ? or update 2016 to get the changes from master which will lead to a cleaner diff.

sbenthall:
@jakevdp, @Carreau, @dfolch can you please provide feedback as to whether you recommend this paper for acceptance?

Carreau:
I'm fine getting this in, there might just need to be a cleaner rebase, as it modify a few things that are not the article as said in previous comment:

> I think you merged master, though your PR target the branch 2016, which leads to extra-diffs.. 
I'm not sure if we should try to unmerge. @stefanv ? @sbenthall ? or update 2016 to get the changes from master which will lead to a cleaner diff.

sarostru:
We can do a rebase and maybe make it a new PR before merge to keep things clean.

ppeder08:
For clarification, I originally forked the repository (named `upstream` for my repo) and worked off my `origin/2016` branch (as instructed). Then, with one of my latest commits (dea7934), I merged changes from the `upstream/2016` branch into my `origin/2016` branch (to ensure that layout had not changed). That is, I have **not** merged the `upstream/master` branch into my `origin/2016` branch directly.

The following may clarify:

```
$ git diff upstream/2016..origin/2016 --numstat
1052    0       papers/patrick_pedersen/patrick_pedersen.rst
90      0       papers/patrick_pedersen/references.bib
```

```
$ git diff upstream/master..origin/2016 --numstat
1052    0       papers/patrick_pedersen/patrick_pedersen.rst
90      0       papers/patrick_pedersen/references.bib
0       4       publisher/build_paper.py
```

Indeed, the only changes from `upstream/master` to `upstream/2016` are to `publisher/build_paper.py` (I haven't fetched any commits since July 12):

```
$ git diff upstream/master..upstream/2016 --numstat
0       4       publisher/build_paper.py
```

Let me know if I should make any changes (e.g. a rebase or something else) for the time being.

sarostru:
We need to be able to merge it in without conflicts, if you could try to figure out why there are conflicts on this PR then it would be really helpful.

There shouldn't be any changes to files outside of your papers directory, if you revert any of those changes then it should be conflict-free.

ppeder08:
Commit 6ff824dc now reverts the changes of dea7934a - leaving only changes to files in /papers/patrick_pedersen.

Carreau:
It's indeed conflict free now ! Thanks !

PR #225, opened 2016-07-11
Print bibtex errors in case of an error.

rainwoodman:
Print the errors such that the user is not scratching their head about what went wrong with their bibtex source file.


sarostru:
good idea

stefanv:
Pull requests at 3:19 and merges at 3:20.  I approve :)

PR #250, opened 2017-04-23
2016 --> 2017

mpacer:
@katyhuff mentioned this needed to be done, should have gotten everything. 

Also unless new content is addressed a simple find and replace can work for 2017 → 2018 next year (because there were no conflicting instances of 2017).


katyhuff:
LGTM

PR #256, opened 2017-04-30
First steps toward notebook submission; it works! Kinda!

mpacer:
So, if this were to be merged and you were to just dump all of your rst into a Raw_NBConvert cell, this works perfectly. 

The problem is covering features of rst that are not easy to support in the notebook itself in cells that are not Raw_Nbconvert cells. 

Particular challenges that are going to be nearly impossible to deal with will be: 
* footnotes
* LaTeX text commands (including references)
* generic custom directives 
* inline figure formatting 
* general captioning and labelling
* …

Partial support 
* figure captions (as indicated with alt-text) 
* tables
* equations w/ labels (technically on first render… labels pose no issue right now see https://github.com/jupyter/notebook/issues/2461) 
* citations (technically we have a mechanism for specifying citations, but currently it's not actually outputting to rst appropriately)
* other language code cells (e.g., right now they will be given the `.. code::<your kernel>` directive even if they are prefixed with an ipython magic like `%%javascript`)
* …

But a good number of things already work! 

So technically this would be reasonable to merge in to meet the goal of "notebook as a submission format". Because… there is support submitting _a_ `your_file.ipynb`; it's just not yet a gigantic improvement over submitting using raw `rst`. And I imagine unless you are editing in nano (or another minimal editor) the notebook wouldn't be that much better to edit a bunch of rst in either. 

But hey! I thought it would take most of the week to get even this far!

cc: @stefanv @dalippa @katyhuff @deniederhut huzzah!

cc: @Carreau @takluyver @minrk — any thoughts on which of these are potentially soluble among the notebook UI issues? 

Also, there's now support for a debug feature if you call `python build_paper.py <your_directory --debug` rather than using the make script. That should help for iterating through different versions of the nb → rst conversion over time.

And with _that_ I'm taking a break for a little while.


mpacer:
NB: just took a nap and may have had a breakthrough don't merge yet more to come!

stefanv:
@mpacer Do I understand correctly that the eventual goal is to submit a notebook with very little restructured text special markup?

mpacer:
Not quite! Just one that allows the use of markdown markup (when appropriate) without _breaking_ restructured text markup.

stefanv:
I guess my question was: can you submit a notebook that looks like a notebook (not a bunch of pasted in RST), and have it compile?  Because otherwise there's probably not much point moving away from RST.

mpacer:
Yes, that's the aim and that's my criterion :).

deniederhut:
:+1: for correct use of the singular form

stefanv:
@mpacer What is the status of this PR?

mpacer:
@stefanv This is still in process and is in bad need of a rebase. I'm going to abandon the `,,,` `,,,` approach and instead will see how far I can get using pandoc's attribute syntax for much of this.

PR #262, opened 2017-05-09
this callout in the readme was unneccessary.

katyhuff:
this callout was unneccessary. I think it was added last year by unnecessarily grateful proceedings chairs

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,7 +131,7 @@ su -c `dnf install python-docutils texlive-collection-basic texlive-collection-f
 
 ## Build Server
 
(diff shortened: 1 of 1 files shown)


mpacer:
Honestly — it seems a little self-congratulatory when put this way. I'd vote to strike the "thanking" parts altogether. Also we can probably strengthen the language from may → should regarding the ability to get a pdf from the server, as that's the reason the server exists.

katyhuff:
Thanks for your comments @mpacer . Fixed!

mpacer:
LGTM merging

Issue #245, opened 2016-08-30
Paper Generation Error with :corresponding: tag


dalippa:
I was working on #224 and found this while reconciling changes:

```
:author: David L
:corresponding: David L
:email: david@l.com
:institution: RealMassive, Inc.

:author: Joe Schmo
:corresponding: Joe Schmo
:email: joe@schmo.com
:institution: Schmo, Inc

:author: A L
:corresponding: A L
:email: a@l.com

:author: Jason V
:corresponding: Jason V
:email: jason@v.com
:institution: RealMassive, Inc.

:author: D L
:corresponding: D L
:email: d@l.com
```

The :author: tag works, but only for the first corresponding author. Adding :corresponding: and :author: tags also cause the error, as does the 00_vanderwalt.rst file on master and 2016 branches.

Stack trace:

```
Traceback (most recent call last):
  File "publisher/build_paper.py", line 229, in <module>
    build_paper(paper_id)
  File "publisher/build_paper.py", line 214, in build_paper
    rst2tex(in_path, out_path)
  File "publisher/build_paper.py", line 88, in rst2tex
    settings_overrides=settings)
  File "/usr/lib/python2.7/dist-packages/docutils/core.py", line 414, in publish_string
    enable_exit_status=enable_exit_status)
  File "/usr/lib/python2.7/dist-packages/docutils/core.py", line 662, in publish_programmatically
    output = pub.publish(enable_exit_status=enable_exit_status)
  File "/usr/lib/python2.7/dist-packages/docutils/core.py", line 219, in publish
    output = self.writer.write(self.document, self.destination)
  File "/usr/lib/python2.7/dist-packages/docutils/writers/__init__.py", line 80, in write
    self.translate()
  File "/usr/lib/python2.7/dist-packages/docutils/writers/latex2e/__init__.py", line 247, in translate
    self.document.walkabout(visitor)
  File "/usr/lib/python2.7/dist-packages/docutils/nodes.py", line 174, in walkabout
    if child.walkabout(visitor):
  File "/usr/lib/python2.7/dist-packages/docutils/nodes.py", line 174, in walkabout
    if child.walkabout(visitor):
  File "/usr/lib/python2.7/dist-packages/docutils/nodes.py", line 174, in walkabout
    if child.walkabout(visitor):
  File "/usr/lib/python2.7/dist-packages/docutils/nodes.py", line 166, in walkabout
    visitor.dispatch_visit(self)
[...]

PR #265, opened 2017-05-20
Issue 245: Paper Generation Error with :corresponding: tag

dalippa:
Show an example of this working now with the latest mainline.

https://github.com/scipy-conference/scipy_proceedings/issues/245


dalippa:
@katyhuff @mpacer Please check to see if this functions as y'all want. Right now, the email addresses are concatenated and grouped together under "Corresponding Author".

mpacer:
@dalippa and I communicated and this should be addressed by now and doesn't produce any issues when building.

PR #273, opened 2017-05-29
Mention inline figures expectation in README

mpacer:


mpacer:
@katyhuff @dalippa @deniederhut Is that text clear?

mpacer:
Closes #271.

deniederhut:
Thanks for clarifying this! I might be a little more explicit and say something like "figures and tables should be positioned inline, close to their explanatory text".

mpacer:
Thank you for the suggestion. Done.

deniederhut:
👍

deniederhut:
It looks like the template paper has all of the figures and tables listed at the end -- for next year, this might be a thing we want to update.

mpacer:
Not all of the images are placed at the end, only those that were too large to put in a per column format

image

However, I **do** need to think more about how to handle images that won't fit in a single column.

deniederhut:
Well, this is definitely an improvement for now.

PR #277, opened 2017-05-30
Paper: Devito

mlange05:
Title: Optimised finite difference computation from symbolic equations

Authors: M. Lange, N. Kukreja, F. Luporini, M. Louboutin, C. Yount, G. Gorman

Abstract:
Domain-specific high-productivity environments are playing an increasingly important role in scientific computing due to the levels of abstraction and automation they provide. In this paper we introduce Devito, an open-source domain-specific framework for solving partial differential equations from symbolic problem definitions by the finite difference method. We highlight the generation and automated execution of highly optimized stencil code from only a few lines of high-level symbolic Python for a set of scientific equations, before exploring the use of Devito operators in seismic inversion problems.


nicholasmalaya:
@katyhuff I'm interested in reviewing this paper, please feel free to assign it to me. @mlange05 I'll try to get you some initial comments from a read-through soon.

katyhuff:
Go for it!

wd15:
The work is highly significant addressing the need to easily pose PDEs
in Python and then generate an efficient numerical
implementation. There is currently a lack of good tools that allow
non-experts to easily pose and solve PDEs natively and interactively
within Python, but also solve those PDEs efficiently. Devito addresses
this issue. Moreover, using Sympy notation to pose PDEs is a sensible
approach that generates a de facto standard for literate programming
of PDE problems, which is greatly needed.

The paper is well written and for the most part clear and easy to
understand. The authors should note that many of the readers of this
paper may not fully comprehend the "Adjoint Test" section of the
paper. A more gentle introduction to the "Adjoint Test" section would
be helpful.

The main criticism of the paper is that the disadvantages of the
approach and limitations of using Devito to pose and solve PDEs is not
discussed and needs to be addressed.

The following is a list of sections where the limitations of the
approach need to be made more clear.

 * The first example, "linear convection", does not highlight that the
   solution in Fig. 2 is not a good solution to a traveling square
   wave. It is a good example for demonstrating how to use Devito, but
   is not a realistic scheme for practical solutions of the advection
   equation.  In general a higher order scheme is required as well as
   a limiting scheme or other specialized scheme to preserve the
   traveling wave. This issue should be discussed. Furthermore, a
   discussion of the limitations or difficulties raised in using
   Devito for more complex schemes is required.

 * Again, the second example is not a good example of using Devito
   either as it is a steady state diffusion problem, which is much
   better resolved with an implicit scheme that Devito does not
   support (or does it?). This issue should at least be discussed. The
   topic of explicit versus implicit should be raised in the paper as
   well. Some discussion of how Devito might move towards supporting
   implicit schemes would be useful.

 * The paper should discuss how Devito developers might address the
   issue of posing boundary conditions symbolically rather than having
   to deal with indexing issues.

 * All three examples only involve one equation. The paper should address
   the capabilities and limitations of using Devito for coupled PDEs.

A separate section discussing the current limitations of Devito and
plans for how the developers might envision addressing some of these
issues would be useful. Some consideration should be made to address
readers that are trying to understand the scope and limitations of
Devito and whether they should invest the time in learning and using
Devito for their own unique PDE problems.

The paper would be better if there are clear links to full code
examples (in the text) for all three of the examples in the paper.
Also, Docker containers for all or each example would allow users to
run the examples in the paper in future when dependencies become
outdated.

nicholasmalaya:
I do not have a great deal of commentary beyond @wd15 (Daniel's) excellent review. I defer to him for later decisions on acceptance for the proceedings.

The paper discusses the DEVITO library, a python framework assisting finite difference implementations of partial differential equations. This is a high impact work that should be of general interest to the python community, and SciPy 2017 in particular. In general the paper is readable, and well written. 

In many ways DEVITO shares philosophical similarities to FEniCS and Firedrake, both of which are python libraries that permit symbolic specification of an operator with optimized (high performance) kernels that implement the underlying PDEs. Unlike these frameworks, which use a finite element discretization, DEVITO is based on stencil methods, e.g. finite difference. This is my greatest principle criticism of the paper, which does not provide an motivation for the use of FD versus FEM. I do not believe it is appropriate to assume that the SciPy audience will all appreciate trade-offs between numerical discretization schemes. Thus, I strongly recommend at least a paragraph providing motivation for the need of a finite difference library be provided in the introduction. A small table highlighting the differences (or pro/con) between frameworks would be highly interesting.

Smaller comments: 

- Figure 4 states 'after convergence' but I do not see anywhere in the text where the convergence criterion is defined. Perhaps it is `while l1norm > 1.e-4`? Regardless, this should be explicitly noted in the text to assist users with reproducibility. 

- The axes, plot labels, and title of figure 6 and 7 are too small. The size is not readable. 

- The vertical dotted line in plot 6 and 7 are labeled 'SO'. In general it is desirable for a plot to be largely self-documented, in that the figure caption should define what SO is, for instance. Otherwise the reader must flip between pages to discover what this means. 

- Finally, can the authors comment on distributed computing environments (likely in the conclusions)? Would it be possible to expand DEVITO to support this? MPI is supported in FEniCS, and while DEVITO supports a shared memory paradigm (openMP), a large swath of HPC users are interested in multi-node parallelism.

mlange05:
Review response
=============

We would like to thank all reviewers for their very clear and insightful comments. In response we have revised several sections of the paper to clarify the respective points. We hope that these changes have both made the paper clearer and addressed the reviewer's concerns. Responses to individual reviews are provided below. 

Reviewer 1 (@wd15)
===============

> The main criticism of the paper is that the disadvantages of the
approach and limitations of using Devito to pose and solve PDEs is not
discussed and needs to be addressed.

> A separate section discussing the current limitations of Devito and
plans for how the developers might envision addressing some of these
issues would be useful. Some consideration should be made to address
readers that are trying to understand the scope and limitations of
Devito and whether they should invest the time in learning and using
Devito for their own unique PDE problems.

We agree that some of the natural limitations of Devito were not discussed sufficiently in our original submission. We have therefore added a "Limitations and Future Work" section to the final discussion to clearly express the current state of development and potential future features that interested users might expect. 
However, with regards to posing and solving PDEs in general, we would like to point out that the primary motivation for Devito is to enable the efficient implementations of finite difference solvers, in a similar fashion to embedded DSLs for finite element methods like FEniCS or Firedrake. As such, we regard Devito as one of several tools to efficiently solve PDEs, leaving the choice of numerical scheme to the user, as it is highly problem-dependent. We have adjusted the introduction to better capture this sentiment, since we believe that a complete discussion and comparison of numeric approaches to solving PDEs is beyond the scope of this paper. 

> The first example, "linear convection", does not highlight that the
solution in Fig. 2 is not a good solution to a traveling square
wave. It is a good example for demonstrating how to use Devito, but
is not a realistic scheme for practical solutions of the advection
equation. In general a higher order scheme is required as well as
a limiting scheme or other specialized scheme to preserve the
travelling wave. This issue should be discussed. Furthermore, a
discussion of the limitations or difficulties raised in using
Devito for more complex schemes is required.

We agree that the initial condition from the original tutorial does lead to an overly diffusive solution that is not ideal if presented on its own. We have therefore updated the example with a smooth initial condition to avoid the numerical error due to discontinuities in the original "hat function". We also added a small discussion on the effective diffusion due to numerical errors and a reference to the LeVeque textbook for further reading.

> Again, the second example is not a good example of using Devito
either as it is a steady state diffusion problem, which is much
better resolved with an implicit scheme that Devito does not
support (or does it?). This issue should at least be discussed. The
topic of explicit versus implicit should be raised in the paper as
well. Some discussion of how Devito might move towards supporting
implicit schemes would be useful.

We again highlight that the primary objective of the presented examples is the illustration of the Devito API and accept that an implicit scheme might be more appropriate in a realistic scientific context. We have adjusted the discussion of the second example to reflect that and highlight a possible future integration with non-linear solver libraries to enable implicit schemes in the future.

> The paper should discuss how Devito developers might address the
issue of posing boundary conditions symbolically rather than having
to deal with indexing issues.

A note on future symbolic extensions for boundary conditions, has been added to the Laplace example, as well as the Future Work section of the paper. We believe that further details on the potential implementation of generic abstractions for FD boundary conditions would be premature at this point, since  there is no overarching concensus in the community on this topic. 

> All three examples only involve one equation. The paper should address
the capabilities and limitations of using Devito for coupled PDEs.

Coupled equations are actually supported with the current Devito API, due to operators allowing multiple SymPy expressions as input. We have highlighted this fact in the description of the Laplace example, which relies on the same mechanism to add boundary conditions. Moreover, a complete example of the coupled Burgers' equation is available as a notebook on the website. However, due to the space limitations of this paper and the additional complexity of the Burgers' example we have chosen to demonstrate this mechanism using the steady-state Laplace problem instead.

> The paper would be better if there are clear links to full code
examples (in the text) for all three of the examples in the paper.
Also, Docker containers for all or each example would allow users to
run the examples in the paper in future when dependencies become
outdated.

The examples presented in this paper are provided as jupyter notebooks on the website. We have provided explicit links to each, as well as including them in the packaged version of the code attached to the updated Zenodo reference.

mlange05:
Reviewer 2 (@nicholasmalaya)
=========

> This is my greatest principle criticism of the paper, which does not provide an motivation for the use of FD versus FEM. I do not believe it is appropriate to assume that the SciPy audience will all appreciate trade-offs between numerical discretization schemes. Thus, I strongly recommend at least a paragraph providing motivation for the need of a finite difference library be provided in the introduction. A small table highlighting the differences (or pro/con) between frameworks would be highly interesting.

As pointed out above, we believe that a comparison of numerical schemes is highly problem-dependent and much too complex to be covered rigorously in this paper. Instead we regard Devito as an additional tool in the scientific Python ecosystem that is specific to a particular widely used numerical method. We have adjusted the introduction accordingly and added an explicit statement for this sentiment to the new Future Work section.

> Figure 4 states 'after convergence' but I do not see anywhere in the text where the convergence criterion is defined. Perhaps it is while l1norm > 1.e-4? Regardless, this should be explicitly noted in the text to assist users with reproducibility.

This section has been rewritten to explicitly explain and state the convergence criterion. 

> The axes, plot labels, and title of figure 6 and 7 are too small. The size is not readable.

The size of the figure has been increased.

> The vertical dotted line in plot 6 and 7 are labeled 'SO'. In general it is desirable for a plot to be largely self-documented, in that the figure caption should define what SO is, for instance. Otherwise the reader must flip between pages to discover what this means.

We have provided additional information on setup, including an explanation of the "SO" abbreviation in the caption.

> Finally, can the authors comment on distributed computing environments (likely in the conclusions)? Would it be possible to expand DEVITO to support this? MPI is supported in FEniCS, and while DEVITO supports a shared memory paradigm (openMP), a large swath of HPC users are interested in multi-node parallelism.

This is a very good point that was missed in the original submission. One of the key drivers behind the ongoing YASK integration, beyond "stencil folding", is that YASK comes with built-in MPI support and hybrid MPI-OpenMP support. This is now mentioned in the Future Work section.

mlange05:
Reviewer 3 (@cyrush)
=================

> I do think the other reviewers have valid points about FEM vs FD. I viewed this as an illustrative example for how to use Devito, not a definitive example for how to best solve the CFD PDEs, but I agree that should be explicitly stated as to not lead folks astray.

The use of the CFD examples as an illustration of the API was indeed the main intention of the paper, and we have tried to make point clearer through changes in the introduction and the new Limitations and Future Work section. We have also tried to underline that the choice of numerical method is a highly problem-dependent issue, a full exploration of which would go beyond the scope of this paper. 

> I also recommend adding a reference or some text to introduce the Seismic Inversion application and some context for the Adjoint Test would help.

We have tried to make the meaning of the adjoint test more explicit and added a reference to Virieux and Operto, 2009 for further background on its use in seismic inversion methods.

We have also tried to fix all further typos and text/label issues.

wd15:
@mlange05, nice work. I'm happy with the changes.

nicholasmalaya:
I am also happy with the edits, and find the paper improved. Thanks for the hard work @mlange05 .

cyrush:
@mlange05 -- looks great, thanks for addressing our suggestions!

nicholasmalaya:
Looks as if all the reviewers are satisfied: @scipy-proceedings-bot you are cleared to merge.

katyhuff:
Sorry for the confusion. That bot isn't one of ours (perhaps it's for euroscipy)  @scipy-conference/proceedings is the name you're after (my fault for emailing the wrong team name.) Thanks for all the hard work, folks!

mlange05:
Just FYI, I pushed a few minor amendments to the acknowledgements and a fix to some example code. This should not really affect the paper much, but hopefully will be included in the final camera-ready version.

mpacer:
Not merged yet so that should be fine!

PR #280, opened 2017-05-30
updates timeline

katyhuff:
as discussed in call. pushes things about a week (ish.)


mpacer:
My one suggestion would be to instead of saying "submit your review by 21st", "submit a comprehensive review by the 21st". Otherwise looks good to go.

PR #285, opened 2017-05-30
Paper: BespON

gpoore:
Here is my first version.

I will have a few revisions and some comments about LaTeX layout and code typesetting shortly.


mpacer:
Just as a heads up, I appreciate your wanting to use the package that you've written for extending `fancyvrb` as well as the custom tweaks that you are applying. I deeply appreciate that you care about typography; we're on the same page about its importance. 

*Aside*: `fvextra` looks like an interesting package that I'm going to have to look into more closely. 

But, for the purposes of review, would you please make your paper buildable on the current build system?

As for supporting these kinds of TeX features… I'll do what I can — but I'm not going to be able to guarantee that we'll be able to support anything like this for this year's version of the build server. If you have any suggested changes to the actual templates or other content of the proceedings server, please make a separate pull request. Then, we can work on that there and get it incorporated into the build system more generally. 

cc @katyhuff @dalippa @deniederhut

gpoore:
I've modified and improved the custom LaTeX so that it works with the version of TeX Live on the build server. I'll look into converting the custom LaTeX into pull requests on the proceedings templates. I maintain the `minted` package for LaTeX, so I have a lot of experience with trying to get the single quotation mark to appear correctly in code.

mpacer:
@gpoore I didn't realise you were the minted maintainer. Your work is awesome; happy that you'll be contributing :).

aweinstein:
Hi: I've been asked to review this paper.

The paper presents BespON, a new configuration file format, and its corresponding Python implementation. The paper reviews current configuration file formats (INI, YAML, JASON, etc), describing their key features and drawbacks. It then presents BespON, showing its main features and how this new format overcome the drawbacks of the existing configuration formats.

The paper is very well written and clear. I have only two suggestions to improve this work:

1. Explain what round-tripping is, and why it is important, particularly in the context of scientific computing. The only explanation I see is in the introductions, where it says "One particularly nice feature of configobj is round-trip support. Data can be loaded, modified, and saved, while preserving the order of values and retaining comments." I think the term is not well know by the scientific computing community (at least it was new to me! :), and it will help to have a paragraph explaining this concept. 

2. Show a complete BespON config file. All the examples are given as Python statements (e.g., `bespon.loads("[none, true, false]")`). It would be nice to show, at the end of the description of BespON, a complete example showing all the features.

mpacer:
@gpoore your paper currently doesn't build on the server, it does seem to work fine on my local machine — I'm not sure what the issue is since it's not related to the spacing issue that should now be fixed. Do you have any insight based on the error that's coming up?

gpoore:
@mpacer It works now, after a new commit. Since the build server was updated with the new `scipy.sty` from #294, I needed to remove the custom LaTeX in the paper.

mpacer:
Excellent. Also, the fontsize/spacing should now match between your local machine and server — does that accord with your experience?

gpoore:
@mpacer Yes, the server and my local machine are now producing the same appearance, including page break locations. Previously, the server produced PDFs that were about 1/2 page longer than local.

gpoore:
@aweinstein Thanks for the review.

1. I will look into adding a more explicit definition and explanation of round-tripping.

2. All of the examples were loading data from strings with `bespon.loads()`. If the string text from any of the examples was saved in a file instead, say `data.bespon`, then it could be loaded like this:
```python
with open('data.bespon') as f:
    data = bespon.load(f)
```
I could convert some of the examples to this form if you think that would improve things. Or are you thinking more about a single example file that uses most of the features, like what is at the top of the page at https://bespon.org/?

aweinstein:
@gpoore I'm thinking showing something like the example at the top of https://bespon.org/. That very same example would be enough. 

The only problem is that you will probably exceed the 8 page limit, but in my opinion a hard limit in the number of pages doesn't make much sense. Perhaps the editors (@mpacer, @katyhuff) would let you have more pages?

gpoore:
@aweinstein The formatting on the build server has changed recently, probably since you looked at the paper, so I now have a little over 1/2 page free to keep within 8 pages. That should be enough for a paragraph about round-tripping and a longer example, without cutting anything out.

gpoore:
@aweinstein I've added a paragraph in the introduction about round-tripping, and added a longer example at the end. The build server doesn't seem to be working, though...I haven't been able to get it to create an updated PDF. Perhaps @mpacer can check on it.

deniederhut:
Hi @gpoore ! I've been asked to provide another review for your submission. 

Let me start off by saying 👏 👏 👏 ! This looks like a useful tool, and the write-up is both clear and informative. I have two comments about the tool specifically, and one about the paper.

1. Am I understanding correctly that comments are read in as strings? I can see how this would be useful, but I can also think of times where I would not want this to happen. In my perfect world, comments would be contained in a separate datastructure, but I understand this could be tricky to implement. Maybe the easiest solution would be an "ignore_comments" flag at read time?
2. Are you planning on providing native support for datetime objects? On a related note, you mention user-defined data types are supported. How do these work?
3. I really like that you mention the motivations behind formats like json and yaml, but a casual reader might not see the conceptual difference between configuration data and serialized data. Could I ask you to include a sentence or two about this?

gpoore:
@deniederhut Thanks for the comments.

1. By default, all data is loaded as standard Python types (not subclasses), and all comments are discarded. Accessing comments as strings is possible (though not yet documented) by using the load functions that return an AST rather than plain data.

2. There will eventually be support for date, time, and datetime objects from the `datetime` module. These will probably be tagged (something like `(datetime)> "2017-06-26T00:53:54+00:00"`) rather than unquoted literals, so that everything is treated uniformly with minimal potential for confusion with integers. I haven't added these yet because I believe there are a few Python 2/3 differences, and I also want to make sure I deal with naive/aware (local/offset) properly. Adding user-defined types will be as simple as creating a custom decoder instance and initializing with things like `DataType(name='set', basetype='list', mutable=True, parser=set)`. This system is already used for built-in types; I just need to add some keyword args and associated logic in the decoder to enable user-defined types.

3. That's a good point about the terminology. I'll add a few sentences soon.

mpacer:
@gpoore I checked on it and it should be up and running again.

gpoore:
@deniederhut I've added some clarifications under the JSON and YAML summaries. Let me know if that provides what you're looking for.

deniederhut:
The new wording looks great to me! And 👍 for taking the time to carefully consider the 2/3 interoperability issue. 

If I could ask just for one more thing before I give the green light - could you mention that user-defined types are a planned future feature? It might be in the paper already and I just missed id.

aweinstein:
@gpoore @mpacer @katyhuff I recommend accepting the paper.

gpoore:
@deniederhut I mention user-defined types as a future feature in the conclusion (page 8, 3rd paragraph under Conclusion).

>Eventually, it will also be possible to enable optional user-defined data types with the tag syntax.

Let me know if you think more is needed.

deniederhut:
@gpoore I must have missed it - LGTM!

PR #286, opened 2017-05-30
Paper: SciSheets

joseph-hellerstein:
This is the submission of the paper "SciSheets: Providing the Power of Programming with the Simplicity of Spreadsheets"


JeanBilheux:
I'm going through the document but have already a few comments. 

I really like the way it's articulated and it's very clear to follow. It clearly shows what is going to talk about, the various parts. I like how it demo the needs for such a tool using real case examples. I'm a little confused though about the presentation of the 2 examples...I think some of the text from example 2 got moved accidentally into example 1. Example 1 were talking about how to calculate formula related to enzyme that suddenly it's talking about student grades and comes back to the enzyme case.

JeanBilheux:
"""We begin with a discussion of the SciSheets user interface in Section 3.1. Then, Sections 3.2, 3.3, and 3.4 in turn present: formula scripts (which addresses expressivity), function export (which addresses reuse and performance), and subtables (which addresses complex data)."""

I would say instead 

"""
We begin with a discussion of the SciSheets user interface in Section 3.1. Then, Sections 3.2, 3.3, and 3.4 present formula scripts (which addresses expressivity), function export (which addresses reuse and performance), and subtables (which addresses complex data) respectively.
"""

JeanBilheux:
"""SciSheets users interact directly with scisheet elements instead of primarily with a menu"""

I'm a little bit confused about what **primarily** really means. Maybe single cell?

JeanBilheux:
I got a little bit confused by part 3.3 but this should be solved if I was running the library at the same time and going through the example step by step myself. 

Very nice detail (almost a tutorial) of the library.

joseph-hellerstein:
Thanks so much for the feedback. My replies to the comments follow:
1. I reread Section 2 (where the examples are) a couple of times and didn't see the misplaced text that you referenced. Can you give me more specifics?
2. I like your wording for the beginning of Section 3 and have incorporated it. Thanks.
3. Thanks for picking up on this. I think that the following wording is more accurate and clear. "SciSheets users interact directly with the scisheet element appropriate for the desired action."

JeanBilheux:
screen_shot_2017-06-20_at_4_31_00_pm

This is what I was talking about in my part 1., but maybe this is more related to a problem of format due to RST!

bmcfee:
I just gave this one an initial read, and it looks to be in great shape already!  Quite clear and easy to follow.

A couple of comments / thoughts I had while reading:

- In section 1, there's an argument that spreadsheets are popular because they tightly couple computation and data.  If I understand correctly, the design of scisheets (necessarily) separates these to facilitate code reuse (even if they are stored in the same serialized object).  The part that's unclear to me is how this separation is exposed to users.  This isn't necessarily a huge deal, but it might be helpful to include some discussion here.

- In 3.1, it states that column variables are numpy arrays, which doesn't completely square with the description of subtables in section 1.  Is it more accurate to say that columns are either numpy arrays or subtables?

- In 3.2, I found myself wondering how exactly formula computations work, in a couple of ways:
  - Are formula scripts assumed to be vectorized?  Do users have to be familiar with numpy to use this effectively? Or can an output be any (finite) iterable, e.g., `X_SQ = [x**2 for x in X]`?
  - Is it possible to (accidentally) share or leak state across formula evaluations?

- The screencap figures are good, but I think it's not totally obvious what the interaction should be for specifying a formula.  In general, do you specify formulas on the input column or the output column?  What about many-to-many mappings?

- Section 5.1: is it possible to expose subtable contents outside the subtable via scoping?  E.g., can I probe `Biology.GPA` from a different subtable? Or from a column formula elsewhere in its containing scisheet? It's fine if not, but it seems like the sort of thing a user might want to do, and I didn't notice it discussed in the draft.

One typo, section 5.2, `course` should be `coarse`.

joseph-hellerstein:
Regarding Brian's comments:

1. Added a paragraph after in 3.3 (the second paragraph).

2. Section 3.1 (2nd par). Clarified what is being referenced as a column in the discussion of numpy Arrays.

3. See text at the end of 3.2

4. It is the users choice where to place a formula. Name scopes are global.

5. Subtable name scopes is being implemented. And yes, the plan is to be able to reference subtables/columns in the manner that you describe.

My latest commit contains the changes above.

katyhuff:
Thank you for your reviews, @JeanBilheux and @bmcfee ! The authors appear to have made changes according to your comments. In your opinion, is this paper now ready for inclusion in the SciPy proceedings?

joseph-hellerstein:
I think there's a bit of reformatting that's still needed on the paper.
Other than that, I don't have other changes planned.

bmcfee:
I just gave it another read, and I think my comments have been sufficiently addressed.  :shipit:

katyhuff:
Excellent! @joseph-hellerstein The camera-ready deadline was yesterday, so please commit any final commits as immediately as possible. There is no guarantee that they will make it into the final proceedings if new commits are made after today. We are trying to get a complete draft of the proceedings out before SciPy happens. Thank you!

joseph-hellerstein:
The issue is that there's some formatting the causes of paper to go beyond
eight pages. I'm away right now but I'll be back and able to make a commit
before midnight tonight.

On Jul 9, 2017 2:45 PM, "Katy Huff" <notifications@github.com> wrote:

Excellent! @joseph-hellerstein <https://github.com/joseph-hellerstein> The
camera-ready deadline was yesterday, so please commit any final commits as
immediately as possible. There is no guarantee that they will make it into
the final proceedings if new commits are made after today. We are trying to
get a complete draft of the proceedings out before SciPy happens. Thank you!

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/scipy-conference/scipy_proceedings/pull/286#issuecomment-313950013>,
or mute the thread

joseph-hellerstein:
I made edits so that the paper fits in 8 pages.

PR #287, opened 2017-05-31
improve error handling for docutils that aren't  0.12

mpacer:
Going to self merge this.


mpacer:
This needs to be fixed to let the build server work

PR #290, opened 2017-06-02
Paper: Parallel Analysis in MDAnalysis using the Dask Parallel Computing Library

mkhoshle:


gmarkomanolis:
Hello,

Some comments for typos/corrections

"We found that the the underlying storage system",  remove the double "the".
You could replace the word "bottle neck" with "bottleneck"
"Scaling was tested on single node" -> "Scaling was tested on a single node"
"examine effect of I/O" -> "examine the effect of I/O"

"These file systems guarantees sequential " -> "These file systems guarantee sequential " 

"XTC file format express reasonably " -> "XTC file format expresses reasonably "

"Therefore, it is expected to achieve speed up, across parallelism up to 50 cores However," -> "Therefore, it is expected to achieve speed up, across parallelism up to 50 cores. However,"

"overal performance" -> "overall performance"

"In fact, the time the jobs take to run are affected by the other jobs" -> "In fact, the time the jobs take to run is affected by the other jobs"

"and as aresult the overal speed-up" -> "and as a result the overall speed-up"

"in computations is to submit many more tasks than the numer of cores" -> "in computations is to submit much more tasks than the number of cores"

"scheduler pluging is used" -> "scheduler plugin is used"

"and to also validate our observationis from Dask web-interface" -> "and to also validate our observations from Dask web-interface"

"available cores is still able " -> "available cores are still able "

"profiles were later use to analyze" -> "profiles were later used to analyze"

"in the present benchmark using Dask parallel libray" -> "in the present benchmark using Dask parallel library"

gmarkomanolis:
Also, you mention "Figure fig:pattern-formats shows the I/O pattern compared between different file formats."  I suppose you mean Figure 1. Fix also for the rest of the reference to figures. 

You don't mention which version of netCDF you are using, but there is compression with netCDF-4 (http://www.unidata.ucar.edu/blogs/developer/entry/netcdf_compression), maybe in the future, you could be interested in testing it.

mrocklin:
### General feedback

This paper shows the experience of parallelizing a standard parallel
programming computation with Dask.  It goes through the experience of
identifying and optimizing a variety of issues that arise in parallel and
distributed computing in common practice.  These issues (network I/O,
stragglers, communication/scheduler overhead) are important and rarely
discussed, so it is nice to see them show up in a published work.

This Paper appears to be above the 8 page limit

This paper does a good job of being reproducible.  It includes code, data, and analysis

### Nit-picky feedback

After a first run-through of the document here are some general nit-picky
remarks.  I apologize for the late review.

I have not seen NetCDF abbreviated to NCDF before.  I recommend adding back in
the "et" in "net" to stick to convention.

Figures 3/4

-   Text is somewhat small.
-   I'm uncertain what the axes mean. I interpret "IO Per frame" as amount of
    time spend to read all blocks, is this correct?  Is the "number of blocks"
    also the same as the number of processes used?  If so then why are these
    tied together?  If my understanding is correct then I would expect speedups
    to visually manifest as a downward slope.  If convenient it might be useful
    to switch axes to show bandwidth (MB/s) on the Y-axis instead.

If number of blocks is the same as number of cpus then I recommend changing
axis labels to cpus instead of blocks.  I believe that this will be more
readily understood by readers.

When using local HDD/SSD in a distributed setting how do you assign files to workers?

If all DCD is doing is ripping bytes from disk without any processing then I
wouldn't expect much dependence on parallelism at all.  It should be entirely
determined by the I/O bandwidth to the CPUs.

The y-axes in figure five should probably be the same.  As is one of them has a
maximum of 10e4 while the others have maxima of 10e3.  This is somewhat
misleading at first glance.

Many Figures for 300x/600x seem to have identical captions.  it might be wise
instead to include these pairs of figures into a single large figure with a 2x3
grid of images and a single caption.  This would also allow you to more easily
highlight differences between the 300x and 600x runs.

The similarity of times measured by your infrastructure and the dask web
interface does not seem like an important point to focus on.  I might skip it
and exclude it from table 1.

Figures labels might be better if they avoided internal variable names.  For
example `Total time t_{total} (s)` might be replacec with `Total time (s)`

Much of this study seems to be about strong scaling, seeing how fast we can run
the same computation on more resources.  As data sizes increase presuambly weak
scaling is also useful.  Seeing how much independent data we can churn through
on larger systems.  Are MDAnalysis workloads amenable to this?  Or do you
process through one frame before another which stops embarrassingly parallel
work across time?


### Examining Scheduler Throughput

Fig 15:  y-axis should be tasks per second?

I am somewhat surprised to see the distributed scheduler manage 40,000 tasks
per second.

I recommend not capitalizing multiprocessing, multithreading,
scheduler, cluster, or distributed.

It might be useful to have a link to your dask scheduler plugin from your
paper.

oleksandr-pavlyk:
I am running out of time to provide a quality review, sorry. However while reading the article I missed having  a reference explaining a concept of "optimal structural superposition of atoms".

orbeckst:
@mrocklin and @oleksandr-pavlyk – many thanks for your very constructive reviews, we will attend to the points raised, especially the difficulties with the figures. We'll also try to reduce "simulation community jargon" such as "optimal structural superposition of atoms" – that was an attempt at brevity in the Methods section that only makes sense to someone who analyzes protein structures for a living...

mkhoshle:
I express my sincerest thanks to the reviewers for their valuable comments that helped us to improve quality of the manuscript which has now been edited to address their concerns. Other comments provided by the reviewers have been addressed in detail as described in the following.

Yours Sincerely,
Mahzad Khoshlessan
On behalf of all the authors

Reviewer's recommendations:

- [ ] This Paper appears to be above the 8 page limit 
       **We have tried to shorten the paper as much as possible. It is about 9 pages now (Including the references)**

- [ ] I have not seen NetCDF abbreviated to NCDF before. I recommend adding back in
the "et" in "net" to stick to convention. **NCDF is used for consistency everywhere in the text** 

- [ ] Figures 3/4 
  - [ ] Text is somewhat small. **Font size is increased for all figures**
  - [ ] I'm uncertain what the axes mean. I interpret "IO Per frame" as amount of time spend to read all blocks, is this correct? Is the "number of blocks" also the same as the number of processes used? If so then why are these tied together? If my understanding is correct then I would expect speedups to visually manifest as a downward slope. If convenient it might be useful to switch axes to show bandwidth (MB/s) on the Y-axis instead. **IO time per frame is the amount of time spent to load one frame into memory. More details are given in Result and discussion section-Paragraph 1. Since, MDAnalysis only loads a single frame into memory at any time, we are interested in IO time per frame instead of throughput. Also, "number of blocks" is the same as the number of processes in some tests. When we increase number of CPU cores, IO time per frame should not change because the frame size will be the same. For more detail refer to Effect of File Format on I/O Performance section paragraph 2.** 
 
  - [ ] If number of blocks is the same as number of cpus then I recommend changing axis labels to cpus instead of blocks. I believe that this will be more readily understood by readers. **Number of CPU cores is used wherever we meant number of processes. In text, M refers to number of blocks and N refers to number of CPU cores (processes)**

- [ ] When using local HDD/SSD in a distributed setting how do you assign files to workers? **We did not use HDD and SSD together, If this is what you meant. We tested our benchmark on HDD and SSD separately and for that the scheduler node and the worker nodes are set up by user and then the Dask distributed network will be used by client** 

- [ ]  If all DCD is doing is ripping bytes from disk without any processing then I wouldn't expect much dependence on parallelism at all. It should be entirely determined by the I/O bandwidth to the CPUs. **Addressed in Effect of File Format on I/O Performance section paragraph 2 and 3**

- [ ] The y-axes in figure 5 should probably be the same. As is one of them has amaximum of 10e4 while the others have maxima of 10e3. This is somewhat misleading at first glance. **Corrected in all figures**

- [ ] Many Figures for 300x/600x seem to have identical captions. it might be wise instead to include these pairs of figures into a single large figure with a 2x3 grid of images and a single caption. This would also allow you to more easily highlight differences between the 300x and 600x runs. **Addressed in Figure 2, 3, 5, 6, 7 and 16** 

- [ ] The similarity of times measured by your infrastructure and the dask web interface does not seem like an important point to focus on. I might skip it and exclude it from table 1. **Addressed, refer to figure 7 and last paragraph in Strong Scaling Analysis for Different File Formats section for more detail**

- [ ] Figures labels might be better if they avoided internal variable names. For example Total time t_{total} (s) might be replacec with Total time (s) **Addressed in all figures** 

- [ ] Much of this study seems to be about strong scaling, seeing how fast we can run the same computation on more resources. As data sizes increase presuambly weak scaling is also useful. Seeing how much independent data we can churn through on larger systems. Are MDAnalysis workloads amenable to this? Or do you process through one frame before another which stops embarrassingly parallel work across time? **Addressed in Results and Discussion section, first paragraph** 

## Examining Scheduler Throughput

- [ ] Fig 15: y-axis should be tasks per second? **Corrected in Figure 10 in the final draft** 

- [ ] I am somewhat surprised to see the distributed scheduler manage 40,000 tasks per second. **The results are double checked, for detail, refer to figure 10 and Scheduler Throughput section**
 
- [ ] I recommend not capitalizing multiprocessing, multithreading, scheduler, cluster, or distributed. **Corrected**

- [ ] It might be useful to have a link to your dask scheduler plugin from your paper. **Link to the main code provided in: (Effect of Over-Subscribing section paragraph 3)**

- [ ] I missed having a reference explaining a concept of "optimal structural superposition of atoms". **Reference given in text (Reference [MM14])**

mkhoshle:
All the revisions are completed and point by point answers to the reviewers' comments are in the above comment!

katyhuff:
@mrocklin, @oleksandr-pavlyk, and @gmarkomanolis : Thank you for your reviews! The authors have made changes according to your comments. In your opinion, is this paper ready to be merged?

PR #294, opened 2017-06-07
Improve `scipy.sty` (no curly quotes in code, more math symbols)

gpoore:
This improves `scipy.sty` by making sure all single quotation marks in code are not curly, and thus both appear and also copy correctly. It also adds the `amssymb` package to provide some additional math symbols. I added some custom LaTeX in my paper submission (#285) to accomplish this; if `scipy.sty` is improved, then that custom LaTeX would no longer be necessary.

### DIFF
--- a/publisher/_static/scipy.sty
+++ b/publisher/_static/scipy.sty
@@ -71,14 +71,17 @@
 \usepackage{fancyvrb}
 
 % Make single quotes non-curly in inline literal text (\texttt)
-% and in code-block directives (which use Pygments; patch Pygments macro)
+% and in code-block directives (which use Pygments; patch Pygments macro).
+% Also fix backtick in inline literal text.
 \AtBeginDocument{%
   \let\scipyoriginaltexttt\texttt
   \def\texttt#1{%
     \begingroup
     \scantokens{%
 	  \catcode`\'=\active
+	  \catcode`\`=\active
 	  \let'\textquotesingle
+	  \let`\textasciigrave
 	  \scipyoriginaltexttt{#1}\noexpand}%
     \endgroup}%
   \def\PYZsq{\textquotesingle}%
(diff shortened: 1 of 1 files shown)


gpoore:
The last commit also adds a fix for backticks in inline code. They were coming through as left single curly quotation marks.

katyhuff:
@mpacer, is it possible for you to test how this impacts existing PRs for this year's submissions?

mpacer:
@katyhuff Not super easily — we could merge it and then revert it if it causes a problem.

@gpoore what is the textcomp package adding? I can tell it's doing some postscript font encoding magic (re: TS1 encodings?), but beyond that I'm having difficulty parsing its effects. 

My hope was to move to xelatex by next year, this seems to work against that in the long term (since from what I gather what we want there are TU compatible fonts). 

I'm ok merging in theory, but I want to make sure I understand the downsides before doing so.

gpoore:
`textcomp` is used to define the backtick and single quotation mark characters if the font isn't Computer Modern or the font encoding isn't OT1 (both are true in the proceedings template, and OT1 is rarely used). It is loaded by `upquote` as necessary, but I loaded it explicitly since it is typically needed. `textcomp` defines all of its symbols in terms of TS1 encoding as a way of mapping code points to font glyphs, but doesn't affect the font encoding of the document otherwise; it's basically just one way to look up the right characters for the PDF.

All of the backtick/quotation mark code is primarily for pdflatex. With xelatex, the backticks and single quotation marks are typically fine by default (so `textcomp` and `upquote` effectively do nothing). If the code font is set like
```latex
\setmonofont{<font>}
```
with xelatex, the backtick and single quotation mark are usually fine by default. However, if the code font is set with
```latex
\setmonofont{<font>}[Ligatures=TeX]
```
then `upquote` and `textcomp` would be needed with xelatex, as well as additional code to keep the *double quotation marks* from being curly.

mpacer:
thank you for that super clear explanation. Didn't run comprehensive tests, but this looks good and seems to do what it says. I'm merging & restarting the server thanks @gpoore!

PR #300, opened 2017-07-01
New : automated Crossref doi metadata generation

deniederhut:
This PR makes slight modifications to the config files and build process, and adds two new modules, the combined effect of which will generate the required Crossref metadata submission file. This happens automatically during the build process itself, and places the file ("doi_batch.xml") in the build directory. Three things:

1. We are not currently attempting to include metadata from the references inside of papers, as I could not find an obvious place inside of `writer/__init__.py` where these are handled. This should be functionality we consider adding for next year.
2. We are currently listing "XXX.org" as the url for the resource location, because I don't yet know the rule for how we are generating the endpoints for individual papers. 
3. It turns out that parsing human names is tricky. We are currently using `nameparser` for this, because it is the first thing that shows up when you google "python library for parsing names". If there is a better solution, I would be happy to modify the PR.

The output of `XrefMeta.make_metadata` has been validated by Crossref, but only for `00_vanderwalt` and only after building on my local machine. I would greatly appreciate test runs on zibi with the actual conference papers.


deniederhut:
In case this facilitates comparison, the output from my local install looks like this:

```
<doi_batch xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="4.4.0" xmlns="http://www.crossref.org/schema/4.4.0" xsi:schemaLocation="http://www.crossref.org/schema/4.4.0 http://www.crossref.org/schemas/crossref4.4.0.xsd"><head><doi_batch_id>Majora.d13c5b1-15cfc43ae4f</doi_batch_id><timestamp>1498861112.0</timestamp><depositor><depositor_name>Dillon Niederhut</depositor_name><email_address>dniederhut@enthought.com</email_address></depositor><registrant>Crossref</registrant></head><body><conference><event_metadata><conference_name>Python in Science Conference</conference_name><conference_acronym>SciPy</conference_acronym><conference_number>15th</conference_number><conference_location>Austin, Texas</conference_location><conference_date>June 28 - July 3 2017</conference_date></event_metadata><proceedings_metadata><proceedings_title>Proceedings of the 15th Python in Science Conference</proceedings_title><proceedings_subject>Scientific Computing with Python</proceedings_subject><publisher><publisher_name>SciPy</publisher_name></publisher><publication_date><year>2017</year></publication_date><noisbn reason="simple_series"/><doi_data><doi>10.25080/Majora-d13c5b1-000</doi><resource>http://XXX.org</resource></doi_data></proceedings_metadata><conference_paper><contributors><person_name contributor_role="author" sequence="first"><given_name>Gaius</given_name><surname>Caesar</surname></person_name><person_name contributor_role="author" sequence="additional"><given_name>Mark</given_name><surname>Anthony</surname></person_name><person_name contributor_role="author" sequence="additional"><given_name>Jarrod</given_name><surname>Millman</surname></person_name><person_name contributor_role="author" sequence="additional"><given_name>Brutus</given_name><surname>MISSING</surname></person_name></contributors><titles><title>A Numerical Perspective to Terraforming a Desert</title></titles><publication_date media_type="print"><year>2017</year></publication_date><pages><first_page>1</first_page><last_page>3</last_page></pages><doi_data><doi>10.25080/Majora-d13c5b1-001</doi><resource>http://XXX.org</resource></doi_data></conference_paper></conference></body></doi_batch>
```

mpacer:
In order to make this work reproducibly on zibi will probably require altering `procbuild` as well.

I feel like we should have some kind of structured object that stores the input to the DOI generating function. Right now, it's generated at runtime and is dependent on the time of the run… information that is lost when the python process is no longer in memory. That will make it hard to know if any future changes in the code might alter the way that we generate DOIs. Not only would this mean that we could _test_ for this kind of change in our conventions, it would allow us to state precisely how (if we do intend to change that convention) how that change would alter the DOIs created under the previous convention. 

Could you add documentation to the xreftools module? It's pretty complicated even if in many ways it's just a series of assignments. 

I think the real check regarding name parsing would be to check with all of the authors that the resulting parsing of their names is more or less as they would like it to be. I don't know of anything better, but I could imagine that there will be complications. It might be worthwhile trying to have at least a couple of test cases and a test. Basically, no matter what we choose to do, it's likely that something will break at some point for someone. That will result in changes to the code and we we want to know if those changes break known cases that already worked.

deniederhut:
I can take a look at procbuild.

The reason that the DOI generator keeps producing unique values is that the DOIs are required to be unique -- much like a UUID. We can store state if you like, but then we incur communication costs / update anomalies / wrong DOIs if we want to generate metadata on more than on computer.

The documentation for the XML schema is here - http://data.crossref.org/reports/help/schema_doc/4.4.0/4.4.0.html.

mpacer:
Ok, per our conversation today, we want to build the DOIs, so merging this is a priority. I'm going to merge this and @deniederhut agreed to work on documentation and testing in the near (but not extremely near) future.

PR #303, opened 2017-07-07
Fix : updates proceedings metadata for 2017

deniederhut:
This includes dates, editors, and organizers, but does not include the scholarship recipients -- I am waiting to hear back about a list of these.

### DIFF
--- a/publisher/_templates/students.tex.tmpl
+++ b/publisher/_templates/students.tex.tmpl
@@ -6,15 +6,28 @@
 
 \begin{document}
 
-\section*{Sponsored Students}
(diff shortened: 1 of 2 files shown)


deniederhut:
This now includes the sponsored students - I separated the ones who have their scholarship sponsored by Jump Trading / NumFOCUS. I'd like input from @katyhuff on whether this is the proper way to handle it -- it comes out looking like this:

screen shot 2017-07-07 at 4 56 37 pm

I've also edited the template file so that it no longer requires geographic location, which we don't have anyway.

mpacer:
Looks good.

PR #305, opened 2017-07-07
Fix : allows more than two editors on copyright page

deniederhut:
This PR corrects two keyword arguments in the template file in `_templates/copyright.tex.tmpl` to match the metadata we currently store in `scipy_proc.json`.


mpacer:
Looks good.

PR #307, opened 2017-07-11
make proceedings-html was not python2/3 compat, this fixes that

mpacer:
Now `make proceedings-html` will work in both python2 & 3. 

Not sure why it's duplicating the add list of proceedings reviewers commit.


mpacer:
merging so that anyone can actually use the build system to make the proceedings.

ivanov:
Yay! 👍  Glad this got figured out!

PR #322, opened 2017-07-18
Better nav

mpacer:
This will avoid many of the issues that currently plague the proceedings website. It should overwrite whatever file is currently present in the directory in which it's unzipped (meaning we can update it in our repo rather than relying on the current value of https://conference.scipy.org/proceedings/proc_links.js. 

This has been already been rebased on #320.

### DIFF
--- a/publisher/_static/proc_links.js
+++ b/publisher/_static/proc_links.js
@@ -1,5 +1,5 @@
 function proc_versions() {
-   var versions = range(2017, 2008)
+   var versions = range('2017', '2008')
    var proc_url = 'http://conference.scipy.org/proceedings/scipy';
    document.write('<ul id="navibar">');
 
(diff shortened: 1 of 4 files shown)


mpacer:
Also this fixes the links to the conference level .bib file, which I honestly think we should reconsider anyway.

publisher/_static/proc_links.js
```diff
@@ -1,5 +1,5 @@
 function proc_versions() {
-   var versions = ['2011', '2010', '2009', '2008'];
+   var versions = range(2017, 2008)
```
dalippa: These are numbers, do they need to be strings? Why not have it be (start, end). This is technically in reverse order, or end, start.

dalippa (changes requested):

publisher/_static/proc_links.js
mpacer: I think the ordering was to be consistent with how it was done before. 

The parameter names for the range function are indicating which direction it should go.

This way you don't need to say `range(2008, 2017, -1)` but just `range(2017, 2008)`.

But, I agree it's not super pythonic API.

As for whether they need to be strings, I don't see why they would, but if that's needed I can modify the internal reasoning of range to track if they are strings, convert them to ints & then cast them back to strings right before they are returned. That way the type of the input will match the type of the output.

dalippa: OK. Just thought I would ask.

mpacer: totally reasonable request! Do you want me to implement the string thing?

mpacer:
Ugh… messed up the logic at some point during the rebase. I'll fix this, soon but right now its broken.

stefanv:
@mpacer Want to get this back into shape?

mpacer:
@stefanv I pushed it and it's now back into shape. 

@dalippa now it allows passing in strings of numbers and I gave it a docstring to explain what it was doing.

publisher/_static/proc_links.js
```diff
@@ -11,3 +11,50 @@ function proc_versions() {
...
    document.write('</ul>');
 }
+
+// range(start, end, step) 
+// provides a range of integers or strings, depending on the type of input provided.
+// If you input strings they will be converted to integers for the iteration logic to work.
+// start: Int or string, first element to include in output
+// end: Int or string, last element to include in output
+// step: Int, amount and direction the array will be stepped through
+var range = function(start, end, step) {
```
stefanv: This range function feels like overkill for our needs.

publisher/_templates/header.html.tmpl
```diff
@@ -3,7 +3,7 @@
     <meta content="text/html; charset=utf-8" http-equiv="content-type">
     <title>{{html_quote(proceedings['title']['short']) | html}}</title>
     <link rel="stylesheet" href="scipy-proc.css" type="text/css"/>
-    <script type="text/javascript" src="http://conference.scipy.org/proceedings/proc_links.js"></script>
+    <script type="text/javascript" src="../proc_links.js"></script>
```
stefanv: This seems incorrect---we want to refer to the future version.

mpacer: That's not referring to the future version right now…it's referring to the static version on the SciPy website servers. In any case, I'm working on updating this for other reasons.

publisher/_static/proc_links.js
mpacer: Really? I'd prefer to make it so that we don't need to remember to add a sequence of strings. 

This is not going to significantly slow down the site or increase the site page. It seems like "overkill" is worth it if it avoids accidental oversights that last more than 5–6 years.

mpacer: It was less overkill prior to including the string catching stuff. I can remove that complexity if that's what you're suggesting — I was trying to meet @dalippa's previous review.

stefanv:
This file still needs to be updated, unfortunately. We'd probably be better 
off reading the current year from the json config upon generation, if you 
want to include the whole range here. That wasn't an option earlier, 
because 2012 was, indeed, not released for many years.

mpacer:
@stefanv @dalippa I'm going close this PR as it doesn't sound like these changes are desired.

stefanv:
@mpacer I think the PR was on the right track; maybe simplify the range argument and find a way to read the date either from the configuration file, or use the current date?  But it's pushing in the right direction.

stefanv:
> I didn't realise that 2012 hadn't been released for many years… I
> assumed that was an oversight. What had happened?

We had a little slip up with reviewing, which we fixed after the fact
(which caused some more excitement).  The table of contents should give
you some more insight: http://conference.scipy.org/proceedings/scipy2012/

mpacer:
Oh ok… I can reopen it and work on it, I just got the impression that this kind of change wasn't desired since if we aren't pointing to the local copy of the js file then there's literally no way for us to see its effects on the html that we're building.

stefanv:
You can still test it locally (make the path change), but not commit
that part of the patch.

mpacer:
Sure but that's entirely what all of my changes to the system are doing is eliminating the need to dive into files to customise them for build steps vs. release steps. It requires a lot more knowledge to be able to do that than if it's easily configurable (see my #337 PR and how we can build the different proceedings versions). If that's what you'd prefer I do… I can write something like this for that…

mpacer:
@dalippa mentioned in the proceedings call today that he would not be able to code much (or review code much) so I'm going to suggest that we don't block on his changing his approval of this PR to merge it (though I still need to meet your review @stefanv).

cbcunc:
Old publication process deprecated.

PR #323, opened 2017-07-24
Typos and minor style improvements

TheChymera:
As mentioned on Slack @stefanv

### DIFF
--- a/papers/horea-ioan_ioanas/paper.rst
+++ b/papers/horea-ioan_ioanas/paper.rst
@@ -14,7 +14,7 @@
 :bibliography: mybib
 
 
-LabbookDB - A Wet-Work-Tracking Database Application Framework
+LabbookDB — A Wet-Work-Tracking Database Application Framework
 ==============================================================
 
 .. class:: abstract
--- a/papers/horea-ioan_ioanas/paper.rst
+++ b/papers/horea-ioan_ioanas/paper.rst
@@ -103,7 +103,7 @@ Output Design
 ~~~~~~~~~~~~~
 
 Outputs include simple human-readable command line reports and spreadsheets, ``.pdf`` protocols, introspective graphs, and dataframes.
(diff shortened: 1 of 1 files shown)


stefanv:
For attention: @mpacer

mpacer:
I'm concerned about actually having a em-dash in the title… not because of anything having to do with our build system per se, but because when I tried to do this in my own tests it produced errors when interpreting the bibTeX file: see this [SO issue](https://tex.stackexchange.com/questions/167196/error-unicode-char-u8-when-importing-bibliography). 

I am considering a shift to bibLaTeX since that supposedly should be able to support that… but I'm not going to lie, the last time I tried to do something with bibLaTeX it introduced a lot of pain into my workflow and I quickly regretted it. So if we're going to do this, it will require some extra time (especially since I cannot spend as much time in the next few weeks as I did to get us to the point of having working proceedings before the conference started). 

@dalippa @katyhuff @deniederhut If you have any thoughts on this I'd love to hear them. It might make sense to schedule a meeting soon so we can address this as well as some of the other issues.

deniederhut:
@mpacer I will happily defer to your verdict on biblatex. In the more general sense of your question, I am in favor of reducing the maintenance burden for future committee chairs, even at the expense of the noble em-dash.

TheChymera:
Certainly I wouldn't want to add to the overhead and delay the final publication just to improve the dash in my title (it looks like it's just me having this issue). To be honest - though I quite like typography - I always used a hyphen to substitute all dashes up until Stefan brought it up. It just seemed to me that the semantic meaning of the em-dash was unambiguous given the flanking spaces - and also the context.

The bibliography management issue also exceeds our build system, Most people still use bibtex, and I don't want my em-dash to break the build systems of people who may want to cite :)

mpacer:
This all seems good… and it is unfortuante about bibTeX and utf-8.

PR #325, opened 2017-08-10
update video links for all authors to SciPy2017 videos, remove links to non SciPy2017 videos

mpacer:
This is one of the last steps we need to complete before we can release the papers. 

I did this by hand, so there could be errors. 

But, I think all of the remaining papers that don't have video links (after this PR is merged) are papers associated with poster presentations (and therefore have no video).

### DIFF
--- a/papers/horea-ioan_ioanas/paper.rst
+++ b/papers/horea-ioan_ioanas/paper.rst
@@ -12,6 +12,7 @@
 :institution: Institute for Biomedical Engineering, ETH and University of Zurich
 
 :bibliography: mybib
+:video: https://youtu.be/yDBu_wSyw-g
 
 
 LabbookDB: A Wet-Work-Tracking Database Application Framework
--- a/papers/ingo_heimbach/paper.rst
+++ b/papers/ingo_heimbach/paper.rst
@@ -75,8 +75,6 @@
 :email: r.jones@fz-juelich.de
 :institution: PGI-1 and PGI/JCNS-TA, Forschungszentrum Jülich, D-52425 Jülich, Germany
 
-:video: https://pgi-jcns.fz-juelich.de/pub/media/pymoldyn_algorithms.mp4
-
 
 #################################################################################################
 pyMolDyn: Identification, structure, and properties of cavities in condensed matter and molecules
--- a/papers/jaime_arias/paper.rst
+++ b/papers/jaime_arias/paper.rst
@@ -37,6 +37,7 @@
 :equal-contributor:
 
 :bibliography: biblio
+:video: https://youtu.be/Usyvds1o4lQ
 
 --------------------------------------------------------------------------------------------------------------------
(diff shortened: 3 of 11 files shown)


mpacer:
This now adds export modes with hardcoded file names: `draft.sty` `camera_ready.sty` `ready.sty`.

mpacer:
It also changes the default export mode to be the `draft.sty` rather than the `status.sty`.

mpacer:
So all of the files are hardcoded in the LaTeX files. This seems like a place for templating, but that seems… extreme.

Also the errors for what happened when it had the wrong status.sty file were impossible to decipher. That is precarious and we should figure out how to improve that experience.

mpacer:
Hrm don't merge this… but I cannot get this to build locally. Something is up with Ingo Heimbach's paper https://github.com/scipy-conference/scipy_proceedings/blob/2017/papers/ingo_heimbach/paper.rst that is making it not build. I'm not sure what it could be…. It seems to be related to the tex file that is produced being invalid. This hold even if I go back to when I released the camera ready so something must have changed on my machine. 

I'll try to figure it out.

mpacer:
@deniederhut Could you look at this… the official papers still don't have links to their videos.

deniederhut:
@mpacer let's discuss shared with in our committee call

mpacer:
@deniederhut that makes sense, but I think i might split out the changes to the build system from the updates to the papers themselves, because we really should fix the changes to the build system.

deniederhut:
Hey M! Thanks for your work on this. I encounter two issues when I execute this branch locally:

1. Building the proceedings now fails for me:

```
make proceedings
./build_template.py title.tex
(cd _build/tex && pdflatex -interaction=batchmode title.tex 1>/dev/null)
make: *** [_build/tex/title.pdf] Error 1
```

2. It looks like some of the character escaping is permanently modifying the urls, such that `https://youtu.be/yDBu_wSyw-g` gets translated to `https://youtu.be/yDBu/_wSyw-g` in the final html output:

screenshot 2018-02-24 16 37 13

stefanv:
What's the status of this PR, @mpacer ?

mpacer:
I need to look into this.

mpacer:
So, I think the issue is that there is a bug in the actual url parsing and passing through code… surprised it hasn't come up before this.

in any case, I'll factor out the build changes, as they are unrelated and easier to merge.

mpacer:
Factored out the build changes to #337, this now only changes the video information. I'm now testing for whether I can locally reproduce the underscore url bug.

mpacer:
Ok, @deniederhut 6611f7d should fix the url bug. Refactoring should have fixed the build bug.

mpacer:
Ping @deniederhut @dalippa does this look ok to you?

deniederhut:
Hey M! Sorry for missing this. I should have time in the next day or two to
have a look.

deniederhut:
Thanks so much for your work on this! It looks like there is some still some tex making its way into the html, but I'll make a separate issue for that.

PR #326, opened 2017-08-20
DOC: inline docs for tools modules

deniederhut:
Adds module, class, and function docstrings to xreftools and doitools.


mpacer:
Beautiful!

PR #336, opened 2018-02-24
change the way we set booktabs to use the internal mechanism

mpacer:
The set_table_style api was borked with 0.13… it expects a string, but
then iterates over it as though it were a list/tuple of strings.
Internally it falls back to the settings.table_style value, so we'll use
that.

Work done with @stefanv.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,9 +111,6 @@ examples of how to:
  - AMSmath LaTeX classes (included in most LaTeX distributions)
  - alphaurl (often packaged as ``texlive-bibtex-extra``, or download from
    [CTAN](https://www.ctan.org/pkg/urlbst)) urlbst BibTeX style
- - Due to a bug in the Debian packaging of ``pdfannotextractor``, you may have
-   to execute ``pdfannotextractor --install`` to fetch the PDFBox library.
-   This command may not work, but the proceedings have run correctly without it.
 
 ### Debian-like distributions:
 
--- a/README.md
+++ b/README.md
@@ -104,18 +104,18 @@ examples of how to:
 
 ## Requirements
 
+ - Install the requirements in the requirements.txt file: `pip install -r requirements.txt`
  - IEEETran (often packaged as ``texlive-publishers``, or download from
    [CTAN](http://www.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/)) LaTeX
    class
  - AMSmath LaTeX classes (included in most LaTeX distributions)
  - alphaurl (often packaged as ``texlive-bibtex-extra``, or download from
    [CTAN](https://www.ctan.org/pkg/urlbst)) urlbst BibTeX style
- - `docutils` 0.8 or later (``easy_install docutils``)
- - `pygments` for code highlighting (``easy_install pygments``)
  - Due to a bug in the Debian packaging of ``pdfannotextractor``, you may have
    to execute ``pdfannotextractor --install`` to fetch the PDFBox library.
+   This command may not work, but the proceedings have run correctly without it.
 
-On Debian-like distributions:
+### Debian-like distributions:
 
 ```
 sudo apt-get install python-docutils texlive-latex-base texlive-publishers \
@@ -123,9 +123,11 @@ sudo apt-get install python-docutils texlive-latex-base texlive-publishers \
                      texlive-bibtex-extra
 ```
 
-Note you will still need to install `docutils` with `easy-install` or `pip` even on a Debian system.
+Note you will still need to install `docutils` with `pip` even on a Debian system.
 
-On Fedora, the package names are slightly different
+### Fedora
+
+On Fedora, the package names are slightly different:
 
 ```
 su -c `dnf install python-docutils texlive-collection-basic texlive-collection-fontsrecommended texlive-collection-latex texlive-collection-latexrecommended texlive-collection-latexextra texlive-collection-publishers texlive-collection-bibtexextra`
@@ -133,9 +135,11 @@ su -c `dnf install python-docutils texlive-collection-basic texlive-collection-f
 
 ## Build Server
 
-There is a server online building the open pull requests 
+There will be a server online building the open pull requests 
 [here](http://zibi.bids.berkeley.edu:7001). You should be able to pull a built PDF 
 for review from there.
+
+TODO: update server link
 
 ## For organizers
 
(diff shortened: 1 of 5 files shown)


publisher/writer/__init__.py
```diff
@@ -57,14 +57,7 @@ def __init__(self, *args, **kwargs):
         self.figure_type = 'figure'
         self.figure_alignment = 'left'
         self.table_type = 'table'
-
-        try:
-            self.active_table.set_table_style('booktabs')
-        except TypeError:
-            raise TypeError(
-                  "You are running into an error because you are using docutils "
-                  "version {}, instead please install docutils version {}".format(dc.__version__, '0.12.0')
```
dalippa: Shouldn't this be docutils 0.14 or greater?

mpacer: well… that was deleted so it's not there anymore.

mpacer:
We should update the README.md

README.md
```diff
@@ -104,39 +104,43 @@ examples of how to:
...
    [CTAN](http://www.ctan.org/tex-archive/macros/latex/contrib/IEEEtran/)) LaTeX
    class
  - AMSmath LaTeX classes (included in most LaTeX distributions)
  - alphaurl (often packaged as ``texlive-bibtex-extra``, or download from
    [CTAN](https://www.ctan.org/pkg/urlbst)) urlbst BibTeX style
- - `docutils` 0.8 or later (``easy_install docutils``)
- - `pygments` for code highlighting (``easy_install pygments``)
  - Due to a bug in the Debian packaging of ``pdfannotextractor``, you may have
    to execute ``pdfannotextractor --install`` to fetch the PDFBox library.
+   This command may not work, but the proceedings have run correctly without it.
```
stefanv: This used to be required to extract page numbers to build the bibtex page number references.  Is there another solution in place now?

README.md
```diff
@@ -104,39 +104,43 @@ examples of how to:
...
+On Fedora, the package names are slightly different:
 
 ```
 su -c `dnf install python-docutils texlive-collection-basic texlive-collection-fontsrecommended texlive-collection-latex texlive-collection-latexrecommended texlive-collection-latexextra texlive-collection-publishers texlive-collection-bibtexextra`
 ```
 
 ## Build Server
 
-There is a server online building the open pull requests 
+There will a server online building the open pull requests
```
stefanv: will be

README.md
mpacer: I am not certain, can you be more concrete about what you're referring to?

publisher/writer/__init__.py
mpacer: @dalippa do you want me to add new error handling? I don't know where that would go now.

stefanv:
The BiBTeX files looks like this:

```
@InProceedings{ geoffrey_poore-proc-scipy-2017,
  author    = { {G}eoffrey {M}. {P}oore },
  title     = { {B}esp{O}{N}:  {E}xtensible config files with multiline strings, lossless round-tripping, and hex floats },
  booktitle = { {P}roceedings of the 15th {P}ython in {S}cience {C}onference },
  pages     = { 12 - 19 },
  year      = { 2017 },
  editor    = { {K}aty {H}uff and {D}avid {L}ippa and {D}illon {N}iederhut and {M} {P}acer }
}
```

Those page numbers need to be calculated from the final PDF, if I recall correctly, and we did those page counts using the tool in question.  Perhaps it has been replaced since, because it looks like you have the correct page numbers.

mpacer:
@stefanv I think these pages are being calculated before the final pdf is the trick… where that's actually happening… I'd need to hunt down if we wanted to know.

mpacer:
Ok, hunted down and it seems that pdfannoteextractor is for getting annotations from pdfs to be included when those pdfs are included in latex document. 

I believe that could have been used to address the page numbers a while in the past, but the start and stop values have been defined on the python side since https://github.com/scipy-conference/scipy_proceedings/commit/9ccaed9c7160262312936c8b07710ff310ac769f, 7 years ago. 

The relevant chunk of code is 
https://github.com/scipy-conference/scipy_proceedings/blob/335a59c7b9140b3575114c6ed6d2195dd170df37/publisher/build_papers.py#L28-L49

and its used iteratively inside the `if __name__ == "__main__ ":` conditional for build_papers.py like this:
https://github.com/scipy-conference/scipy_proceedings/blob/335a59c7b9140b3575114c6ed6d2195dd170df37/publisher/build_papers.py#L64-L65

mpacer:
Conclusion: I think we can remove pdfannoteextractor entirely, i'll add that to this PR as well as a little sprucing up in the build_papers.py script

stefanv:
I checked again, and my recollection was incorrect; it was used not to find page numbers, but to ensure that links in the PDF Table of Contents worked as they should:

```
commit bf83d60d97d3e30f5df541f4fe49c01fba0984e7
Author: Jarrod Millman <jarrod.millman@gmail.com>
Date:   Wed Aug 17 00:32:58 2011 -0700

    refactor to \usepackage{pax} to perserve internal links of included
    PDFs
```

and 

```
commit 43b34543139353374ec863037a4b768d4983aed2
Author: Jarrod Millman <jarrod.millman@gmail.com>
Date:   Tue Aug 16 18:41:39 2011 -0700

    refactored to \usepackage{pdfpages} in order to have links in the PDFs
    toc
```

stefanv:
Seems to be fine: http://conference.scipy.org/proceedings/scipy2017/pdfs/proceedings.pdf

Although, why is there a watermark on the papers that says "Conference Ready"?

mpacer:
> Although, why is there a watermark on the papers that says "Conference Ready"?

It's probably because they were never updated to have the non-"Conference Ready" version of files there. I'm not entirely sure why it wasn't updated. 

But it's worth noting that there are issues with all of those pages, for example, they don't show all of the links in sidebar (2012 is missing):

image

Fixing some of that was what #322 was trying to fix.

stefanv:
Sidebar can be fixed quite easily: http://conference.scipy.org/proceedings/proc_links.js

mpacer:
I know the sidebar cna be fixed quite easily… that's why I did it in #322.

stefanv:
Sorry, I only noticed that later, hence the removed comment (but you don't 
get updates for those).

PR #348, opened 2018-03-10
Fixing docutils 0.14 build issues: footnotes & code in captions

mpacer:
Basically the problem was that `self.out` was a record of the whole document up until that point, and so rather than stripping the unnecessary `%` from the footnote that was being created (which was why the line was originally introduced in https://github.com/scipy-conference/scipy_proceedings/commit/7cb4c6a8f169b3667ba6326877cbb0bf98bd9556), it was replacing it with the second element of the complete document. 

In this case it was replacing it with the header, which was causing LaTeX to break when it tried to define commands more than once.

a debugging session with @stefanv and @Carreau helped me figure this one out

### DIFF
--- a/papers/00_vanderwalt/00_vanderwalt.rst
+++ b/papers/00_vanderwalt/00_vanderwalt.rst
@@ -97,6 +97,8 @@ Or a snippet from the above code, starting at the correct line number:
    for (int i = 0; i < 10; i++) {
        /* do something */
    }
+   
+Inline code looks like this: :code:`chunk of code`.
 
 Important Part
 --------------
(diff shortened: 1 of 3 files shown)


mpacer:
This removes all the cases of `self.non_breaking_paragraph`. As that seems to have not been a docutils thing at least as far back as 0.8.1. @stefanv could you say a bit more about why you introduced it originally?

mpacer:
As mentioned in #349  we need to provide a \DUrolecite command for docutils 0.14 because otherwise code inside captions is not properly rendered. I also added inline-code inside the captions of two of the images in the vanderwalt example paper so that we could make sure that we're covering that in manual tests.

mpacer:
With this PR you should be able to run `make proceedings` and it should run without error.

stefanv:
Excellent sleuthing, @mpacer !

PR #351, opened 2018-03-10
ENH: adds issn and series-level doi

deniederhut:
Includes new ISSN and series-level DOI in config data. Adds code to include both in metadata xml upload. This currently passes Crossref's schema-checker, but this does not necessarily guarantee correct uploads.

screen shot 2018-03-10 at 1 36 43 pm


stefanv:
/ping @mpacer @deniederhut If you have some more background info, do you mind reviewing this PR?

deniederhut:
So one of the things that I've been working on since the last SciPy is acquiring an ISSN for the proceedings, such that all individual conference DOIs get linked as the products from an ongoing series of conferences. In theory, this helps with the cross-linking and discovery. 

To do this, we need to generate a series-level DOI, so the new structure looks like:

- ISSN + Series DOI (all SciPy proceedings - this is the topic of this PR)
    - Conference DOI for SciPy 2018 (code exists to create this with new proceedings)
    - Conference DOI for SciPy 2017 (already have this)
        - Paper DOI for Poore 2017
        - Paper DOI for Niederhut 2017
        - etc.

The organization that administers ISSN requests that DOIs be constructed following [these guidelines](http://www.issn.org/understanding-the-issn/issn-uses/use-of-issn-in-doi/).

Submitting these to crossref.org requires an xml file with [the following schema](http://data.crossref.org/reports/help/schema_doc/4.4.1/index.html). The documentation is not super easy for humans to parse, so the screenshot above is meant to serve as some limited guarantee of correctness.

mpacer:
@stefanv I don't have too much in the way of background info… @deniederhut could you provide a bit more background info at tomorrow's meeting?

mpacer:
@deniederhut could you move that into a "xref_docs.md" inside the publisher (along with any other pertinent information about the problem)? 

eventually we'll get around to making proper docs, but for now even a simple markdown file would be hugely helpful.

PR #354, opened 2018-03-11
add doi example to 00_bibderwalt citation

mpacer:
This creates a linked doi for the associated doi in the bibliography entry: at `http://dx.doi.org/<doi>`

So in this case, it is http://dx.doi.org/10.1017/CBO9780511808432. 

I also added "documentation" to the 00_bibderwalt example.

Also, I reverted 5dd888dd4fe72d9b168e33a05758bbf8f368c68e

`self.non_breaking_paragraph` isn't a docutils feature, but it is a custom feature we use in a super roundabout fashion. 

Basically we have an elif looking for  `self.non_breaking_paragraph` in `self.visit_paragraph()`. The only point of that elif is to preëmptively catch stuff before it had newlines added to it in https://github.com/scipy-conference/scipy_proceedings/blob/82253931a114463b31856f93169b5a65553eae3a/publisher/writer/__init__.py#L306-L310

There is probably a better way to implement this, but I can't think of it off the top of my head so I reverted my previous commit that had introduced issues with footnote spacing.


dalippa:
👍

deniederhut:
Woo! Thanks M!

PR #360, opened 2018-03-18
DOC: doitools and crossref metadata

deniederhut:
At the request of the excellent @mpacer, adds some background info
on DOIs, ISSN, and Crossref to publisher/README.


mpacer:
Me too, merging.

PR #363, opened 2018-04-07
New year docs

mpacer:
This adds docs for how to get the proceedings ready for the next year. 

It also adds the new git workflow and branch logic as discussed at the meeting on March 25, 2018.


mpacer:
I added the todo, because while our repo is pretty clean, and I think we could probably clean out the 2015, 2016, and 2017 proceedings branches into their own GitHub repos like was done in earlier years.

This would substantially reduce the size of cloning the current repo.

mpacer:
@dalippa Could you review this?

dalippa (approved):
👍

dalippa:
@mpacer Excellent work on this. Great explanations. You may want to squash before merge.

mpacer:
I don't actually want to squash it — I wrote it all in one go and then tried to organise the changes into a sequence of semantically related tasks so that they could be hunted through separately when someone runs into an issue.

mpacer:
Kk going to merge.

PR #369, opened 2018-04-14
Update mailings

mpacer:
So this does a few things to improve our existing email generating system.

The first thing it does is it standardises a few of the fields that had previously been template specific.

The second thing it does is introduces a new Mailer class which contains the common logic for all of the email scripts and reduces the amount of boilerplate we need to keep around in each script.

this new Mailer object is intended to be subclassed with the ability to customise the fields needed for the particular template with the `custom_data` property, as can be seen in the call_for_papers example.

@dalippa and @deniederhut If you have a chance to look at this, I'd appreciate it… i'm probably going to use this to send out the emails as soon as I have the list and I figure out how to hook this up to my gmail.

My (hard) todos on this are:

- [x] Get session to connect appropriately
- [x] Fix/clarify authentication logic
- [x] setup new authors json based on the papers info

My (soft) todos on this are:
- [ ] update other scripts to use Mailer
- [x] setup nicer addresses for the recipients

### DIFF
--- a/publisher/mail/_mailer.py
+++ b/publisher/mail/_mailer.py
@@ -2,7 +2,8 @@ import argparse
 import smtplib
 import os
 import getpass
-from email.mime.text import MIMEText
+
+from contextlib import contextmanager
 
 import sys
 sys.path.insert(0, '..')
@@ -49,6 +50,16 @@ def get_password(sender):
 
 
 def email_addr_from(name_email):
+    """
+    Parameters
+    ----------
+    name_email: dict
+        A dictionary of names and emails. 
+        Expected fields: 
+            'name': name of person
+            'email': email of person
+    """
+    
     return '"%s" <%s>' % (name_email['name'], name_email['email'])
 
 
@@ -111,3 +122,170 @@ def send_template(sender, recipient, template, template_data,
 
     session.sendmail(sender['name'], recipient, message)
     session.quit()
+
+
+def load_data_file(source_file):
+        if os.path.exists(source_file):
+            return cfg2dict(source_file)
+        else:
+             print('file at {} not found'.format(os.path.abspath(s)))
+
+
+def load_data_sources(sources):
+    """Combines dictionaries starting with the first included.
+    """
+    new_dict = {}
+    for s in sources:
+        if isinstance(s, dict):
+            data = s
+        elif isinstance(s, str):
+            data = load_data_file(s)
+        new_dict = {**new_dict, **data}
+    return new_dict
+
+class Mailer:
+    
+    def __init__(self,
+                 sender=None,
+                 template='',
+                 base_dir='../mail/templates/',
+                 data_sources=None,
+                 smtp_server='smtp.gmail.com', 
+                 smtp_port=587,
+                 dry_run=True,
+                 ):
+        if sender is not None and isinstance(sender, dict):
+            self.sender = sender
+        
+        
+        if data_sources is None:
+            data_sources = []
+        elif isinstance(data_sources, (str, dict)):
+            data_sources = [data_sources]
+        elif isinstance(data_sources, list):
+            data_sources = data_sources
+        # we always need the email metadata, so let's make that default
+        
+        self.data_sources = ['./email.json'] + data_sources
+        
+        self.dry_run = True
+        self.base_dir = base_dir
+        self.template = template
+        self.smtp_server = smtp_server
+        self.smtp_port = smtp_port
+        self.aux_data = {}
+        
+    # TODO: update this to take a list of recipient dicts (name, email)
+    # that will require changing the rest of the recipients logic
+    # def recipient_emails(self):
+    #     """ get emails from list 
+    #     Parameters
+    #     ----------
+    #     name_email: dict
+    #         A dictionary of names and emails. 
+    #         Expected fields: 
+    #             'name': name of person
+    #             'email': email of person
+    #     """
+    #     return ', '.join([email_addr_from(r) for r in self.recipients]))
+        
+    def recipient_greeting(self, names):
+        if len(names) == 1:
+            name_string = names[0]
+        else:
+            name_string = ', '.join(names[:-1]) + ', and ' + names[-1]
+        return name_string
+            
+
+    @property
+    def recipients(self):
+        return 'blah@blah.org'
+        
+    @property
+    def sender(self):
+        return self._sender if self._sender else self.template_data["sender"]
+        
+    @sender.setter
+    def sender(self, value):
+        if (isinstance(value, dict) and all(k in value for k in ('name','email'))):
+            self._sender = value
+        else:
+            raise ValueError("You tried to set {} as the sender "
+                             "but it has no 'name' and 'email' keys.".format(value))
+        
+    @property
+    def password(self):
+        if not self.dry_run and not self._password:
+            self._password = getpass.getpass(sender + "'s password:  ")
+        return self._password
+    
+    @property
+    def template_data(self):
+        return {**load_data_sources(self.data_sources), **self.aux_data}
+        
+    @property
+    def template(self):
+        return self._template if os.path.exists(self._template+'.tmpl') else ""
+            
+    @template.setter
+    def template(self, value):
+        self._template = resolve_template(value, self.base_dir)
+        
+    def prep_data(self, data=None):
+        data = data if (data and isinstance(data, dict)) else {}
+        self.aux_data = {**self.common_data, **self.custom_data, **data}
+    
+    @property
+    def common_data(self):
+        return {'editor_email_string':  editor_email_string(),
+                'committee': create_committee()}
+    
+    @property
+    def custom_data(self):
+        return {}
+            
+    def send_from_template(self, recipients=None, data=None):
+        """
+        
+        Parameters
+        ----------
+        recipient: str
+            email 
+        
+        """
+        data = {} if data is None else data
+        recipients = self.recipients if recipients is None else recipients
+        self.prep_data({**data, 'recipients':recipients})
+
+        message = _from_template(self.template, self.template_data)
+        
+        if self.dry_run:
+            self.display_message(recipients, message)
+        else:
+            self.send_mail(recipients, message)
+
+
+    def display_message(self, recipients, message):
+        print('Dry run -> not sending mail to %s' % recipients)
+        print("=" * 80)
+        print(message)
+        print("=" * 80)
+    
+    def send_mail(self, recipients, message):
+        with self.session() as session:
+            session.sendmail(self.sender['name'], recipients, message)
+
+    @contextmanager
+    def session(self):
+        
+        self.get_password(self.sender['login'])
+        print('-> %s' % self.recipients)
+        session = smtplib.SMTP(self.smtp_server, self.smtp_port)
+
+        session.ehlo()
+        session.starttls()
+        session.ehlo
+        session.login(self.sender['login'], self.password)
+        yield session
+
+        session.quit()
--- a/publisher/mail/_mailer.py
+++ b/publisher/mail/_mailer.py
@@ -251,6 +251,12 @@ class Mailer:
         if data is None:
             data = cfg2dict(proc_conf).get('proceedings', {})
         return cls.fancy_emails(data, name_key='editor', email_key='editor_email')
+        
+    @classmethod
+    def editor_email_list(cls, data=None):
+        if data is None:
+            data = cfg2dict(proc_conf).get('proceedings', {})
+        return cls.fancy_email_list(data, name_key='editor', email_key='editor_email')
 
     @classmethod
     def create_committee(cls, data=None):
@@ -312,6 +318,10 @@ class Mailer:
     @property
     def custom_data(self):
         return {}
+    
+    @property
+    def cc_list(self):
+        return self.editor_email_list()
             
     def send_from_template(self, recipients_list=None, data=None):
         """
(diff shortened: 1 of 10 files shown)


mpacer:
@gforsyth this is the email system we currently have up and running, if you have any suggestions on how to adapt it for the json that you'll be creating I'd appreciate it!

publisher/mail/_mailer.py
```diff
@@ -46,25 +50,67 @@ def get_password(sender):
...
+    """
+    
     return '"%s" <%s>' % (name_email['name'], name_email['email'])
 
 
+def create_committee(data=None):
+    if data is None:
+        data = cfg2dict(proc_conf)
+    proc = data.get('proceedings', {})
+    editors = proc.get('editor', [])
```
dalippa, line 70: to be consistent, consider using editor or editors, not both

publisher/mail/_mailer.py
```diff
@@ -46,25 +50,67 @@ def get_password(sender):
...
+        A dictionary of names and emails. 
+        Expected fields: 
+            'name': name of person
+            'email': email of person
+    """
+    
     return '"%s" <%s>' % (name_email['name'], name_email['email'])
 
 
+def create_committee(data=None):
```
dalippa, line 66: Why not use `**kwargs` here?

publisher/mail/_mailer.py
```diff
@@ -46,25 +50,67 @@ def get_password(sender):
...
-    message = _from_template('../mail/templates/' + template, template_data)
+    
+    message = create_message(recipient, template, template_data)
 
     if args.dry_run:
         print("=" * 80)
         print(message)
         print("=" * 80)
-
         return
```
dalippa, line 116: Consider moving the return statement below the `smtplib.SMTP` call so you can use the dry run to verify credentials against the mail server.

publisher/mail/_mailer.py
```diff
@@ -76,3 +122,170 @@ def send_template(sender, recipient, template, template_data,
 
     session.sendmail(sender['name'], recipient, message)
     session.quit()
+
+
+def load_data_file(source_file):
+        if os.path.exists(source_file):
+            return cfg2dict(source_file)
+        else:
+             print('file at {} not found'.format(os.path.abspath(s)))
```
dalippa: If the data file doesn't exist, would it be better to raise an exception since the dependent functions cannot do what they need to do otherwise?

publisher/mail/_mailer.py
```diff
@@ -76,3 +122,170 @@ def send_template(sender, recipient, template, template_data,
...
+
+
+def load_data_sources(sources):
```
dalippa, line 138: I think there is a way to merge dictionaries with `x.update(y)`. Why not use that instead of creating the new_dict over and over?

publisher/mail/_mailer.py
```diff
@@ -46,25 +50,67 @@ def get_password(sender):
...
+    editors = proc.get('editor', [])
+    # editor_email = [x.strip() for x in proc.get('editor_email', '').split(',')]
+    editor_email = proc.get('editor_email')
+    assert len(editors) == len(editor_email)
+    return [{"name": name, 
+             "email": email} 
+            for name, email 
+            in zip(editors, editor_email)]
+
+def editor_email_string(data=None):
```
dalippa, line 80: Minor comment: these methods are implementation details used by the class; from a design standpoint, prefixing "private" methods with `_` is the usual convention.

publisher/mail/_mailer.py
```diff
@@ -76,3 +122,170 @@ def send_template(sender, recipient, template, template_data,
...
+        if len(names) == 1:
+            name_string = names[0]
+        else:
+            name_string = ', '.join(names[:-1]) + ', and ' + names[-1]
+        return name_string
+            
+
+    @property
+    def recipients(self):
+        return 'blah@blah.org'
```
dalippa: Is this a comma or semicolon delimited list? Usually a property that is in plural form returns a list, not a single string. At a minimum, consider documenting so classes that override this method would know the contract.

publisher/mail/_mailer.py
```diff
@@ -76,3 +122,170 @@ def send_template(sender, recipient, template, template_data,
...
+            data_sources = []
+        elif isinstance(data_sources, (str, dict)):
+            data_sources = [data_sources]
+        elif isinstance(data_sources, list):
+            data_sources = data_sources
+        # we always need the email metadata, so let's make that default
+        
+        self.data_sources = ['./email.json'] + data_sources
+        
+        self.dry_run = True
```
dalippa: You use `self._sender` for member variables later, but `self.dry_run` here. Consider a consistent naming convention.

publisher/mail/_mailer.py
```diff
@@ -76,3 +122,170 @@ def send_template(sender, recipient, template, template_data,
...
+        return name_string
+            
+
+    @property
+    def recipients(self):
+        return 'blah@blah.org'
+        
+    @property
+    def sender(self):
+        return self._sender if self._sender else self.template_data["sender"]
```
dalippa: What if the template_data has an empty string or something? Consider `self.template_data.get("sender")` which would raise an exception or `self.template_data.get("sender", "cochairs@scipy.org")` or something like that to provide a default.

publisher/mail/_mailer.py
```diff
@@ -76,3 +122,170 @@ def send_template(sender, recipient, template, template_data,
...
+
+        message = _from_template(self.template, self.template_data)
+        
+        if self.dry_run:
+            self.display_message(recipients, message)
+        else:
+            self.send_mail(recipients, message)
+
+
+    def display_message(self, recipients, message):
```
dalippa: maybe `display_dryrun_message`?

publisher/mail/_mailer.py
```diff
@@ -46,25 +50,67 @@ def get_password(sender):
...
     else:
         get_password(sender['login'])
         print('-> %s' % recipient)
-
-    template_data['email'] = recipient
-    message = _from_template('../mail/templates/' + template, template_data)
+    
+    message = create_message(recipient, template, template_data)
 
     if args.dry_run:
```
dalippa, line 112: consider making the dry run message in the class below a static method and call it here.

publisher/mail/email.json
```diff
@@ -1,22 +1,24 @@
...
+   "readme_url": {
+     "authors": "https://github.com/scipy-conference/scipy_proceedings#instructions-for-authors",
+     "reviewers": "https://github.com/scipy-conference/scipy_proceedings#instructions-for-reviewers"
+   },
 
-   "reviews_url": "http://jarrodmillman.com/scipy2011-reviews",
-   "built_proceedings_url": "http://jarrodmillman.com/scipy2011",
+   "built_proceedings_url": "http://procbuild.scipy.org",
 
    "reviewers": [
```
dalippa, line 24: Change the reviewers too?

publisher/mail/templates/call_for_papers.txt.tmpl
```diff
@@ -0,0 +1,46 @@
...
+Subject: [{{conference}}] invitation to submit to proceedings
+To: {{recipients | html}}
+Cc: {{editor_email_string | html}}
+MIME-Version: 1.0
+Content-Type: text/plain
+
+Dear {{author}},
+
+Congratulations on having your abstract accepted for this year's event -- we 
+look forward to seeing your poster or presentation! As in the past, SciPy2018
```
dalippa: Use `{{year}}` here as well

dalippa (changes requested):

publisher/mail/templates/call_for_papers.txt.tmpl
mpacer: I think this might be `{{conference}}`, but definitely a good idea.

publisher/mail/_mailer.py
```diff
@@ -76,3 +122,170 @@ def send_template(sender, recipient, template, template_data,
 
     session.sendmail(sender['name'], recipient, message)
     session.quit()
```
dalippa: try/finally on session quit?

publisher/mail/_mailer.py
mpacer: that might be possible… but I generally try to avoid using **kwargs and then testing if it's none. But I see where you're coming from.

publisher/mail/_mailer.py
mpacer: I was using underscores to indicate private member variables not just member variables in general. I usually do that whenever I have a property for the public member variable, but still need to store the info somewhere.

publisher/mail/_mailer.py
mpacer: You're right, there's gotta be a better way than this… i was trying to avoid update because of the way it changes everything in place, but you can't unpack dictionaries in a comprehension as far as I could tell.

publisher/mail/_mailer.py
mpacer: yes, I'll make that happen.

publisher/mail/_mailer.py
mpacer: If we use `self.template_data.get("sender")` it would return None, which wouldn't return an exception locally, it would raise an exception somewhere further down the line. 

also sender looks to be expected as a dict not a string, so I should probably make that work consistently.

mpacer: This is making me realise just how much of this logic is only for validating and working with these "name" "email" pairs… I should probably factor that out into its own class…

publisher/mail/email.json
mpacer: They're just test reviewers… it actually helped me realise that because they were from so long ago.

publisher/mail/_mailer.py
mpacer: I got this from the old way of approaching the problem… we could do that, but it would require passing in a password each time you wanted to do the dry_run, which would have put a huge amount of overhead when trying to iterate quickly over this code.

publisher/mail/_mailer.py
mpacer: I moved these methods inside the class, and now they should be able to be overwritten if needed… whether that would be necessary I'm not sure, but I can make them private if you want.

publisher/mail/_mailer.py
```diff
@@ -175,27 +175,60 @@ def __init__(self,
...
+        -----------
+        data: dict
+            this should have at least two fields, the vals of name_key and email_key
+        name_key: str
+            this is the key to indicate names
+        email_key: str
+            this is the key to indicate emails
+        """
+        people_gen = (p for p in cls.fancy_prep(data, name_key=name_key, email_key=email_key))
+        are_str = lambda x: (isinstance(x['name'], str) and isinstance(x['email'], str))
```
dalippa, line 248: not sure what `are_str` is here. Do you mean `is_str`?

publisher/mail/_mailer.py
mpacer: Honestly, I was treating this as a string because that's sort of how it was being done before. 

In any case, I was slightly inclined to raise a `NotImplementedError`… since it shouldn't need to be set unless someone wants to do so explicitly.

publisher/mail/_mailer.py
mpacer: Great idea… I'll implement that shortly.

publisher/mail/mail_call_for_papers.py
```diff
@@ -2,30 +2,32 @@
...
-template = args.template or 'call_for_papers.txt'
 
-scipy_proc = cfg2dict(proc_conf)
-toc = cfg2dict(toc_conf)
 
 class AuthorMailer(Mailer):
     @property
     def recipients(self):
-        return ', '.join(self.template_data['author_email'])
+        # this is for testing
```
dalippa: Minor: Consider moving this into a test file and creating a "fake" class `TestAuthorMailer` that subclasses `AuthorMailer` and does this.

publisher/mail/templates/call_for_papers.txt.tmpl
```diff
@@ -22,8 +22,11 @@ scientific computing in Python, such as:
 - hardware systems designed to support the Python stack
 - applications of scientific computing in education
 
-Papers must be submitted as pull requests by {{submission_date}}. At that
-point, the six-week open review period will begin. During this time, members
+To submit a paper to the proceedings, submit a pull request to {{github_url}}.
+
+Papers must be submitted as pull requests by {{submission_date}}.
```
dalippa, line 27: Consider combining this sentence with the prior and following sentences:

"To submit a paper to the proceedings, submit a pull request to {{github_url}} by {{submission_date}}, when the six-week open review period begins."

publisher/mail/_mailer.py
mpacer: I'll add this in both here and the method.

dalippa (approved):
Minor recommendations. Feel free to fix them before merging, or after.

publisher/mail/_mailer.py
mpacer: actually… doesn't the context manager already handle this?

publisher/mail/templates/call_for_papers.txt.tmpl
mpacer: I originally did it and it read rather long & not obviously containing crucial information. These are both important pieces of information so I wanted them to be highlighted separately.

mpacer:
This was able to send off most of the emails we needed to send. With minimal issues.

For some reason it swallowed the `+scipyproc` from mpacer.phd+scipyproc@gmail.com, not sure why.

mpacer:
Ah — I think it it probably swallowed that because of the "login" value. I'll play around with that a bit to see if adding `.` and `+*` works as a login value.

PR #371, opened 2018-04-18
ENH: publish DOIs

### LINKED ISSUE #332: ENH: publish DOIs and ISSN
To make the unique identifiers for the conference, proceedings, and individual papers easier to find, we may want to include them on the SciPy website. A reasonable strategy might be:

1. Include paper DOIs in bibfile and html page
2. Include proceedings DOI and conference ISSN at http://conference.scipy.org/proceedings/scipy2017/
3. Include conference DOI and ISSN at conference.scipy.org (this will require PR to https://github.com/scipy-conference/conference.scipy.org)

The identifiers themselves end up in the json metadata files while the proceedings are built, so it should be possible to inject them into the tex, html, and/or bibfiles.

deniederhut:
Adds conference, issue, and paper DOIs to copyright page, html pages, and bibfiles. Closes #332. 

## Rendered examples:

screenshot 2018-04-18 13 16 18
screenshot 2018-04-18 13 09 56
screenshot 2018-04-18 13 10 18
screenshot 2018-04-18 13 10 27

### DIFF
--- a/publisher/Makefile
+++ b/publisher/Makefile
@@ -38,21 +38,21 @@ front-pdf: title-pdf copyright-pdf organization-pdf students-pdf
 papers:
 	./build_papers.py
 
-proceedings: front-pdf papers $(TEXDIR)/proceedings.tex
+proceedings: papers $(TEXDIR)/proceedings.tex front-pdf
 	($(TEX2PDF) proceedings 1>/dev/null)
 	($(TEX2PDF) proceedings 1>/dev/null)
-	cp $(TEXDIR)/proceedings.pdf $(PDFDIR)/proceedings.pdf 
+	cp $(TEXDIR)/proceedings.pdf $(PDFDIR)/proceedings.pdf
 
-html: 
+html:
 	python build_html.py
 	-convert $(STATIC)/logo.png -resize x100 $(HTMLDIR)/logo.png
 
-zip: 
+zip:
 	cp -r $(HTMLDIR) $(YEARDIR)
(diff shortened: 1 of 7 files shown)


publisher/Makefile
```diff
@@ -38,21 +38,21 @@ front-pdf: title-pdf copyright-pdf organization-pdf students-pdf
 papers:
 	./build_papers.py
 
-proceedings: front-pdf papers $(TEXDIR)/proceedings.tex
+proceedings: papers $(TEXDIR)/proceedings.tex front-pdf
```
mpacer, line 41: cute! I'm happy this worked!

mpacer:
Is there any way to get the dois into the actual papers themselves? That’s the one bit that I’d like to see in this before we merge, because that’s going to require a different approach (I think) from how this addresses getting dois into individual papers.

Note I really want to avoid the solution of running the build papers scripts twice. Is it possible to pull the logic of building a single paper into a context manager (or something) that we then can use to inject the info about paper dois so that by the final build step it’s including it?

mpacer:
Also apologies for not saying that before; I thought I had already left that comment but I must have accidentally not done so.

deniederhut:
AFAIK there is no existing logic for injecting data into the rst files during processing. We decided at the last committee meeting to postponse discussion/work on that particular thing until after the conference.

mpacer:
Unfortunately, I don't remember that conversation — this is an improvement on the current state of affairs, so I'll merge it.

PR #376, opened 2018-05-11
Paper: "Design and Implementation of pyPRISM: A Polymer Liquid-State Theory Framework"

martintb:


hameerabbasi:
There seems to be an issue with the build server... The PDF built seems to be for the `papers/00_bibderwalt` directory rather than the `papers/tyler_martin` directory. I'm not sure what would be best here, but I see two ways to solve this:

- Patch to the correct directory on the build server.
- Close this PR and open a new one.

I'm not entirely sure what the best way forward is here...

deniederhut:
Interesting -- these should be explicitly ignored as part of [this list](https://github.com/scipy-conference/procbuild/blob/1aa60f1c8c3652d75b52e65554bf5565de556e1f/procbuild/builder.py#L14). @mpacer ?

martintb:
My apologies but I'm confused. I just rebuilt the paper and it seems to produce the correct pdf....

hameerabbasi:
Well, that solves that then. 😅

gtelzur:
Hello,
I was asked to review your paper. I would appreciate if you could point me to the (final?) PDF.
Best regards,
Guy

martintb:
Hi Guy,

If you go to [http://procbuild.scipy.org/](http://procbuild.scipy.org/), you should be able to view the final PDF. If you don't see a "View PDF" button next to our paper, click the arrows on the right to build it. 

Let me know if you can't see the final, compiled PDF.

Thanks!
Tyler

gtelzur:
Hi Tyler,

Now I see it.
Thanks,

Guy

gtelzur:
Hello,
I finished reading the paper. I think it is an excellent paper and I recommend it according to any common metric (relevance; originality; significance; presentation; abstract,  introduction, figures, summary, references and paper formatting - they are all adequate). 
Three comments:
1) On page 6: I think it would be nice to further explain the tradeoffs between the slow, cpu intentnsive MC/MD way and the quick PRISM way. How come a HPC problem that may run for days all of a sudden can be solved in seconds on a laptop?
2) I liked very much the holistic approach of solving a scientific problem using advanced computing methods and paying attention to the pedagogical aspect!
3) Please drop me an email when your final version will be ready so that I will have a final look at it.
Good luck.
Guy Tel-Zur

martintb:
I've modified the end of the **Example Discussion** section to be as follows. Please let me know if this clarifies your Point 1 above. I am happy to modify to add more description if this is still unclear. 

> While it would be feasible to study this polymer nanocomposite system *via*
> simulation methods such as MD or MC, the use of PRISM theory offers some
> distinct advantages. PRISM theory does not suffer from finite-size or
> equilibration effects, both of which limit simulation methods. Furthermore, a
> simulation of sufficient size to study the large nanoparticles and relatively
> long polymer chains in this example would require many hours to days of CPU or
> GPU time from a supercomputing resource. This is due to the computational
> expense of evaluating the pairwise interactions at each simulated configuration
> and the many millions of configurations that must be generated in order to
> properly equilibrate and sample such a nanocomposite.  In contrast, PRISM theory
> allows for the globally averaged spatial correlation functions to be solved for
> directly, rather than by iteratively sampling the positions of each atom or
> coarse-grained bead. In this way, the PRISM equation can be numerically solved
> in seconds even on modest hardware such as a laptop computer.  In addition, once
> the full set of pairwise spatial correlation functions is solved for, a variety
> of properties can quickly be screened without having to process large simulation
> trajectories.
> 
> PRISM theory provides a powerful alternative or complement to traditional
> simulation approaches, but we should note that it is not without limitation.
> There are restrictions on the types of systems and thermodynamic state points to
> which PRISM theory can be applied and the numerical closures are approximations
> and therefore sources of error.  See Section IV.D of :cite:`pyPRISM` for a
> discussion on the known limitations of PRISM theory.

nicoguaro:
Hello @martintb, I was asked to review your paper as well.

I would adhere to @gtelzur points. Regarding point 1 above, I think that it is still not clear enough why PRISM calculations are much faster than "traditional" simulation methods. I mean, you are mentioning this

> PRISM theory allows for the globally averaged spatial correlation functions to be solved for directly, rather than by iteratively sampling the positions of each atom or coarse-grained bead.

But describing a little bit how these correlation functions are obtained (or where they come from) would not hurt.

I also think that the pedagogic aspect of the paper adds value to it. In that sense I would like to mention that I went through the documentation looking for a first example, but it takes several pages until one appears, maybe it would be a good idea to provide a short example in the opening page … and explain it later.

Regarding the software, I installed without trouble in a computer running Linux Mint 18.3. When I ran ``NB6.Cases.Studies.Nanocomposites`` the following error appear

> ValueError: Jacobian inversion yielded zero vector. This indicates a bug in the Jacobian approximation.

martintb:
Is the language below clearer? In short, MD or MC involves the generation of configurations while PRISM theory is simple a set of equations that are numerically solved. 

> In contrast, PRISM theory
> can be numerically solved in seconds even on modest hardware such as a laptop
> computer. This is because, unlike MD or MC, solving PRISM theory does not
> involve generating molecular configurations, but rather is a set of integral
> equations which are numerically solved for the spatial correlation functions,
> :math:`H_{\alpha,\beta}(r)` and :math:`C_{\alpha,\beta}(r)`. This numerical
> solution process is briefly described above at the end of the `PRISM Theory`_
> section and is described in detail in Section II.E of :cite:`pyPRISM`. In
> addition to the computational performance benefits of PRISM theory over MD or
> MC, once the full set of pairwise spatial correlation functions is solved for, a
> variety of properties can quickly be screened without having to process large
> simulation trajectories.


> I also think that the pedagogic aspect of the paper adds value to it. In that sense I would like to mention that I went through the documentation looking for a first example, but it takes several pages until one appears, maybe it would be a good idea to provide a short example in the opening page … and explain it later.

Great suggestion! The front page of the docs now has an example that points to a detailed discussion of that example. 

> Regarding the software, I installed without trouble in a computer running Linux Mint 18.3. When I ran NB6.Cases.Studies.Nanocomposites the following error appears

Unfortunately this is a known bug that we can't seem to figure out because it only occurs on some users machines. That tutorial example runs fine on several of my machines and those of others that I have directly tested. See usnistgov/pyPRISM#12 (which I thought I had created a while ago). Any help or suggestions on this would be greatly appreciated.

gtelzur:
Hello,
We are nearly at the deadline. I am ok with the recent modifications and from my point of view the paper is ready.

nicoguaro:
@proceedings, I also consider that the paper is ready.

deniederhut:
Thanks reviewers! To clarify, this week is the deadline for reviews - authors have another two weeks to work with their reviewers to get their papers ready for the proceedings.

nicoguaro:
> Unfortunately this is a known bug that we can't seem to figure out because it only occurs on some users machines. That tutorial example runs fine on several of my machines and those of others that I have directly tested. See usnistgov/pyPRISM#12 (which I thought I had created a while ago). Any help or suggestions on this would be greatly appreciated.

@martintb, I tried the Notebooks in another computer and they worked. The funny thing is that the error appear in the mybinder version.

ricleal:
When I first saw the paper I also had issues with the explanation of PRISM calculations being faster than "traditional" simulation methods. This was already addressed as I can see. 
@proceedings For me the paper is ready.

@martintb Completely separated from the review: Have you though about implementing the Debye method in the GPU? It might give you a few orders of magnitude in performance...

martintb:
@ricleal Absolutely! I actually had a C++ GPU code for the Debye method that I wrote a number of years ago but I found that the limited memory of GPU's (at the time) made the calculation somewhat tricky to write efficiently. My lack of experience with CUDA was likely also a factor. With more modern Python tools that support GPU execution, I should definitely revisit this. 

For a performant but non-python solution to this problem, I have also written a C++ MPI parallelized tool for the CPU called [`correlate`](https://github.com/martintb/correlate). While I haven't benchmarked it against other codes, `correlate` scales well to hundreds CPUs and has been applied to systems with ~1e6 atoms/beads.

PR #282, opened 2017-05-30
Paper:  Composable Multiprocessing and Multithreading for Numeric Libraries 

anton-malakhov:
Title: Composable Multi-Threading and Multi-Processing for Numeric Libraries

Authors:	Anton Malakhov, Anton Gorshkov, Terry Wilmarth

Keywords: Multi-threading, Multi-processing, Over-subscription, Parallel Computations, Nested Parallelism, Multi-core, Python, GIL, Dask, Joblib, NumPy, SciPy, Numba, TBB, OpenMP

PR #384, opened 2018-05-14
Paper: Addressing Multithreading and Multiprocessing in Transparent and Pythonic Methods

anton-malakhov:
Title: Composable Multi-Threading and Multi-Processing for Numeric Libraries

Authors:	Anton Malakhov, David Liu, Anton Gorshkov, Terry Wilmarth

Keywords: Multi-threading, Multi-processing, Over-subscription, Parallel Computations, Nested Parallelism, Multi-core, Python, GIL, Dask, Joblib, NumPy, SciPy, Numba, TBB, OpenMP


anton-malakhov:
@triskadecaepyon, welcome! I used unpublished paper https://github.com/scipy-conference/scipy_proceedings/pull/282 from the last year proceeding as a starting point and added you as co-author.

hameerabbasi:
Perhaps a does not build on server label? There's no "View PDF" link there.

deniederhut:
You should be able to see it now. If there is no "view pdf", you can ask for the paper to be rebuilt by clicking this button:
screenshot 2018-05-19 20 06 20
on the righthand side.

katyhuff:
Hi @anton-malakhov This seems to still be titled with last year's title. Is this a real submission?

anton-malakhov:
Yep, I hope to publish this work this year. Waiting when @triskadecaepyon comes up with a different focus in the paper, which is more in line with the abstract, then we can change the title

sosey:
Hi @anton-malakhov - I was assigned to review the paper, I think overall it's clearly thought out and well put together. I just have a few comments below, mostly grammar, you might consider addressing. The comments are based on the PDF I grabbed from a build done today.

**Just some minor grammar corrections**:
- at the end of the motivation section: `Even if problem size unlimited as it is for "Big Data", it still has to decomposed into` --> `Even if the problem size is unlimited, as it is for "Big Data", it still has to be decomposed into`
- `For example, Intel OpenMP* runtime library (used by NumPy/SciPy) may keep its threads active to facilitat` --> need to add an e to facilitate
- page 3: `By default itis 2, which means that in our example,` --> `By default it is`
- references like this ` in p3.5` should probably refer to `section 3.5`?
- it might be more readable to change `with -f 1 specified on the command line` to `with an oversubscription factor of 1 specified on the command line`
- suggest rewording  ```To enable this mode, KMP_COMPOSABILITY environment variable should be set, for example: env KMP_COMPOSABILITY=mode=exclusive python app.py
This enables each OpenMP parallel region to run exclusively, eliminating the worst oversubscription effects.```   --> ```Setting the KMP_COMPOSABILITY environment variable as follows should enable each Open MP parallel region to run exclusively, eliminating the worst oversubscription effects: env KMP_COMPOSABILITY=mode=exclusive python app.py ```
- suggest rewording for clarity: `The third approach using smp module and specifying -f 1 does similar optimizations automatically, and shows the same level of performance as for OMP_NUM_THREADS=1` --> `The third approach using smp module and specifying an oversubscription factor of 1 ...`
- suggest replacing "less deep" with "shallower" here: `time, we wish to keep the interface for this simple, requiring less deep knowledge`
- `Figure 3 shows benchmark execution time using the same modes as inthe QR ` --> `Figure 3 shows benchmark execution time using the same modes as in the QR `
- `As a result, just a few threads are used inthe first stage, ` --> `As a result, just a few threads are used in the first stage, `
- `These improvements are achieved with all different ap- proaches, ` --> `These improvements are all achieved with different approaches`

**Medium grammar change suggestions**:

- This could be made a stronger paragraph by changing the wording: `This empirical evidence might not be enough to properly generalize our experience while there are a lot of other variables and moving targets, but we did our best to summarize conclusions and suggest practical guidance in the following table as a starting point for tuning performance of applications with nested parallelism:` --> `The evidence presented in this paper doesn't explore the full problem parameter space, however it does provide practical guidance that can be used as a starting point to tune the performance of applications with nested parallelism. ` I think ending here with a period is fine.

- suggest rewording: `In this paper, we discussed the necessity of broader usage of nested parallelism for multi-core systems, and defines threading composability.` --> `This paper provides a working definition for threading composability, specifically discussing the necessity for broader usage of nested parallelism on multi-core systems. `

- Suggest rewording: `We address the issues, such as GIL and oversubscription, of Python programs and libraries that use parallelism with multi-core systems. These issues affect the performance of Python programs that use libraries like NumPy, SciPy, SciKit- learn, Dask, and Numba.` --> `We also addressed performance issues related to the GIL and oversubscription of threads,  for python libraries using parallelism with multi-core processors, such as NumPy, SciPy, SciKit-learn, Dask, and Numba.`

- `We have compared suggested approaches and provided recommendations for when it makes sense to employ each of them.` --> `Our comparison of the suggested approaches provides recommendations for when it makes sense to employ each of them.`

**Other comments**
- I don't think I saw any mention of python2 vs python3, outside of your list of packages installed for the testing. I know there are many in the community still working on converting their code base, it may be worth touching on this aspect especially since you mention concurrent.futures. Perhaps in your paragraph discussing blocking I/O and async.

deniederhut:
Hi @anton-malakhov ! Is your paper ready for the reviewers to take another look?

anton-malakhov:
@deniederhut I'll update performance diagrams, but they are not going to change much, though, some numbers in the text might need corrections because of that. @triskadecaepyon might want to add few words on what's the most pythonic approach here to be better aligned with the talk. However, we decided that the title of the paper will stay the same, not matching the title of the talk of this year (though it's matching last year's talk). So, it's not yet the final version, but it's pretty stable anyway and if reviewers are expected to provide another round of modification suggestions, then they are welcome! If it's for final approval, then I'd like to take a couple more days for polishing stuff.

deniederhut:
I understand, but we also want to make sure we are being respectful of our reviewers' time, and the deadline is coming up quickly.

@sosey , do you feel this paper is ready to be included in the proceedings?

sosey:
@deniederhut  I can take one more look this evening and provide any further comments tonight or tomorrow.

sosey:
@deniederhut I think the paper is ready for inclusion

anton-malakhov:
Thank you Megan for the review and nice actionable feedback!
@deniederhut thanks for merging! I have a question, I haven't had a chance to check the PDF layout of the build server with the latest commit and the build server does not have it already. Is there a different link/way how I can check how the paper has been rendered? Thanks

deniederhut:
> I haven't had a chance to check the PDF layout

So this probably should have been done when you pushed your last commit. In any case, you can see the layout [here](http://conference.scipy.org/proceedings/scipy2018/anton_malakhov.html)

PR #388, opened 2018-05-20
Paper: "Sparse: A more modern sparse array library"

hameerabbasi:
Title: Sparse: A more modern sparse array library

Author: Hameer Abbasi

Keywords: sparse, sparse arrays, sparse matrices, scipy.sparse, ndarray, ndarray interface


papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+Sparse arrays are important in many situations and offer both speed and memory benefits
+over regular arrays when solving a broad spectrum of problems. For example, they can be
+used in solving systems of equations :cite:`liu1989limited`, solving partial differential
+equations :cite:`mu1991organization`, machine learning problems involving Bayesian models
+:cite:`tipping2001sparse` and natural language processing :cite:`nickel2011three`.
+
+Traditionally, within the SciPy ecosystem, sparse arrays have been provided within SciPy
+:cite:`scipy` in the submodule :code:`scipy.sparse`, which is arguably the most
+feature-complete implementation of sparse matrices within the ecosystem, providing support
+for basic arithmetic, linear algebra and graph theoretic algorithms.
```
mrocklin, line 53: It might also be worth mentioning that scipy.sparse is also depended upon by many downstream projects.  This dependence adds constraints to how the ecosystem evolves.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+* It doesn't follow the :code:`ndarray` interface (rather, it follows NumPy's deprecated
+  :code:`matrix` interface)
+* It is limited to two dimensions only (even one-dimensional structures aren't supported)
+
+This is important for a number of other packages that are quite innovative, but cannot take
+advantage of :code:`scipy.sparse` for these reasons, because they expect objects following
+the :code:`ndarray` interface. These include packages like Dask :cite:`dask` (which is
+useful for parallel computing, even across clusters, for both NumPy arrays and Pandas
+dataframes) and XArray :cite:`xarray` (which extends Pandas dataframes to multiple
+dimensions).
```
mrocklin, line 72: There are probably other projects here that consume this interface.  It may be worth compiling a list.  I may work on this separately in a bit.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+
+Both of these frameworks could benefit tremendously from sparse structures. In the case of
+Dask, it could be used in combination with sparse structures to scale up computational tasks
+that need sparse structures. In the case of XArray, datasets with large amounts of missing
+data could be represented efficiently.
```
mrocklin: I'm not sure that "missing data" is the right term to use here.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+transpose, reshape and a number of other features. The primary format in this library is based on
+the coordinate format, which stores indices where the array is nonzero, and the corresponding data.
+
+Since a full explanation of usage would be a repeat of the NumPy user manual and the package
+documentation, we only provide brief demos of certain features of interest. Then, we move on
+to some of the design decisions that went into making this package, some optimizations,
+applications and possible future work.
+
+Usage Overview
+--------------
```
mrocklin: Much of this section seems too detailed to be this early in the paper.  I recommend avoiding a ton of discussion on how to use the package and instead focus more on the context of the problem, the challenges of representing sparse arrays and how that affects algorithms, etc..  The details of usage, installation instructions, conversions between different formats, etc. is possibly better saved for documentation.  I recommend taking this paper as an opportunity to speak to a much more informed audience.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+:cite:`sparse`.
+
+Algorithms Used
+---------------
+
+Storage Format
+..............
+
+Although not the most efficient in any respect, for simplicity of operations, we chose the COO
+format for its storage simplicity. In this format, two dense arrays are required to store the
```
mrocklin: I recommend starting off header paragraphs with the topic, rather than with modifiers or disclaimers.  

> We chose the COO format because ..., even though ...

Otherwise the reader doesn't understand how to interpret the text until they're a few phrases in.

Feel free to ignore, just a style suggestion.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+
+   ==== ==== ==== === ====
+   dim1 dim2 dim3 ... data
+   ==== ==== ==== === ====
+      0    0    0 ...   10
+      0    0    3 ...   13
+      0    2    2 ...    9
+    ...  ...  ... ...  ...
+     3    1     4 ...   21
+   ==== ==== ==== === ====
```
mrocklin, line 118: This is an informative table, it might be wise to place it higher in the layout.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+
+Element-wise Operations
+.......................
+
+There was a challenge around designing a function that could perform any kind of element-wise
+operation on any number of variables. The design we decided to adopt at the end was slow compared
+to :code:`scipy.sparse`, but rather fast compared to dense arrays.
```
mrocklin: I recommend first discussing the approach taken, and then discussing the results of taking that approach.  Specifically I recommend waiting to discuss performance until the end of this subsection.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+       data = apply function to data corresponding
+              to these coordinates
+
+       all_coords.append(coords)
+       all_data.append(data)
+
+   concatenate all_coords and all_data
+
+This gets a lot more complex when dealing with broadcasting. Below, we show some simplified
+psuedocode in the broadcasting case. New psuedocode is in parentheses::
```
mrocklin: It might be worth laying out the challenges to implementing elemwise operations at the top of the section.  I might use the following layout:

1.  Elemwise operations are important and common 
2.  Yet they are surprisingly complex for the following reasons ...
3.  We chose to approach this problem as follows...
4.  This was effective but has some drawbacks...
5.  In the future this might be improved in the following ways...

Describing the context of the problem and the challenges that one meets in trying to resolve it are just as important as the implementation or benchmarks.  When reading papers I often find myself appreciating papers more when they teach me about challenges that I didn't know existed.  This is often far more interesting than learning how they solved the problem individually.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+
+Reshaping corresponds to linearizing the coordinates and then doing the reverse for the new shape, similar to
+:code:`np.ravel_multi_index` and :code:`np.unravel_index`. However, we write our own custom implementation for
+this, in order to save on memory.
+
+:code:`dot` and :code:`tensordot`
+.................................
+
+For :code:`tensordot`, we currently just use the NumPy implementation, replacing :code:`np.dot` with
+:code:`scipy.sparse.csr_matrix.dot`.
```
mrocklin: Note that the reader may not know the tricks that numpy uses.  Instead, this might be a good time to inform the user that all tensordot operations can be efficiently reshaped into calls to dot, after which efficient versions can be called (in this case `csr_matrix.dot`)

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+
+For :code:`dot`, we simply dispatch to :code:`tensordot`, providing the appropriate axes.
+
+Optimizations
+-------------
```
mrocklin: Thoughts on transposing the organization so that you discuss elemwise operations all at once, then reductions all at once, etc. rather than going through each twice?  I suspect that this will be easier on the reader.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+to the second if the number of slices are too large.
+
+Benchmarks
+----------
+
+Because of our desire for clean and generic code as well as using mainly pure Python as opposed to
+Cython/C/C++ in most places, our code is not as fast as :code:`scipy.sparse.csr_matrix`. It, however,
+does beat :code:`numpy.ndarray`. The benchmarks were performed on a laptop with a Core i7-3537U
+processor and 16 GB of memory. Any arrays used had a shape of :code:`(10000, 10000)` with a density
+of :code:`0.001`. The results are tabulated in table :ref:`tab:bench`.
```
mrocklin: Including numpy in these results seems odd.  Presumably we can tune the sparsity of the problem to beat dense numpy arrays arbitrarily.

mrocklin: Instead, I recommend comparing sparse to scipy.sprase and, where results are poor, focus instead on why this is the case, either through profiling or by discussing the challenges of general n-dimensional layouts.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,385 @@
...
+   | Sum, Axis=1    | 959 µs ± 23.7 µs  | 641 µs ± 83.9 µs  | 62.7 ms ± 4.86 ms |
+   +----------------+-------------------+-------------------+-------------------+
+
+Outlook and Future Work
+-----------------------
+
+There are a number of areas we would like to focus on in the future. These include, in very broad terms:
+
+* Better performance
+* Better integration with community packages, such as scikit-learn, Dask and XArray
```
mrocklin, line 410: If you're interested, the dask.array stuff could be done today:

```python
In [1]: import dask.array as da

In [2]: x = da.random.random((100000, 100000), chunks=(10000, 10000))

In [3]: y = da.random.random((100000, 100000), chunks=(10000, 10000))

In [4]: x[y < 0.99] = 0

In [5]: import sparse

In [6]: x = x.map_blocks(sparse.COO.from_numpy)

In [7]: x[0].compute()
Out[7]: <COO: shape=(100000,), dtype=float64, nnz=1002>
```

mrocklin (commented):
Nice work!  I've left some comments, mostly high-level.  Feel free to ignore.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: I'm not entirely sure what a better term here would be... Pandas represents missing/invalid data with `NaN` values, so does XArray AFAICT (being an extension of Pandas).

mrocklin: XArray's of sparse data would presumably have many zeros, and not many NaNs.

hameerabbasi: Actually, @shoyer requested that we support `NaN` fill-values as well as zero fill-values (see pydata/sparse#1 and pydata/sparse#143), and it isn't too hard to implement.

It also handles a lot of other cases (such as `y = x + 5` (`y.fillvalue = 5`)).

I haven't mentioned this explicitly since it isn't implemented yet.

mrocklin: Understood.  I expect this to be more of a special case than the standard case. When I read the text above I assumed that you were saying that missing data was the main reason why someone might want to use sparse with xarray, which I don't believe to be the case.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: Is there way to break a given sparse array into chunks? I could include Dask's multithreaded performance benchmarks here as well, although I suspect that for an array this small, it won't make much of a difference.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: Done, a little below where you suggested.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: There are many, I'm just listing the ones most used in the ecosystem.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: I've clarified to say that using Sparse with XArray could have other possible benefits.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: LaTeX places it where it wants in the compiled version, I've placed it close to where it has been referenced.

papers/hameer_abbasi/27_abbasi.rst
mrocklin: Ideally this would work:

```python
In [1]: import sparse

In [2]: import dask.array as da

In [3]: s = sparse.random((100000, 100000), density=0.0001)

In [4]: x = da.from_array(s, chunks=(10000, 10000)).persist()
```

However it looks like this triggers a densification step, which it shouldn't.  That densification step also fails, which is an issue

<details>

```python
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-4-d1e335a35e74> in <module>()
----> 1 x = da.from_array(s, chunks=(10000, 10000)).persist()

~/workspace/dask/dask/base.py in persist(self, **kwargs)
    127         dask.base.persist
    128         """
--> 129         (result,) = persist(self, traverse=False, **kwargs)
    130         return result
    131 

~/workspace/dask/dask/base.py in persist(*args, **kwargs)
    581         postpersists.append((rebuild, a_keys, state))
    582 
--> 583     results = schedule(dsk, keys, **kwargs)
    584     d = dict(zip(keys, results))
    585     results2 = [r({k: d[k] for k in ks}, *s) for r, ks, s in postpersists]

~/workspace/dask/dask/threaded.py in get(dsk, result, cache, num_workers, **kwargs)
     73     results = get_async(pool.apply_async, len(pool._pool), dsk, result,
     74                         cache=cache, get_id=_thread_get_id,
---> 75                         pack_exception=pack_exception, **kwargs)
     76 
     77     # Cleanup pools associated to dead threads

~/workspace/dask/dask/local.py in get_async(apply_async, num_workers, dsk, result, cache, get_id, rerun_exceptions_locally, pack_exception, raise_exception, callbacks, dumps, loads, **kwargs)
    519                         _execute_task(task, data)  # Re-execute locally
    520                     else:
--> 521                         raise_exception(exc, tb)
    522                 res, worker_id = loads(res_info)
    523                 state['cache'][key] = res

~/workspace/dask/dask/compatibility.py in reraise(exc, tb)
     66         if exc.__traceback__ is not tb:
     67             raise exc.with_traceback(tb)
---> 68         raise exc
     69 
     70 else:

~/workspace/dask/dask/local.py in execute_task(key, task_info, dumps, loads, get_id, pack_exception)
    288     try:
    289         task, data = loads(task_info)
--> 290         result = _execute_task(task, data)
    291         id = get_id()
    292         result = dumps((result, id))

~/workspace/dask/dask/local.py in _execute_task(arg, cache, dsk)
    269         func, args = arg[0], arg[1:]
    270         args2 = [_execute_task(a, cache) for a in args]
--> 271         return func(*args2)
    272     elif not ishashable(arg):
    273         return arg

~/workspace/dask/dask/array/core.py in getter(a, b, asarray, lock)
     75         c = a[b]
     76         if asarray:
---> 77             c = np.asarray(c)
     78     finally:
     79         if lock:

~/Software/miniconda/lib/python3.6/site-packages/numpy/core/numeric.py in asarray(a, dtype, order)
    490 
    491     """
--> 492     return array(a, dtype, copy=False, order=order)
    493 
    494 

~/workspace/sparse/sparse/coo/core.py in __array__(self, dtype, **kwargs)
   1119 
   1120     def __array__(self, dtype=None, **kwargs):
-> 1121         x = self.todense()
   1122         if dtype and x.dtype != dtype:
   1123             x = x.astype(dtype)

~/workspace/sparse/sparse/coo/core.py in todense(self)
    322 
    323         if coords != ():
--> 324             x[coords] = data
    325         else:
    326             if len(data) != 0:

IndexError: index 21475 is out of bounds for axis 0 with size 10000
```

</details>

hameerabbasi: [Looks like we'll have to use `da.from_array(..., asarray=False, fancy=False)`](https://dask.pydata.org/en/latest/array-api.html#dask.array.from_array)

I also checked the indexing on a fairly large array to make sure the other data path is triggered and it seems like both paths are fine. I also commented both paths out one by one.

mrocklin: I'm not sure if I find this encouraging, discouraging, or both :)

```python
In [1]: import sparse

In [2]: import dask.array as da

In [3]: s = sparse.random((100000, 100000), density=0.0001)

In [4]: x = da.from_array(s, chunks=(10000, 10000), asarray=False, fancy=False).persist()

In [5]: %%time 
   ...: for i in range(4):
   ...:     s = (s + s.T) * s.sum(axis=0)
   ...: print(s.sum())
   ...: 
2.693194645866806e+23
CPU times: user 10.6 s, sys: 492 ms, total: 11.1 s
Wall time: 11.1 s

In [6]: %%time 
   ...: for i in range(4):
   ...:     x = (x + x.T) * x.sum(axis=0)
   ...: y = x.sum()
   ...: print(y.compute())
   ...: 
1.4251771693961912e+23
CPU times: user 8.63 s, sys: 225 ms, total: 8.86 s
Wall time: 3.71 s
```

Also, this fails:

```python
In [14]: x.dot(x.T).sum().compute()
```

<details>

```python-traceback
/home/mrocklin/workspace/dask/dask/array/core.py:2507: UserWarning: Increasing number of chunks by factor of 10
  (nparts / max_parts))
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-14-9b38e47a300b> in <module>()
----> 1 x.dot(x.T).sum().compute()

~/workspace/dask/dask/base.py in compute(self, **kwargs)
    154         dask.base.compute
    155         """
--> 156         (result,) = compute(self, traverse=False, **kwargs)
    157         return result
    158 

~/workspace/dask/dask/base.py in compute(*args, **kwargs)
    398     keys = [x.__dask_keys__() for x in collections]
    399     postcomputes = [x.__dask_postcompute__() for x in collections]
--> 400     results = schedule(dsk, keys, **kwargs)
    401     return repack([f(r, *a) for r, (f, a) in zip(results, postcomputes)])
    402 

~/workspace/dask/dask/threaded.py in get(dsk, result, cache, num_workers, **kwargs)
     73     results = get_async(pool.apply_async, len(pool._pool), dsk, result,
     74                         cache=cache, get_id=_thread_get_id,
---> 75                         pack_exception=pack_exception, **kwargs)
     76 
     77     # Cleanup pools associated to dead threads

~/workspace/dask/dask/local.py in get_async(apply_async, num_workers, dsk, result, cache, get_id, rerun_exceptions_locally, pack_exception, raise_exception, callbacks, dumps, loads, **kwargs)
    519                         _execute_task(task, data)  # Re-execute locally
    520                     else:
--> 521                         raise_exception(exc, tb)
    522                 res, worker_id = loads(res_info)
    523                 state['cache'][key] = res

~/workspace/dask/dask/compatibility.py in reraise(exc, tb)
     66         if exc.__traceback__ is not tb:
     67             raise exc.with_traceback(tb)
---> 68         raise exc
     69 
     70 else:

~/workspace/dask/dask/local.py in execute_task(key, task_info, dumps, loads, get_id, pack_exception)
    288     try:
    289         task, data = loads(task_info)
--> 290         result = _execute_task(task, data)
    291         id = get_id()
    292         result = dumps((result, id))

~/workspace/dask/dask/local.py in _execute_task(arg, cache, dsk)
    268     elif istask(arg):
    269         func, args = arg[0], arg[1:]
--> 270         args2 = [_execute_task(a, cache) for a in args]
    271         return func(*args2)
    272     elif not ishashable(arg):

~/workspace/dask/dask/local.py in <listcomp>(.0)
    268     elif istask(arg):
    269         func, args = arg[0], arg[1:]
--> 270         args2 = [_execute_task(a, cache) for a in args]
    271         return func(*args2)
    272     elif not ishashable(arg):

~/workspace/dask/dask/local.py in _execute_task(arg, cache, dsk)
    265     """
    266     if isinstance(arg, list):
--> 267         return [_execute_task(a, cache) for a in arg]
    268     elif istask(arg):
    269         func, args = arg[0], arg[1:]

~/workspace/dask/dask/local.py in <listcomp>(.0)
    265     """
    266     if isinstance(arg, list):
--> 267         return [_execute_task(a, cache) for a in arg]
    268     elif istask(arg):
    269         func, args = arg[0], arg[1:]

~/workspace/dask/dask/local.py in _execute_task(arg, cache, dsk)
    269         func, args = arg[0], arg[1:]
    270         args2 = [_execute_task(a, cache) for a in args]
--> 271         return func(*args2)
    272     elif not ishashable(arg):
    273         return arg

~/workspace/dask/dask/compatibility.py in apply(func, args, kwargs)
     47     def apply(func, args, kwargs=None):
     48         if kwargs:
---> 49             return func(*args, **kwargs)
     50         else:
     51             return func(*args)

~/workspace/dask/dask/array/routines.py in _tensordot(a, b, axes)
    189     x = max([a, b], key=lambda x: x.__array_priority__)
    190     tensordot = tensordot_lookup.dispatch(type(x))
--> 191     x = tensordot(a, b, axes=axes)
    192     ind = [slice(None, None)] * x.ndim
    193     for a in sorted(axes[0]):

~/workspace/sparse/sparse/coo/common.py in tensordot(a, b, axes)
    144     at = a.transpose(newaxes_a).reshape(newshape_a)
    145     bt = b.transpose(newaxes_b).reshape(newshape_b)
--> 146     res = _dot(at, bt)
    147     if isinstance(res, scipy.sparse.spmatrix):
    148         if res.nnz > reduce(operator.mul, res.shape) / 2:

~/workspace/sparse/sparse/coo/common.py in _dot(a, b)
    197     if isinstance(b, COO) and not isinstance(a, COO):
    198         return _dot(b.T, a.T).T
--> 199     aa = a.tocsr()
    200 
    201     if isinstance(b, (COO, scipy.sparse.spmatrix)):

~/workspace/sparse/sparse/coo/core.py in tocsr(self)
   1293             self._csr = csr = self._tocsr()
   1294         else:
-> 1295             csr = self._tocsr()
   1296         return csr
   1297 

~/workspace/sparse/sparse/coo/core.py in _tocsr(self)
   1256         # Pass 3: count nonzeros in each row
   1257         indptr = np.zeros(self.shape[0] + 1, dtype=np.int64)
-> 1258         np.cumsum(np.bincount(row, minlength=self.shape[0]), out=indptr[1:])
   1259 
   1260         return scipy.sparse.csr_matrix((self.data, col, indptr), shape=self.shape)

~/Software/anaconda/lib/python3.6/site-packages/numpy/core/fromnumeric.py in cumsum(a, axis, dtype, out)
   2163 
   2164     """
-> 2165     return _wrapfunc(a, 'cumsum', axis=axis, dtype=dtype, out=out)
   2166 
   2167 

~/Software/anaconda/lib/python3.6/site-packages/numpy/core/fromnumeric.py in _wrapfunc(obj, method, *args, **kwds)
     50 def _wrapfunc(obj, method, *args, **kwds):
     51     try:
---> 52         return getattr(obj, method)(*args, **kwds)
     53 
     54     # An AttributeError occurs if the object does not have

ValueError: provided out is the wrong size for the reduction
```

```python
In [15]: debug
ipdb> up
> /home/mrocklin/Software/anaconda/lib/python3.6/site-packages/numpy/core/fromnumeric.py(2165)cumsum()
   2163 
   2164     """
-> 2165     return _wrapfunc(a, 'cumsum', axis=axis, dtype=dtype, out=out)
   2166 
   2167 

ipdb> 
> /home/mrocklin/workspace/sparse/sparse/coo/core.py(1258)_tocsr()
   1256         # Pass 3: count nonzeros in each row
   1257         indptr = np.zeros(self.shape[0] + 1, dtype=np.int64)
-> 1258         np.cumsum(np.bincount(row, minlength=self.shape[0]), out=indptr[1:])
   1259 
   1260         return scipy.sparse.csr_matrix((self.data, col, indptr), shape=self.shape)

ipdb> pp indptr.shape
(10001,)
ipdb> pp np.bincount(row, minlength=self.shape[0]).shape
(42950,)
```

</details>

mrocklin: Interstingly both operations work fine with smaller data around with size 30000x30000 or smaller with  the same sparsity.

hameerabbasi: It seems that our design choice of having the smallest possible dtype for `coords` is causing most of these issues. I've fixed it over in pydata/sparse#158. :-)

hameerabbasi:
@deniederhut Is there something I can do to get more reviewers aboard? I'm willing to work with everyone to fix things. 😄

mrocklin:
If he has time @jcrist might enjoy reviewing this

deniederhut:
Sorry 😬 we had a bit more reviewer dropout than we were expecting. I'll
try to take a look tonight.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+In addition, :code:`scipy.sparse` is depended on by many downstream projects, which makes
+removing NumPy's :code:`matrix` interface that much more difficult, and limits usage of
+both :code:`ndarray` style duck arrays and :code:`scipy.sparse` arrays within the same
+codebase.
+
+This is important for a number of other packages that are quite innovative, but cannot take
+advantage of :code:`scipy.sparse` for these reasons, because they expect objects following
+the :code:`ndarray` interface. These include packages like Dask :cite:`dask` (which is
+useful for parallel computing, even across clusters, for both NumPy arrays and Pandas
+dataframes) and XArray :cite:`xarray` (which extends Pandas dataframes to multiple
```
deniederhut, line 71: Does Ray belong in this list as well?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+   dim1 dim2 dim3 ... data
+   ==== ==== ==== === ====
+      0    0    0 ...   10
+      0    0    3 ...   13
+      0    2    2 ...    9
+    ...  ...  ... ...  ...
+     3    1     4 ...   21
+   ==== ==== ==== === ====
+
+To save on memory, we always choose the smallest possible data type for the coordinates array.
```
deniederhut: If I understand the comments correctly, this is no longer true?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+
+Element-wise operations
+.......................
+
+Element-wise operations are an important and common part of any array interface. For example,
+arithmetic, casting an array, and all NumPy :code:`ufunc` s are common examples of element-wise
+operations.
+
+These turn out to be simple for NumPy arrays, but are surprisingly complex for sparse arrays.
+The first problem to overcome was the lack of dependency on Numba :cite:`numba`/Cython
```
deniederhut: I'm not sure what you mean here. Why is a lack of dependency a problem?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+Consider two arrays, one shaped :code:`(n,)` and another shaped :code:`(m, n)`, both with only
+one nonzero entry. If all we wanted to do was multiply them, the result would have just one
+nonzero entry, yet broadcasting the first array would result in an array with :code:`m` nonzero
+entries (which clearly isn't the most optimal way to do things). For this reason, we chose to
+handle broadcasting within the algorithm itself, instead of broadcasting all inputs upfront.
+
+Effectively, this resulted in the following algorithm, which doesn't have the limitation mentioned
+above. This is because any zeros are filtered out before any broadcasting is done::
+
+   all_coords = []
```
deniederhut, line 180: I'm not sure I understand why you have chosen to go with pseudo code here. Isn't this implemented somewhere in python?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+
+   concatenate all_coords and all_data
+
+While this algorithm is effective at applying all sorts of element-wise operations for
+any amount of inputs, it does have a few drawbacks:
+
+* It's slower than :code:`scipy.sparse`, because
+
+  * It loops over all possible combinations of zero/nonzero
+    coordinates, which makes it :math:`O \left( \left(2^\text{nin} - 1 \right) \times \text{nnz} \right)`
```
deniederhut, line 215: Did we define what these abbreviations stand for? E.g. number nonzero?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+
+Only some reductions are possible with this algorithm at the moment, but most common ones are supported.
+Supported reductions must have a few properties:
+
+* They must be implemented in the form of :code:`ufunc.reduce`
+* The :code:`ufunc` must be reorderable
+* Reducing by multiple zeros shouldn't change the result
+* An all-zero reduction must produce a zero.
+
+Although these restrictions may seem crippling, in practice most reductions such as :code:`sum`,
```
deniederhut: There might be a better way to phrase this. Maybe "Although these criteria seem restricting"? Or "prohibitive"?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+:code:`prod`, :code:`min`, :code:`max`, :code:`any` and :code:`all` actually fall within the class
+of supported reductions. We used :code:`__array_ufunc__` protocol to allow application of :code:`ufunc`
+reductions to COO arrays. Notable unsupported reductions are :code:`argmin` and :code:`argmax`.
```
deniederhut: Is there a reason for this?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+does beat :code:`numpy.ndarray`, provided the sparsity of the array is small enough. The benchmarks
+were performed on a laptop with a Core i7-3537U processor and 16 GB of memory. Any arrays used had a
+shape of :code:`(10000, 10000)` with a density of :code:`0.001`. The results are tabulated in table
+:ref:`tab:bench`.
+
+The NumPy results are given only for comparison, and for the purposes of illustrating that using sparse
+arrays does, indeed, have benefits over using dense arrays when the density of the sparse array is
+sufficiently low.
+
+.. table:: Benchmarks :label:`tab:bench`
```
deniederhut: I think you're missing a caption here

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,351 @@
...
+    ...  ...  ... ...  ...
+     3    1     4 ...   21
+   ==== ==== ==== === ====
+
+To save on memory, we always choose the smallest possible data type for the coordinates array.
+
+Element-wise operations
+.......................
+
+Element-wise operations are an important and common part of any array interface. For example,
```
deniederhut, line 129: Are you planning on including a discussion of the dense result error that might get thrown?

deniederhut (changes requested):

deniederhut:
So I took a quick look through this, and I'm super happy to see an n-dimensional sparse data library! I was a little surprised that this wasn't implemented in C++ or Cython though, and I'm not sure I understood from the paper why this design choice was made. 

I'm not in love with the ValueErrors for operations that might return dense arrays -- is the idea that every operation gets wrapped in a try/catch? They show up in some places I wouldn't expect, like mutiplying a sparse array of zeros by a dense array of zeros:

```python
sparse.COO(np.zeros((5, 2))) * np.zeros((5, 2))
```
raises:
```
ValueError: Performing this operation would produce a dense result: <ufunc 'multiply'>
```

I think comprehensibility could be helped with a motivating example toward the beginning (e.g. showing a slow operation with sparse data using numpy), and maybe some more exposition on the tradeoffs of memory use, indexing, and common linear algebra tasks.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: From what I could tell from a quick glance at the docs, no. It (maybe?) supports Pandas, but I couldn't find anything on NumPy. Compiling an extensive list of NumPy consumers is a huge task, I just noted a few I hope to integrate with in the near future.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: Yes, I wasn't expecting this to change when I wrote this. I'll amend this.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: When I wrote this, I was still waiting for things to settle out on which of these to add (see pydata/sparse#126). I wasn't sure if adding detailed background on design decisions is appropriate here. In any case, I've amended the text to say: "The first problem to overcome was that there was no dependency on Numba :cite:`numba`/Cython
:cite:`cython`/C++ at the time that this algorithm was to be implemented. I, therefore wished to solve the problem in pure NumPy, ..."

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: It is, but the actual implementation has many gotchas and corner cases, and is a bit too long to include. I could link to the current full implementation here. Are links to library code appropriate?

hameerabbasi:
I have two questions in general about your comments:

1. Is it appropriate to link to actual implementations in the library, for the reason that the code is too large to fit in the paper or handles corner cases?
2. Is it appropriate to add speculation on what might or might not be supported in the future?

hameerabbasi:
In general, for that `ValueError`, we're aware of it and we're coming up with ways to reduce it. For example, multiplication produces a sparse array if either operand is sparse. We're trying to automatically handle this; see pydata/sparse#124.

Another thing that might help are arbitrary fill values for things like `y = x + 5` (if `x.fill_value == 0`, `y.fill_value == 5`). See pydata/sparse#143.

In general, though; if the result is dense, we raise an error. The idea is that it only gets raised on certain operations or types. E.g. `dense + sparse`, etc. The idea is that we don't want to allow performance degradation so we raise an error. If the user wants to do an operation with dense results, we require the user to be explicit about it instead of implicitly degrading performance.

I can put all of this into a discussion section, but it relies partially to an answer to 2 above (some of it is speculation about future features that will hopefully exist).

About C++/Cython, see pydata/sparse#126.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: I just did. :-)

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: See comment down below.

hameerabbasi:
To speed things along, I did things the way I thought you meant (see commit messages).

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: Done.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: `argmin` and `argmax` can be implemented with a certain algorithm, but it isn't implemented yet. The generic version only supports `ufunc.reduce`-style reductions, which is the reason here. I've modified the paper to say that.

papers/hameer_abbasi/27_abbasi.rst
hameerabbasi: Added.

deniederhut:
Speculation is okay, as long it is marked as such, e.g. "we anticipate" or "we hope to", or "we expect". As for linking to source code, yes that's fine, but I think I was imagining a version of the source that had been simplified for space/clarity. I haven't dug through the actual code base you are presenting though, so that might more work that it's worth.

Feel free to @ me when you're ready for re-review.

hameerabbasi:
@deniederhut I just gave it a read. The current version is review-ready and hopefully addresses all of your comments.

If the review passes, feel free to merge from my side.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,410 @@
...
+arithmetic, casting an array, and all NumPy :code:`ufunc` s are common examples of element-wise
+operations.
+
+These turn out to be simple for NumPy arrays, but are surprisingly complex for sparse arrays.
+The first problem to overcome was that there was no dependency on Numba :cite:`numba`/Cython
```
deniederhut, line 134: It's not really clear to me what the intended meaning is here. My guess is that, for a proof of concept, you didn't want the implementation to have complex dependencies or compilation requirements. Is that right? Could we change the wording to make your intentions more clear?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,410 @@
...
+
+* Looping over possibly nonzero coordinates with something like Numba or Cython.
+
+  * This approach will solve most of the speed issues.
+  * Sorting will be rendered unnecessary.
+  * Specialized code paths introduce a large maintenance burden, but can be implemented.
+
+* Introducing multidimensional CSR/CSC.
+
+You can see the current performance of the code in table :ref:`tab:bench`.
```
deniederhut: I don't know that we have an "official" style guide for SciPy, but usually the _T_ in Table and _F_ in Figure are capitalized.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,410 @@
...
+planned to allow mixed :code:`ndarray` -sparse operations if such operations do not produce dense
+results e.g. multiplication. :cite:`sparse-elemwise-ndarray`. Also, we are planning to allow arbitrary
+fill values in arrays, which will allow for operations such as :code:`y = x + 5` (if :code:`x.fill_value`
+was zero, :code:`y.fill_value` will be five). :cite:`sparse-arbitrary-fill-value`
+
+Reductions
+..........
+
+We implemented reductions by the simple concept of a "grouped reduce". The idea is to first group the
+coordinates by the non-selected axes, and then reduce along the selected axes. This is simple to
```
deniederhut: Grouped reductions are described as "simple" twice in a row here.

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,410 @@
...
+
+For other cases, it's wise to initially use the second procedure (to filter out some sub-arrays), and then
+switch to the first. For example, for :code:`x[:500, :500, :500]`, as using just the second procedure will
+require a large amount of binary searches (:math:`500^3` in this case).
+
+So we used a hybrid approach where the second method is used until there are a sufficiently low
+number of coordinates left for filtering, then we fall back to simple filtering. Where we do the
+switch is determined by a heuristic: will the expected number of binary searches be faster in a
+specific case, or directly filtering the number of left-over coordinates? The overall algorithm
+is implemented in Numba.
```
deniederhut: I thought we had said above that this library did not want to depend on numba?

papers/hameer_abbasi/27_abbasi.rst
```diff
@@ -0,0 +1,410 @@
...
+.........................
+
+Transposing corresponds to a simple reordering of the dimensions in the coordinates, along with a re-sorting
+of the coordinates and data to make the coordinates C-contiguous again.
+
+Reshaping corresponds to linearizing the coordinates and then doing the reverse for the new shape, similar to
+:code:`np.ravel_multi_index` and :code:`np.unravel_index`. However, we write our own custom implementation for
+this.
+
+:code:`dot` and :code:`tensordot`
```
deniederhut: The fixed width font makes it harder to see that this is a subheading (I initially thought this was a sentence fragment accidentally left undeleted). Would it be possible to remove the two `:code:` directives here?

deniederhut:
This is excellent! I've pointed out a few minor typographical issues in the edits. My ony remaining concern is that I'm still not understanding the part about not depending on Numba/Cython for the library. Once that paragraph is clarified, I think this paper is good to go!

hameerabbasi:
@deniederhut Sorry for the late edits, I was traveling from Friday to Monday. I've tried to clarify the timeline on why and when Numba was added, and why it's used in some features but not others, including citing the relevant issue. I've also made the typographical edits you suggested.

hameerabbasi:
This should be ready for review. Feel free to merge if it passes.

deniederhut:
Thanks for your edits on this! @proceedings I think we're good to go!

deniederhut (approved):

hameerabbasi:
Thanks so much @deniederhut for reviewing this paper! It means a lot to me that you personally stepped in in the absence of reviewers.

PR #390, opened 2018-05-20
Fix order of commands.

mkcor:
Hello!

We need to cd into the repo before running subsequent git commands.

See you at the conference :)


mpacer:
good catch! Thanks for contributing! 

BTW, future changes to the system outside of a paper submission should be made against the dev branch not the 2018 branch. (not a big deal in this case though)

mkcor:
Oops, good to know @mpacer!

PR #386, opened 2018-05-18
Paper: [WIP] Project Jupyter - Binder 2.0

choldgraf:
Build server here: http://procbuild.scipy.org/

Note: open review goes until July 2nd.

PR #393, opened 2018-05-20
adds explanation of the PR name convention used by the build server

mpacer:
This is a simple fix to the problem that came up with #386 

Does this do enough to explain what's happening?

cc: @choldgraf, @deniederhut @dalippa, @FatihAkici


deniederhut:
👏 👏 👏 Thanks M!

PR #395, opened 2018-05-21
Paper: Bringing ipywidgets support to plotly.py

jonmmease:
Hello reviewers,
     Here is the first draft of my SciPy 2018 paper: Bringing ipywidgets support to plotly.py
Thanks!

[Mease_SciPy18.pdf](https://github.com/scipy-conference/scipy_proceedings/files/2039528/Mease_SciPy18.pdf)


jasongrout:
Very nice!

A few comments from the ipywidgets perspective:

1. Widget models and views actually only need to inherit from [ipywidgets.Widget](https://github.com/jupyter-widgets/ipywidgets/blob/766cad54a47c07520e9d695534c4664c3391e7ec/ipywidgets/widgets/widget.py#L286), or on the js side the [WidgetModel](https://github.com/jupyter-widgets/ipywidgets/blob/766cad54a47c07520e9d695534c4664c3391e7ec/packages/base/src/widget.ts#L71) or [WidgetView](https://github.com/jupyter-widgets/ipywidgets/blob/766cad54a47c07520e9d695534c4664c3391e7ec/packages/base/src/widget.ts#L587) classes. The DOM classes are for widgets that have an actual DOM element representation, but some widgets like many of the ones in pythreejs do not have a DOM element associated with them (they represent three.js objects in pythreejs, for example: [here](https://github.com/jupyter-widgets/pythreejs/blob/be116d1686d3bd971be481c1555b18a1e0a6701e/pythreejs/_base/Three.py#L8), [here](https://github.com/jupyter-widgets/pythreejs/blob/be116d1686d3bd971be481c1555b18a1e0a6701e/js/src/_base/Three.js#L84), etc.)

2. The paper conflates the comm layer and the syncing message protocol that ipywidgets uses in its communication. The comms infrastructure is a Jupyter infrastructure for creating an object in the kernel and an object in the frontend that have a communication channel. What comms abstract is creating two objects, and sending arbitrary messages between them (it does not specify a message format). What ipywidgets layers on top of comms is a state syncing protocol (i.e., serializing attributes into specific message types sent over the comm channel for syncing, as well as a message format for custom messages). This message protocol is detailed in https://github.com/jupyter-widgets/ipywidgets/blob/master/packages/schema/messages.md#widget-messaging-protocol-version-2

jonmmease:
@jasongrout Thanks for the helpful comments and clarifications.  I was a bit foggy on the relationship between the comms layer and the syncing protocol, and your explanation here helps a lot. I'll go back through and work on de-conflating them!

jonmmease:
@jasongrout I think these latest updates take both of your comments into account. Thanks again for taking a look!

rgbkrk (approved):

chandanbose:
I find this work very interesting!
@proceedings, The paper is ready to be merged.

rgbkrk:
Congrats!

PR #403, opened 2018-05-23
Paper: Safe handling instructions for missing data

deniederhut:


msarahan:
This is a really good overview of the subject with a lot of helpful links.  It would be more relatable if you could point at real-world example datasets (or just mention real-world fields) that demonstrate the missing data pathologies that people may encounter, and thus where they should start looking closer at your sources on what to do next.

ghost:
well thought paper. As like previous reviewer, I felt the similar probably requires real world data to signify the importance and provide the value to the researchers.

msarahan:
Nice.  I like the addition.  It's a good discussion of why a given categorization of missing data would apply.

I vote that this paper is ready.

deniederhut:
Thanks for the feedback! Addressing sources of missingness is a bit outside the scope of the paper due to how much they vary field by field. For medical studies it can be the death of the subject; for epidemiology it can be participant dropout; for NMR it can be cell phone interference; and so on.

I have added a case study with a real dataset (albeit a whimsical one) that I think addresses the concerns about relatability. Please let me know if there are other changes you would like to see!

dalippa:
@dataneupane Please let us know your name and if you are attending the conference so you can get your reviewer sticker.

PR #409, opened 2018-05-23
Add decode to build_paper output to make it format better

ksunden:
Was printing Byte-string, which was horrible to read

### DIFF
--- a/publisher/build_paper.py
+++ b/publisher/build_paper.py
@@ -149,7 +149,7 @@ def tex2pdf_singlepass(out_path):
     if b"Fatal" in out or run.returncode:
         print("PDFLaTeX error output:")
         print("=" * 80)
-        print(out)
+        print(out.decode())
         print("=" * 80)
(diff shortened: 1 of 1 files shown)


publisher/build_paper.py
```diff
@@ -149,7 +149,7 @@ def tex2pdf_singlepass(out_path):
     if b"Fatal" in out or run.returncode:
         print("PDFLaTeX error output:")
         print("=" * 80)
-        print(out)
+        print(out.decode())
```
mpacer: My intuition is that we should decode it to `'utf-8'` for consistency with Python3 (which is currently planned to be the only supported python as of next year). Is there a reason not to do that?

mpacer:
🎉 For making this PR against `dev`. You rock!

deniederhut:
I would like to second to motion for you rocking.

deniederhut:
1pfr

PR #413, opened 2018-05-24
Paper: A Bayesian’s journey to a better research workflow

mkcor:
Dear reviewers,

Here is our contributed paper.

We are aware of some formatting issues in code and code-like blocks. We would welcome your help and suggestions in this regard. 

Thank you!

Konstantinos (@bayesways) and Marianne (@mkcor)


alberto-antonietti:
Dear @mkcor and @bayesways,
I am currently reading and reviewing your paper, I will come back to you very soon with my comments! ;-)

junipertcy:
Dear @mkcor and @bayesways, I am reviewing the paper, too. Will come back soon with my comments.

aweinstein:
@mkcor and @bayesways: The following are my comments about your paper.

The paper describes a workflow for performing Bayesian analysis. The authors illustrate the use of this workflow through a clinical trial analysis.

The paper is well written and clear. The following are some minor comments:

- In sec. 2.c, give a reference for the LKJ distribution. I think this distribution is not well know for people (like me!) outside the field.

- Fix the formatting issues: code of Sec. 4 overrun over the right column. The output of the Stan `fit` object overrun over Fig. 3, and in the right column, the table is cut.

- I wonder if showing the table with all these number add any value to the paper. Without an explanation of what they mean, or how they are useful, I suggest to remove the table.

- Increase the size of the legend of Fig. 4. As it is, I need to zoom in a lot to see it.

- You mention a Jupyter notebook with all the analysis, but I can't find a link to it. Please provide a link.

alberto-antonietti:
Dear @mkcor and @bayesways,

I have really appreciated your paper "A Bayesian's journey to a better research workflow", even if I am not a (Bayesian) statistician, you provide the reader with the general background necessary to understand it. The approach that you propose is generalizable to a wealth of research activities and it could lead to a paradigm shift in statistical reports found in current papers and clinical trial reports.

A couple of general comments:

1. The paper alternates parts more easy to follow for non-experts in the field with some technicalities that are difficult to grasp. For example, while the concept of the choice of the priors is understandable (Pag. 2), the values for the example (Pag. 4) are chosen without an explanation. I suggest adding some more details in order to help non-experts to easily follow the whole manuscript.

2. Since your workflow can be split into clear different actions, it could be useful to split also the code and the figure of Pag. 5 between the different steps. In addition, it would be very instructive to have a sort of live demo of your code that the user can try while he/she is reading the paper. (e.g. https://codeocean.com/).

Some other comments:

1. Pag. 4, Table 1, Last Row the Subject ID is the same as the previous line
2. Pag. 4 last row. This is the first time that you mention the Stan program, it should be introduced before.
3. Pag. 5 Various formatting issues as already pointed out by @aweinstein. I also agree with his comment about the usefulness of the output table.
4. Pag. 5. Fig. 1, 2 and 3. I suggest adding more information to the labels of the figures, in order to make them self-explainable. In addition, it is not clear to me the meaning of the color code (red vs orange) and of the black vertical lines. 
5. Pag. 6 HMC acronym is here used, but not defined (it is defined in Pag. 7).
6. Pag. 6 "Note how the Bayesian approach results in predictive distributions rather than point estimates...". Is it possible to extract also quantitative indexes and clinical questions (e.g. what is the probability _p_ that a patient's hemoglobin level will reach a certain value with and without the treatment?)?
7. Pag. 7. "automatic differentiation whereas Stan (relies?) on its own algorithm". Possible missing word.
8. Pag. 7 I would put the links for PyMC3 and Stan comparison as footnotes or as references.
9. Reproducibility section. There are various definitions of Reproducibility/Replicability/etc. I would follow the more rigorous approach highlighted in [this paper](https://www.frontiersin.org/articles/10.3389/fninf.2017.00076/full). So, if you want to obtain the same results with the same data and methods I would say "replicability" instead of "reproducibility".
10. Pag. 7 ": By reducing" -> ". By reducing"
11. Pag 8. Other possible solutions could be cloud-based platforms like CodeOcean or the embedding of the necessary software in a Docker container.
12: Pag. 8. I assume that it is possible (maybe preferred) to publicly share data coming from clinical trials even with human subjects, after the mandatory anonymization.

I hope that my comments should be useful to improve your paper and I apologize if my ignorance of the topic made me raise inappropriate ones.

Congratulations again on your fruitful collaboration that made possible this shared effort!

junipertcy:
Dear @mkcor and @bayesways,

Your contribution brings an actionable (and reproducible 😍) workflow to a Bayesian inference task. Here, I comment on the `A Bayesian Workflow` section in hope to clarify the working logic of the workflow.

The goal of a successful Bayesian inference is to obtain the posterior distribution of the model parameters, which either (1) maximizes the posterior probability; or (2) achieves the highest predictive performance. The posterior distribution can help us understand the underlying structure of the data. The two principled criteria are not equivalent unless we correctly capture the underlying generative process of the dataset (this means the model is _consistent_ to the data), often in the case of synthetic datasets. Empirically, we strive to find a consistent model that fits the two criteria in practice.

With that said, some components in the submission are potentially vague, if not misleading.

#### In 1) Scope the problem
* Try to state Bayes' rule in this subsection.
* First paragraph: correctly retrieving samples from the posterior distribution is the _predictive_ goal of Bayesian inference (i.e. one of the 2 goals as I just suggested). It might be nice to provide more background information. 
* Last paragraph: a good prediction replies on a successful inference of the parameter values (i.e. _maximum a posterior_ inference), provided that a good model was chosen. Hence, model selection is important. This should be addressed.

#### In 2) Specify the likelihood and priors
* It is good to include *inference algorithm* as one important ingredient to specify in this step of the workflow. Model, prior, and algorithm are three different things. 
* Second paragraph; saying that "information flow from the data to the parameters": I liked this expression. If not Bayes' rule is introduced till here, this may be an ideal time to state it.
* Without more citations, I think it is assertive to say that uninformative priors are bad choices. Sometimes a Bayesian just want to be agonistic about the data generative process.
* Uninformative and vague priors are not inappropriate for model selection. In model selection, we can compute the _posterior odds ratio_. The _can_, without any inappropriateness, compare models with these priors.
* The need to choose priors is a common critic to Bayesian; however, it is misguided because alternative maximum likelihood approaches are simply equivalent to: (i) assuming that the priors are uniform; (ii) approximating the integral over the parameters as `f(D | M) ≈ f(D | \theta*, M)`, where `\theta*` is the maximum likelihood estimator of M’s parameters. The integral is the marginal likelihood, which is obtained by marginalizing over plausible parameter values (i.e. `f(\theta | M)`).

#### In 3) Generative fake data
* I suggest replacing *fake* with *synthetic*.
* Last paragraph: when the authors say, "all models are wrong, but some models are useful," they should explain when the users are allowed to make a statement that the model `M` is _close enough_ to the real generative process. Without that said, it is vague.

#### In 4) Fit the model to the fake data
* First paragraph: The paper is about a full probabilistic approach (even parameters is a distribution) to learning from data. Hence, the maximum likelihood estimation (MLE) is misleading. 
* Last sentence in the second paragraph - we settle for an algorithm that returns samples from the posterior distribution: Are we settle to a _model_ or an _algorithm_? Again, introducing the concept of  *inference algorithm* in the second subsection may avoid such confusion.
* The reader lacks the definition of a posterior confidence interval.
* The sub-subsection *b*: the whole paragraph may be redundant. Here, the authors discuss how model critique works in practice (i.e. with real data, not synthetic data). However, it is in the later stage of the workflow. In this subsection, we only concern about the sensitivity of the model and inference algorithm to the synthetic dataset.

#### In 5) Fit the model to the real data
* The reader lacks an understanding of when he/she should be _satistied_ with the fit of the model.
* Last paragraph: when the authors suggest "A model that predicts well is a model that fits the data well." However, this is misleading. It is only correct if the model is consistent with the data. In other words, It is possible to find a model that doesn't fit the data so well but still predicts well (low generalization error).
* In `Further reading`, the reader wants to know what a utility function is. This term first appears in the 9th step in the advanced workflow.

That's it. I hope my comments help. Looking forward to your responses.

bayesways:
Dear reviewers, 

Thank you very much for your comments and feedback. We will study them carefully and get back to you soon. 

Konstantinos

deniederhut:
Hey @mkcor ! How are the revisions coming along?

mkcor:
Hi @deniederhut and others!

We know that July 2nd is coming up quickly...  Thanks for all the feedback. We'll push revisions and address points that were raised by the end of the weekend.

Thanks for your patience!

mkcor:
Very minor comment.
@alberto-antonietti "Pag. 7 ": By reducing" -> ". By reducing" (point 10 you raised) is intentional. I tried googling quickly for a reference and found https://www.happywebdiva.com/2015/12/02/strunk-white-colons-web/.

mkcor:
@alberto-antonietti Your point 5: HMC acronym is actually properly introduced page 5, so I removed its 're-introduction' page 7 (via https://github.com/mkcor/scipy_proceedings/commit/9cdfdae2c7301c939826136494c2159fb9f460f1).
By the way, your comments are extremely useful, helpful, and valuable!!

mkcor:
@junipertcy Thank you for the detailed comments!

We agree that 'synthetic data' is a better term than 'fake data' (changed via https://github.com/mkcor/scipy_proceedings/commit/9cdfdae2c7301c939826136494c2159fb9f460f1).

> **In 4) Fit the model to the fake data**
> First paragraph: The paper is about a full probabilistic approach (even parameters is a distribution) to learning from data. Hence, the maximum likelihood estimation (MLE) is misleading.

MLE can be used is various statistical analysis contexts, whether probabilistic or not. So there is no contradiction... We mention it in a Bayesian context and we make this very clear and explicit--don't we?

gdesjardins:
Hi there,

I'm not too sure why I'm on this CL, but I would wager you have the wrong
Guillaume Desjardins and/or email address :). Best of luck with the paper ;)

mkcor:
Sorry @gdesjardins this was an autocompletion problem, I meant to tag @deniederhut (now edited) -- do you mind deleting your comment? I will remove this one too, the thread being long enough as it is.

mkcor:
Hi @deniederhut,

We are working on the various issues that were raised, but we won't be able to address them all before tomorrow (July 2). Is it possible to ask for a little extension at all?

Thank you!
/cc @bayesways

mkcor:
Dear @alberto-antonietti ,

I'm confused about your 9th point:

> Reproducibility section. There are various definitions of Reproducibility/Replicability/etc. I would follow the more rigorous approach highlighted in this paper. So, if you want to obtain the same results with the same data and methods I would say "replicability" instead of "reproducibility".

We *are* taking the definition of the reference you pointed to (https://www.frontiersin.org/articles/10.3389/fninf.2017.00076/full); we happen to mention another reference (https://www.practicereproducibleresearch.org/core-chapters/2-assessment.html) but you can see that the "references therein" are the same and the communities are the same... Well, to get back to the point, the definition of "reproducibility" is the same.

Unlike what you report, these references use "reproducibility" to mean that results are recovered by using the *same* datasets/methods, and "replicability" when using other/new datasets/methods.

alberto-antonietti:
Dear @mkcor,

> Very minor comment.
> @alberto-antonietti "Pag. 7 ": By reducing" -> ". By reducing" (point 10 you raised) is intentional. I tried googling quickly for a reference and found https://www.happywebdiva.com/2015/12/02/strunk-white-colons-web/.

OK, thanks for the clarification. If you prefer using colons, I think that "by" should then have a lower-case "B", right?

> @alberto-antonietti Your point 5: HMC acronym is actually properly introduced page 5, so I removed its 're-introduction' page 7 (via mkcor/scipy_proceedings@9cdfdae).

Perfect.

> By the way, your comments are extremely useful, helpful, and valuable!!

Thank you! :-) Glad to contribute to your effort!

> We are taking the definition of the reference you pointed to (https://www.frontiersin.org/articles/10.3389/fninf.2017.00076/full); we happen to mention another reference (https://www.practicereproducibleresearch.org/core-chapters/2-assessment.html) but you can see that the "references therein" are the same and the communities are the same... Well, to get back to the point, the definition of "reproducibility" is the same.
> 
> Unlike what you report, these references use "reproducibility" to mean that results are recovered by using the same datasets/methods, and "replicability" when using other/new datasets/methods.
> 

You are right, excuse me for the misleading comment!

deniederhut:
Howdy! The proceedings committee is in discussions about this, but right now it looks pretty likely that the deadline will be pushed back toward the end of this week.

mkcor:
@alberto-antonietti Hai un indirizzo email? Così ti parlo di tipografia americana in privato. ;)

alberto-antonietti:
@mkcor nome.cognome@polimi.it :-)

mkcor:
Dear @junipertcy,

> **In 1) Scope the problem**
>    Try to state Bayes' rule in this subsection.

We disagree. Stating Bayes' rule right at the start and without any context would give the misleading impression that Bayesian modeling is about Bayes' rule. This is a common misconception which we actually want to dispel (as a key take-away from this paper and our upcoming presentation).

Bayes' rule holds trivially from probability theory. Just using it does not make a modeling approach Bayesian. What makes Bayesian modeling Bayesian is the probabilistic description of quantities of interest, offering uncertainty quantification for model parameters and predictions.

/cc @bayesways

bayesways:
Dear @junipertcy, 

Thank you for your detailed comments on the Bayesian workflow section. I try to address your points below. I agree with some and disagree with others but I found all of them very useful, so thank you for taking time to write them. I hope my answers are helpful. There are a few outstanding edits to the text that still need to be done by the end of the review deadline. 

--

“The goal of a successful Bayesian inference is to obtain the posterior distribution of the model parameters, which either (1) maximizes the posterior probability; or (2) achieves the highest predictive performance. The posterior distribution can help us understand the underlying structure of the data. “

Not accurate. Bayesian inference, and Bayesian modeling in general, can serve many different goals depending on the research project. Broadly speaking Bayesian modeling aims at modeling the joint distribution of all the parameters. The two goals stated here are more appropriate for a machine learning task, not for Bayesian modeling. For example Point 1) misses a common goal of Bayesian inference of quantifying the uncertainty around parameters. As for Point 2), although prediction is a common task it is not the only one. For example often times scientists would compare how different models fit the data in an effort to validate a certain scientific theory, without necessarily caring about the prediction error.

In 1) Scope the problem
“Try to state Bayes' rule in this subsection.” - stating the rule here out of any context will give the (misleading) impression that Bayesian modeling is centered around the use of Bayes Rule. Just using Bayes Rule does not make a model Bayesian. 
“First paragraph: correctly retrieving samples from the posterior distribution is the predictive goal of Bayesian inference (i.e. one of the 2 goals as I just suggested). It might be nice to provide more background information.”
 - Addressed in other comments

“Last paragraph: a good prediction replies on a successful inference of the parameter values (i.e. maximum a posterior inference), provided that a good model was chosen. Hence, model selection is important. This should be addressed.”
 - This is a valid point and is true for prediction, as well as the rest of the questions stated here. We intentionally avoid discussing this point here for pedagogical reasons. Instead we introduce the concept of "designing the model" in the sentence that follows directly below.  However we appreciate that the reviewer wishes to stress the fact that any inference is dependent on the model chosen, so we add a footnote in the next section to ensure that this point is not lost on the reader. 

In 2) Specify the likelihood and priors
“It is good to include inference algorithm as one important ingredient to specify in this step of the workflow. Model, prior, and algorithm are three different things.”
- This paragraph refers to the things that the modeler chooses, these are the likelihood and the prior.

“Second paragraph; saying that "information flow from the data to the parameters": I liked this expression. If not Bayes' rule is introduced till here, this may be an ideal time to state it.”
 - Good idea, we added it 

“Without more citations, I think it is assertive to say that uninformative priors are bad choices. Sometimes a Bayesian just want to be agonistic about the data generative process.”
 - The reviewer seems to not be aware of the Bayesian literature on this topic. There are two important points here:
i) Uninformative or vague priors are not exactly "agnostic". In fact such priors introduce all sorts of "prior knowledge" in the model, often times without the knowledge of the modeler, especially in higher dimensions.
ii) Uninformative priors can be used for characterizing the posterior distribution of the parameters, but their use is very restricted in other tasks, notably in Bayesian model choice (see Jeffreys-Lindley paradox or sources in the further reading). In that sense their use does not allow the full Bayesian approach.

“Uninformative and vague priors are not inappropriate for model selection. In model selection, we can compute the posterior odds ratio. The can, without any inappropriateness, compare models with these priors.”
 - The reviewer seems to not be aware of the issues arising in Bayesian model comparison. This issue is often summarized as the Jeffreys-Lindley paradox. There is a rich literature on the choice of priors and its effect on Bayesian analysis. Most textbooks in the further reading section will include a discussion on this, e.g. :cite:`gelman2013bayesian` chapter 7.4

“The need to choose priors is a common critic to Bayesian; however, it is misguided because alternative maximum likelihood approaches are simply equivalent to: (i) assuming that the priors are uniform; (ii) approximating the integral over the parameters as f(D | M) ≈ f(D | \theta*, M), where \theta* is the maximum likelihood estimator of M’s parameters. The integral is the marginal likelihood, which is obtained by marginalizing over plausible parameter values (i.e. f(\theta | M)).”
 - It is true that Bayesian point estimates (MAP) agree with frequentist MLE estimates under certain conditions. We intentionally avoid opening this discussion in this paper because a comparison between Bayesian and frequentist methods is beyond the scope of the paper.

In 3) Generative fake data
“I suggest replacing fake with synthetic.”
 - Replaced

“Last paragraph: when the authors say, "all models are wrong, but some models are useful," they should explain when the users are allowed to make a statement that the model M is close enoughto the real generative process. Without that said, it is vague.”
 - It is true that we do not discuss in any detail how to evaluate whether the model captures the reality to a satisfactory extent as that would need a longer treatment. We state that from the begining as one of the advanced steps that is beyond the scope of the paper. In this paper we can only hint to model fit questions. Given the space constraints we believe that using the commonly used motto "all models are wrong, but some models are useful," helps convey an important message without going into details. 

In 4) Fit the model to the fake data

“Last sentence in the second paragraph - we settle for an algorithm that returns samples from the posterior distribution: Are we settle to a model or an algorithm? Again, introducing the concept ofinference algorithm in the second subsection may avoid such confusion.”
- We assume that at this point in the paper the reader should have understood that the model M is "chosen and assumed known" as we state in the first two sentences of 2) Specify the likelihood and priors. We have added an extra footnote in section 2 to make sure that this is understood.

“The reader lacks the definition of a posterior confidence interval.”
- Good point, we corrected this and called it by its proper name, credible interval.[TO DO]

“The sub-subsection b: the whole paragraph may be redundant. Here, the authors discuss how model critique works in practice (i.e. with real data, not synthetic data). However, it is in the later stage of the workflow. In this subsection, we only concern about the sensitivity of the model and inference algorithm to the synthetic dataset.”
 - There seems to be a confusion here, so we added an example to clarify. [TO DO]
Model critique in practice is not only on real data. It is equally valid (and necessary) to critic the model on the fake data as well, ideally before moving on the real data. The reason is that there are problems that arise with models even before we look at the fit to the real data. For example there could be identifiability issues that have nothing to do with the real data. 
More generally model critique is a very difficult and delicate task. A big part of it is, undoubtedly, the issue of how well the model captures the reality. But there is also the issue of how well the model is structured, independently of whether the reality agrees with the model. By fitting to synthetic data and checking model fit in a controled environment with synthetic data we hope to catch errors that are present at the model level irrespective of how well it agrees with the real world.

In 5) Fit the model to the real data
“The reader lacks an understanding of when he/she should be satistied with the fit of the model.”
- True, again we cannot go into details in this topic. We add a sentence to explain that this is important but not adequately addressed here as it's beyond the intended level of this paper. 

“Last paragraph: when the authors suggest "A model that predicts well is a model that fits the data well." However, this is misleading. It is only correct if the model is consistent with the data. In other words, It is possible to find a model that doesn't fit the data so well but still predicts well (low generalization error).”
 - That’s a valid point in general. However, within the context discussed here (predictive checks) many argue that posterior predictive checks is a consistent measure of model fit. 

“In Further reading, the reader wants to know what a utility function is. This term first appears in the 9th step in the advanced workflow.”
 - We removed it.

mkcor:
Thanks again to all the reviewers!
@deniederhut Please consider for final review.
/cc @bayesways

mkcor:
Dear @aweinstein,

> In sec. 2.c, give a reference for the LKJ distribution. I think this distribution is not well know for people (like me!) outside the field.

Yes, this is a very valid point--addressed via https://github.com/scipy-conference/scipy_proceedings/pull/413/commits/aae47467d24d6d760982a7ba565a250e760de885, as done in the PyMC3 docs.

/cc @bayesways

deniederhut:
Dear @junipertcy and @alberto-antonietti , do you feel this paper is now ready to be included in the proceedings?

alberto-antonietti:
Hi @deniederhut, green light from my side!

aweinstein:
Hi @deniederhut : I was assigned to review this paper. I think it's good to go.

junipertcy:
Thank you @mkcor and @bayesways for your detailed response! 

Hi @deniederhut - I am slow but I think the paper is ready. Best wishes.

bayesways:
Hi @deniederhut, 

Thank you for helping organize a great conference. Looking forward to the next one! 

Concerning the publication of the proceedings: 

- is it available online?
- is there an DOI link?
- what's the right citation?

I found this [link](http://conference.scipy.org/proceedings/scipy2018/) but it does not contain our paper. 

Thanks again,
Konstantinos

deniederhut:
Of course! The only papers that are published online right now are the ones
that were ready before the deadline. We'll republish the proceedings with
all of the papers, bibfiles, and DOIs later this week.

PR #446, opened 2019-03-20
ENH: update conference metadata for 2019

deniederhut:
This is necessary so the pdf and html proceedings have the correct list of organizers, conference dates, etc. This also grabs the most recent logo from the website, which.... is probably bigger than we want. Someone should tweak this before publishing.


deniederhut:
@dalippa have you had a chance to look at this?

cbcunc (approved):

PR #449, opened 2019-03-20
ENH:Enable LaTeX submissions (WIP)

stargaser:
Work in Progress. Enable LaTeX submissions as an alternative to reStructuredText.

* if no `.rst` files are present, search input directory for `.tex`
* if processing `.tex`, copy the necessary style files.
* add an example LaTeX paper based on `00_vanderwalt`.

TODO: refactoring of code that was copied from `rst2tex` into a separate function to be used by both.


deniederhut (commented):
Holy cow, this is great! 

I'll have a couple of low level suggestions about the implementation in a bit, but the big question I have for you right now is about the latex preamble. For the rst files, it is constructed by docutils, with specific parts overridden by `publisher/writer/__init__.py`. For the tex files, it looks like we are counting on people copying and modifying the example file correctly. How hard would it be to append the preamble during compilation and have the papers start with `\begin{document}`? Do we know that the preamble is going to be the same for every paper? 

On a semi-related note, this feels... very autogenerated

```tex
Brutus$^{\setcounter{footnotecounter}{6}\fnsymbol{footnotecounter}\setcounter{footnotecounter}{2}\fnsymbol{footnotecounter}}$\setcounter{footnotecounter}{6}\thanks{\fnsymbol{footnotecounter} Unaffiliated}
```

stargaser:
Thanks @deniederhut for taking a look. The example file `00_vanderwalt.tex` is indeed auto-generated and looks forbidding for someone to work with. I can see a few options:
* do not include `00_vanderwalt.tex`, telling authors that we recommend they use reStructuredText and that LaTex is only for die-hards -- and that they need to take an autogenerated LaTeX file from `build_paper.py` as their starting point.
* make an alternative style file like `scipy-latex.sty` that includes as much as the docutils-generated preamble as possible
* tell people to start their LaTeX with `begin{document}`. We prepend the preamble before compiling.

The last option prevents anyone from working with their document in Overleaf, or compiling it with LaTeX themselves -- they would always have to use `build_paper.py`. If possible, I would like to make `scipy-latex.sty` and simplify the example file, so that standalone LaTeX can work.

Most of the example after `\begin{document}` looks to me like standard LaTeX. The exceptions are the author list (as you noted) and the colorized code section.

cc @dalippa @cbcunc for opinions on these options.

deniederhut:
Hm... so the second option seems best, but the third option seems easiest. 

I think it is ultimately your judgement call whether you would prefer to implement the scipy-latex.sty solution now, or implement the text-append solution now and defer making the Overleaf compatible style for later. I'm happy to support either, but would like to be able to ship a clean and working version of something by 12 April. What do you think?

stargaser:
After experimenting with a `scipy_latex.sty` file with the docutils preamble, I am thinking that a goal of enabling users to run LaTeX standalone is very daunting. I am inclined towards the first option I outlined above: do not include the example `00_vanderwalt.tex`; tell authors wanting to submit LaTeX that they need to start with a Markdown file, but they can take the intermediate LaTeX file generated by `build_paper.py`, modify that, and ultimately submit it.

stargaser:
Not having docutils generate `paper_stats.json` might be a showstopper.

stargaser:
Ah, although `make_paper.sh` prints an Error! message when there is no `paper_stats.json`, the process runs to completion and a paper is made. I'm not sure where `paper_stats.json` is used elsewhere in the system?

deniederhut:
So we chatted about this during the last committee call, but just to capture this somewhere:

paper_stats.json is produced during the first build of each paper. When they are combined into a full proceedings, the paper_stats.json for each paper (and scipy_proc.json in the root dir) are used to generate the title page, table of contents, bibfiles, etc.

deniederhut:
Closing in favor of adopting a pandoc-based approach some time next year.

PR #455, opened 2019-04-27
[2017] Correct video link for NEXT talk

stsievert:
When I wrote the paper I left the default link in thinking it'd automatically be changed (by default, it linked to a Monty Python video; it's down now).

This corrects the link to the link I thought would be automatically inserted.


deniederhut:
> automatically

 😅 we do this by hand after the conference videos are published, before publishing the final version of the proceedings. We must have missed yours when we went through the list - sorry! I'll merge this now, but I probably won't have time to republish the 2017 proceedings until this weekend.

PR #459, opened 2019-05-19
SJB edits to DESI description

sbailey:
Here are some updates to the DESI description including a new figure with example simulated data, in response to Rollin's suggestions.  Edit at will.

This is claiming that I changed all the png, which I did not purposefully do.


sbailey:
sorry; I think I'm targeting the wrong destination branch; trying again.

PR #462, opened 2019-05-20
Paper: PyLZJD: An Easy to Use Tool for Machine Learning

EdwardRaff:
process.

for more information.


papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+
+To help new users avoid this difficult task, we have developed the PyLZJD library. PyLZJD makes it easy to get started with ML algorithms and retrieval tasks without needing any kind of feature specification, selection, or engineering, to be done by the user. Instead, all a user needs to do is represent their data as a file (i.e., one file for every data point, for :math:`n` total files). PyLZJD will automatically process the file and can be used with Scikit-Learn to tackle many common tasks. While PyLZJD will likely not be the best method to use for most problems, it provides an avenue for new users to begin using machine learning with minimal effort and time. 
+
+The Lempel Ziv Jaccard Distance
+-------------------------------
+
```
bmcfee, line 44: Overall, the introduction here is great and clearly sets up the problem addressed by this tool.

In this section though, it would be helpful to clearly state up front what exactly the tool does: what it takes as input and what it produces as output.  As it's currently written, the functionality of the tool comes later, after some exposition about compression distance and hashing, and it's a little confusing to a first-time reader.

I think the rest of the section is great, but a preamble that outlines the procedure first (byte sequence => set => distance / vector) would strengthen it.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+What if we wanted to compare the similarity of two files, :math:`x` and :math:`y`? We can use compression to help us do that. Consider two files :math:`x` and :math:`y`, with absolutely no shared content. Then we would expect that if we concatenated :math:`x` and :math:`y` together to make one larger file, :math:`x \Vert y`, then compressing the concatenated version of the files should be about the same size as the files compressed separately, :math:`|C(x \Vert y)| = |C(x)| + |C(y)|`. But what if :math:`|C(x \Vert y)| < |C(x)| + |C(y)|`? For that to be true, there must be some overlapping content between :math:`x` and :math:`y` that our compressor :math:`C(\cdot)` was able to reuse in order to achieve a smaller output. The more similarity between :math:`x` and :math:`y`, the greater difference in file size we should see. In which case, we could use the ratio of compressed file lengths to tell us how similar the files are. We could call this a "Compression Distance Metric" :cite:`Keogh:2004:TPD:1014052.1014077` as shown in Equation :ref:`cdm`, where CDM(:math:`x,y`) returns a smaller value the more similar :math:`x` and :math:`y` are, and a larger value if they are different. 
+
```
bmcfee, line 50: It's a little misleading to call Eq 1 a "distance metric", since it does not satisfy identifiability.  (Triangle inequality may or may not be satisfied, depending on the choice of C.)  I think this should be reworded, or at least clarified in the text.

Note that this does not apply to the derived Jaccard similarity used later, which is a proper metric.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+.. raw:: latex
+
+	\subsection{Mechanics of LZJD} 
+
+While the above strategy has seen much success, it also suffers from drawbacks. Using a compression algorithm for every similarity comparison %makes prior methods
+is slow, and the mechanics of standard compression algorithms are not optimized for machine learning tasks. LZJD rectifies these issues by converting a specific compression algorithm, LZMA, into a dedicated distance metric :cite:`raff_lzjd_2017`. By doing so, LZJD is fast enough to use for larger datasets, and maintains the properties of a true distance metric\footnote{symmetry, indiscernibility, and the triangle inequality
+}. LZJD works by first creating the compression dictionary of the Lempel Ziv algorithm :cite:`Lempel1976`.
```
bmcfee: Typos; `%makes` and footnote did not render properly.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+
+.. code-block:: python
+
+    dist = 1.0-sim(lzset(a),lzset(b))
+
+
+While the procedure above will implement the LZJD algorithm, it does not include the speedups that have been incorporated into PyLZJD. Following :cite:`raff_lzjd_2017` we use Min-Hashing :cite:`Broder:1998:MIP:276698.276781` to convert a set :math:`A` into a more compact representation :math:`A'`, which is of a fixed size :math:`k` (i.e., :math:`|A'|=k`) but guarantees that :math:`J(A, B) \approx J(A', B')`. :cite:`raff_lzjd_digest` reduced computational time and memory use further by mapping every sub-sequence to a hash, and performing :code:`lzset` construction using a rolling hash function to ensure every byte of input was only processed once. To handle class imbalance scenarios, a stochastic variant of LZJD allows over-sampling to improve accuracy :cite:`raff_shwel`. All of these optimizations were implemented with Cython :cite:`behnel2010cython` in order to make PyLZJD as fast as possible.
```
bmcfee: Minor detail: Is this using permutation-based minhash, or the (open-vocabulary) hashing approximation?  It's maybe more detail than is strictly necessary for this article, but it might be a relevant point to mention since the purpose here is exactly to capture jaccard similarity.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+
+Working in such high dimensional spaces often requires greater consideration and expertise. To make PyLZJD easier for novices to use, we have developed a different vectorization strategy. To make this possible, we use a new version of Min-Hashing called "SuperMinHash", :cite:`Ertl2017`. The new SuperMinHash adds a fixed overhead compared to the prior method, but enables us to use what is known as :math:`b`-bit minwise hashing to convert sets to a more compact vectorized representation :cite:`Li:2011:TAB:1978542.1978566`. Since :math:`k \leq 1024` in most cases, and :math:`b \leq 8`, we arrive at a more modest :math:`d=k\cdot b \leq 8,192`. By keeping the dimension smaller, we make PyLZJD easier to use and a wider selection of algorithms from Scikit-Learn should produce reasonable results.
```
bmcfee: I didn't understand what "fixed overhead" means in this context.  A bit more detail would be appreciated here.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+.. raw:: latex
+
+	\subsection{Over-Sampling Data }
+
+Another feature introduced in :cite:`raff_shwel` is the ability to stochastically over-sample data to create artificially larger datasets. This is particularly useful when working with imbalanced datasets. Given a value :code:`false_seen_prob`, their approach modifies the inner if statement of :code:`lzset` to falsely "see" a sub-string that it has not seen before. This is a one line change that looks like the following:
+
+.. code-block:: python
+
+    if b_s not in s 
+      or random.uniform() < false_seen_prob:
```
bmcfee: Can this be random-seeded for reproducibility?

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+    from pyLZJD import digest, sim, vectorize
+
+
+These three functions work as follows:
+
+
+- :code:`digest`: takes in a string as data to convert to a digest, or a path to a file, and converts the file's content to an LZJD digest. If a list is given as input, each element of the list will be processed to return a list of digests. 
+- :code:`vectorize`: works the same as digest, but instead of returning a list, returns a numpy array representing a feature vector. 
+- :code:`sim`: takes two LZJD digests, and returns the similarity score between two files. 1.0 indicating they are exactly similar, and 0.0 indicating no similarity. 
+
```
bmcfee, line 147: General comment: it would be helpful to readers to quote call signatures for these functions instead of just the names, which are rather abstract.  E.g., `digest(list_or_files, processes=...)` instead of `digest`.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+
+The above is all that is needed for practitioners to use pyLZJD in their code. Below we will go through three examples of how to use these functions in conjunction with Scikit-Learn to get decent results on these problems. For new users, we recommend considering LZJD as a first-pass easy-to-use algorithm so long as the length of the input data is 200 bytes/characters or more. This recommendation comes from the fact that LZJD is compression based, and it is difficult to compress very short sequences. Another way to consider if LZJD may work for your data is to try manually compressing data points with your favorite compression algorithm. If the files compress well, LZJD may work. If the files do not compress well, LZJD is less likely to work. 
+
+.. raw:: latex
+
+	\subsection{T5 Corpus Example}
+
```
bmcfee, line 156: I found this example a bit confusing on a first read.  The previous section described the vectorization operation, which I think is how most people would expect this to interface with sklearn.  If you want to present the distance-based implementation first, it might help to explicitly state that that's what you're doing.  Otherwise, later statements like "Scikit-Learn needs a distance function between data..." seem strange.

bmcfee: Relatedly: is there any reason/benefit to using the set/distance representation over the vectorized representation?  If so, this would be a great place to mention it.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+To use LZJD with Scikit-Learn, we need to massage the files into a form that it expects. Scikit-Learn needs a distance function between data stored as a list of vectors (i.e., a matrix :math:`X`). However, our digests are not vectors in the way that Scikit-Learn understands them, and needs to know how to properly measure distances. An easy way to do this, which is compatible with other specialized distance a user may want to leverage, is to create a 1-D list of vectors. Each vector will store the index of its digest in the created :code:`X_hashes` list.  Then we can can create a distance function which uses the index, and returns the correct value. While wordy to explain, it takes only a few lines of code:
+
+.. code-block:: python
+
+    X = [ [i] for i in range(len(X_hashes))]
+    
+    def lzjd_dist(a, b):
+    	a_i = X_hashes[int(a[0])]
+    	b_i = X_hashes[int(b[0])]
+    	return 1.0-sim(a_i, b_i)
```
bmcfee: I did not understand this code block.  The `X` variable doesn't seem to be used, and it's not clear what's so special about the first elements `a[0]` and `b[0]`.

Aside: if this kind of distance function is standard, commonly used, and applies to any digest, why not include it in the library?

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+
+.. code-block:: python
+
+    from pyLZJD import digest, sim, vectorize
+
+
+These three functions work as follows:
+
+
+- :code:`digest`: takes in a string as data to convert to a digest, or a path to a file, and converts the file's content to an LZJD digest. If a list is given as input, each element of the list will be processed to return a list of digests.
```
bmcfee: The term `digest` was not defined previously.  Readers not familiar with minashing might be confused at this point.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+    def lzjd_dist(a, b):
+    	a_i = X_hashes[int(a[0])]
+    	b_i = X_hashes[int(b[0])]
+    	return 1.0-sim(a_i, b_i)
+
+
+This is all we need to use the tools already built into Scikit-learn. For example, we can perform :math:`k`-nearest-neighbor classification with cross validation to get an idea about how accurately we are able to predict a file's type. 
+
+.. code-block:: python
+
```
bmcfee, line 192: General question: is it possible to use digest or vectorize as feature extractors in sklearn that can be pipelined?  If so, that might simplify presentation a lot.  (It would also be pretty slick!)

If not, I'd humbly suggest adding that functionality. :)

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+    knn_model = KNeighborsClassifier(n_neighbors=5,
+        algorithm='brute', metric=lzjd_dist)
+    
+    scores = cross_val_score(knn_model, X, Y)
+    print("Accuracy: %0.2f (+/- %0.2f)" 
+        % (scores.mean(), scores.std() * 2))
+
+
+The above code returns a value of 91\% accuracy. This was all done without us having to specify anything about the associated file formats, how to parse them, or any feature engineering work. We can also leverage other distance metric based tools that Scikit-Learn provides. For example, we can use the t-SNE :cite:`Maaten2008` algorithm to create a 2D embedding of our data that we can then visualize with matplotlib. Using Scikit-Learn, this is only one line of code:
```
bmcfee: Example cites 91% accuracy, but does not provide a baseline for comparison.  What is the class balance in the data / what accuracy would a majority-vote classifier get?

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+
+    spam_paths = glob.glob("personal_image_spam/*")
+    ham_paths = glob.glob("personal_image_ham/*")
+    
+    all_paths = spam_paths + ham_paths
+    yBad = [1 for i in range(len(spam_paths))]
+    yGood = [0 for i in range(len(ham_paths))]
+    y = yBad + yGood
+    X = vectorize(all_paths)
+
```
bmcfee, line 241: A figure with examples of ham and spam images would go a long way to making this example more intuitive to readers.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+	\begin{tabular}{lcc}
+	\hline
+	\multicolumn{1}{c}{Metric} & Score    & Over-sampled Score \\ \hline
+	Accuracy                   & 0.946429 & 0.956767           \\
+	Precision                  & 0.950437 & 0.953824           \\
+	Recall                     & 0.965926 & 0.979259           \\
+	F1-Score                   & 0.958119 & 0.966374           \\
+	AUC                        & 0.987108 & 0.991602           \\ \hline
+	\end{tabular}
+    \end{table}
```
bmcfee, line 301: 6 decimal places is almost certainly excessive for data of `n~=1e4`.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+
+LZJD won't always be effective for images, and convolutional neural networks (CNNs) are a better approach if you need the best possible accuracy. However, this example demonstrates that LZJD can still be useful, and has been used successfully to find slightly altered images :cite:`Faria-joao`. This example also shows how to build a more deployable classifier with pyLZJD and tackle class-imbalance situations. 
+
+.. raw:: latex
+
+	\subsection{Text Classification}
+
+As our last example, we will use a text-classification problem. While other methods will work better, the purpose it to show that LZJD can be used in a wide array of potential applications. For this, we will use the well-known 20 Newsgroups dataset, which is available in Scikit-Learn. We use this dataset because LZJD works best with longer input sequences. For simplicity we will stick with distinguishing between the newsgroup categories of 'alt.atheism' and 'comp.graphics'. An example of a email from the later group is shown below.
```
bmcfee: `a email` => `an email`

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,322 @@
...
+
+.. code-block:: python
+
+    paths_train, paths_test, y_train, y_test = 
+      train_test_split(all_paths, y, 
+        test_size=0.2, random_state=42)
+    
+    X_train = vectorize(paths_train*10, 
+      false_seen_prob=0.05)
+    X_test = vectorize(paths_test)
```
bmcfee, line 281: It's a little strange up front to see train and test data treated differently here.

One suggestion for making this come across more clearly is to describe over-sampling in terms of data augmentation.  If you do that, then it might be a good idea to include `X_train` both with and without over-sampling.  Eg:

```python
X_train_clean = vectorize(paths_train)
X_train_aug = vectorize(paths_train * 9, false_seen_prob=0.05)
X_test = vectorize(paths_test)
```
This way, it's clear that the over-sampled data are "extra" in a sense, but you still have the original training data included.

bmcfee (commented):
I just gave this a first pass, and it already seems to be in pretty good shape -- well done!

I've made a few suggestions to help improve the draft.  The main thing that I think needs improving is a clear statement of what the package technically does: specifying its inputs and outputs.  The how and why can come later (and are clearly articulated as is).

EdwardRaff:
@bmcfee Thank you for the first review! 

Since this is a little different from normal academic review, I had two quick questions. 1) Should I be waiting for comments from other reviewers before making adjustments (or is it one-to-one)?  2) Should I discuss the comments in-line before making changes, or make changes and address the inline comments after? I'm not sure what the protocol is here.

bmcfee:
@EdwardRaff I'll defer to @deniederhut on this, but the [author instructions](https://github.com/scipy-conference/scipy_proceedings#author-workflow-steps) do imply iteration during review, so I think it's fine to make changes as you see fit.

deniederhut:
Correct - you should do whatever is easier/better for you. As for reviewers, we assigned multiple to each paper, but there is no need to wait for everyone to chime in before making changes.

papers/pylzjd/main.rst
EdwardRaff: Not in the current implementation, no. I'm not a cython/python guru, and had trouble getting random to work well/fast. So I use the current time and hash it as a PRNG.

papers/pylzjd/main.rst
EdwardRaff: Not currently, but a good suggestion! I will add looking into that to my TODO list.

papers/pylzjd/main.rst
EdwardRaff: Added that to the end of the first paragraph in this section. I also chose that as a good place to introduce what a digest is, per one of your other comments.

papers/pylzjd/main.rst
EdwardRaff: I mean to include a statement about that in the next section, which has now been added! I wanted to keep the conversation at this section at a "intuition level", without getting too into-the-weeds.

papers/pylzjd/main.rst
EdwardRaff: Fixed!

papers/pylzjd/main.rst
EdwardRaff: The bottom-k approach with a single hash function is used. Added that detail as a footnote.

papers/pylzjd/main.rst
EdwardRaff: Just that it's not as fast as the default method. Changed it to a more descriptive 'up to 40% slower"

papers/pylzjd/main.rst
EdwardRaff: Added!

papers/pylzjd/main.rst
EdwardRaff: Added text explaining this. Digest is preferable when using distance based algorithms because it will be both faster and more accurate.

papers/pylzjd/main.rst
EdwardRaff: I've tried to add some more inline comments to the code to further explain. 

re-aside: This approach is recommended by the [scikit-learn docs](https://scikit-learn.org/stable/faq.html#how-do-i-deal-with-string-data-or-trees-graphs). It seems more appropriate/easier to stick with how they recommend to do such things, rather than ask users to learn a different/specialized approach. I've added a footnote about that to the paper.

papers/pylzjd/main.rst
EdwardRaff: Defined way earlier now, see above!

papers/pylzjd/main.rst
EdwardRaff: Added. Majority vote gets you 24.5%

papers/pylzjd/main.rst
EdwardRaff: Added!

papers/pylzjd/main.rst
EdwardRaff: Trimmed down to 3.

papers/pylzjd/main.rst
EdwardRaff: Fixed!

papers/pylzjd/main.rst
EdwardRaff: I like having the "clean" and "aug" versions, and added that / slightly adjusted the following text. I'm a bit conflicted on describing it as data augmentation. I can see it helping at an intuition level, but I'm not _really_ augmenting the data. I haven't modified or transformed the underlying data in any realistic way. Instead we have different sub-sequences of the same underlying data stream. 

In our previous paper, we described/compared this approach with SMOTE. Do you think that would be helpful? I didn't include it in this as I wanted to minimize external ML background for the reader to follow along.

papers/pylzjd/main.rst
bmcfee: This looks great

papers/pylzjd/main.rst
bmcfee: Gotcha.  This point is not a blocker, but it might be worth documenting here.

For future reference, sklearn has a pretty simple pattern that you can follow for this sort of thing, though I don't know how well it plays with cython.

papers/pylzjd/main.rst
bmcfee: That makes sense -- it's much more clear now.

papers/pylzjd/main.rst
bmcfee: Maybe that's just a slight difference of perspective then.  I generally think of any synthetic transformation of data (realistic or not) for the purpose of improving robustness as DA, but it's of course up to you.

I think you're right that analogy to SMOTE might be more confusing for the scipy audience.

papers/pylzjd/main.rst
EdwardRaff: The random issue I remember became hairy when using multiple threads, and the python random isn't thread safe. But cython doesn't seem to provide the concept of an atomic integer, which is what I would want to use to increment a thread-local seed. 

Would you want to see that documented as a footnote?

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+:bibliography: Bib
+
+
+------------------------------------------------
+PyLZJD: An Easy to Use Tool for Machine Learning
+------------------------------------------------
+
+.. class:: abstract
+
+    As Machine Learning (ML) becomes more widely known and popular, so too does the desire for new users from other backgrounds to apply ML techniques to their own domains. A difficult prerequisite that often confounds new users is the feature creation and engineering process. This is especially true when users attempt to apply ML to domains that have not historically received attention from the ML community (e.g., outside of text, images, and audio). The Lempel Ziv Jaccard Distance (LZJD) is a compression based technique that can be used for many machine learning tasks. Because of its compression background, users do not need to specify any feature extraction, making it easy to apply to new domains. We introduce pyLZJD, a library that implements LZJD in a manner meant to be easy to use and and apply for novice practitioners. We will discuss the intuition and high-level mechanics behind LZJD, followed by examples of how to use it on problems of disparate data types.
```
ankurankan: Typo: `... easy to use and and apply ...`

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+For this first example, we will stick to using LZJD as a similarity tool and distance metric. When you want to use distance based algorithms, you want to use the :code:`digest` and :code:`sim` functions instead of :code:`vectorize`. :code:`vectorize` will be less accurate and slower when computing distances. 
+
+To use LZJD's digest with Scikit-Learn, we need to massage the files into a form that it expects. Scikit-Learn needs a distance function between data stored as a list of vectors (i.e., a matrix :math:`X`). However, our digests are not vectors in the way that Scikit-Learn understands them, and needs to know how to properly measure distances. An easy way to do this [#], which is compatible with other specialized distance a user may want to leverage, is to create a 1-D list of vectors. Each vector will store the index of its digest in the created :code:`X_hashes` list.  Then we can can create a distance function which uses the index, and returns the correct value. While wordy to explain, it takes only a few lines of code:
```
ankurankan: Typo `Then we can can create a ...`.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+        metric=lzjd_dist).fit_transform(X)
+
+.. figure:: t5_perp5.pdf
+    :align: center
+    :figclass: h
+    
+    Example of t-SNE visualization created using LZJD. Best viewed digitally and in color.
+
+
+A plot of the result is shown in Figure 1, where we see that the groups are mostly clustered into separate regions, but that there is a significant collection of points that were difficult to organize with their respective groups. While a tutorial on effective use t-SNE is beyond the scope of this paper, LZJD allows us to leverage this popular tool for immediate visual feedback and exploration.
```
ankurankan: Maybe `use of t-SNE` instead of `use t-SNE`?

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. class:: keywords
+
+   compression, complex data, machine learning
+
+Introduction
+------------
+
+Machine Learning (ML) has become an increasingly popular tool, with libraries like Scikit-Learn :cite:`scikit-learn` and others :cite:`xgboost,JMLR:v18:16-131,JMLR:v17:15-237,Hall2009` making ML algorithms available to a wide audience of potential users. However, ML can be daunting for news and amateur users to pick up and use. Before even considering what algorithm should be used for a given problem, feature creation and engineering is a prerequisite step that is not easy to perform, nor is it easy to automate.
```
ankurankan: Maybe `new and amateur` instead of `news and amateur`?

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+While the above strategy has seen much success, it also suffers from drawbacks. Using a compression algorithm for every similarity comparison makes prior methods
+slow, and the mechanics of standard compression algorithms are not optimized for machine learning tasks. Equation :ref:`cdm` also does not have the properties of a true distance metric [#]_, which can lead to confusing behavior and prevents using tools that rely on these properties. LZJD rectifies these issues by converting a specific compression algorithm, LZMA, into a dedicated distance metric :cite:`raff_lzjd_2017`. By doing so, LZJD is fast enough to use for larger datasets, and maintains the properties of a true distance metric. LZJD works by first creating the compression dictionary of the Lempel Ziv algorithm :cite:`Lempel1976`.
+
+.. [#] The properties of a true distance metric are symmetry, indiscernibility, and the triangle inequality.
+
+
+.. code-block:: python
+
+    def lzset(b): #b should be a list
```
ankurankan: Is `b` supposed to be a string instead of list? Because the code wouldn't work if it is a list, won't be able to add `b_s` (list) to `s` (set).

ankurankan (commented):
Sorry for the late review. @bmcfee covered most of the issues which have already been fixed. The paper seems to be in great shape already so I don't really have much to add except for a few very minor suggestions (mostly typos). Cheers!!

papers/pylzjd/main.rst
EdwardRaff: Ah, ok, thats correct. 

This code isn't how LZJD is implemented, so I'm a little conflicted on modifications. It should be a string the way it is written, but I don't want to imply that lzset only works over strings. I also don't want to necessarily convert the sub-list into a tuple so it would go into a set. 

Do you have any suggestions/thoughts on removing the comment, changing the comment to say a string, vs changing the code to accept tuples? I'm going to continue giving some thought on which might be best for the reader.

ankurankan: I am not sure what's the best solution in this case because as a reader I would expect the code to be working for all possible inputs. So, maybe it can be explicitly mentioned somewhere that the example code only works on one data type and the full implementation works on other types as well.
Or another possibility could be to show pseudocode instead of actual code to give the general idea of what's happening.

papers/pylzjd/main.rst
mfenner1: there are several references to feature extraction, creation, engineering and in the conclusion "feature specification" ... they are all related and varyingly similar ... but I'd pick one (probably feature extraction or feature engineering) and stick with it ... you might say a few words about terminology and that you are using one term as a blanket term for all of them

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+In normal use, we as ML practitioners would describe our data as a matrix :math:`\boldsymbol{X}` that has :math:`n` rows and :math:`d` columns. Each of the :math:`n` rows corresponds to one of our data points (i.e., an example), and each of the :math:`d` columns corresponds to one of our features. Using cars as an analogy problem, we may want to know what color a car is, how old it is, or its odometer mileage, as features. We want to have these features in every row :math:`n` of our matrix so that we have the information for every car.  Once done, we might train a model :math:`m(\cdot)` to perform a classification problem (e.g., is the car an SUV or sedan?), or use some distance measure :math:`d(\cdot, \cdot)` to help us find similar or related examples (e.g., which used car that has been sold is most like my own?).
```
mfenner1: using cars, we may ... [as an analogy problem is awkward]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+To help new users avoid this difficult task, we have developed the PyLZJD library. PyLZJD makes it easy to get started with ML algorithms and retrieval tasks without needing any kind of feature specification, selection, or engineering, to be done by the user. Instead, all a user needs to do is represent their data as a file (i.e., one file for every data point, for :math:`n` total files). PyLZJD will automatically process the file and can be used with Scikit-Learn to tackle many common tasks. While PyLZJD will likely not be the best method to use for most problems, it provides an avenue for new users to begin using machine learning with minimal effort and time.
```
mfenner1: without any kind of feature specification, selection, or engineering, to be done by the user. Instead, all a user needs to do is represent their data as a file ----> without any kind of feature specification, selection, or engineering from the user. Instead, a user represents their data as a file

will automatically process the files [otherwise it seems awkward since each data point is a file .... we have many to process]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+The Lempel Ziv Jaccard Distance
+-------------------------------
+
+LZJD stands for "Lempel Ziv Jaccard Distance" :cite:`raff_lzjd_2017`, and is the algorithm implemented in PyLZJD. LZJD takes a byte or character sequence :math:`x` (i.e., a "string"), converts it to a set of sub-strings, and then converts the set into a *digest*. This digest is a fixed-length summary of the input sequence, which requires a total of :math:`k` integers to represent. We can then measure the similarity of digests using a distance function, and we can trade accuracy for speed and compactness by decreasing :math:`k`. We can optionally convert this digest into a vector in euclidean space, allowing greater flexibility to use LZJD with other machine learning algorithms.
```
mfenner1: no comma after reference.  The form of the sentence is: LZJD stands for .... and .... .  No comma with that and. 
 There are several of these throughout the paper, which I'll note below.

recommend:  :math:`x` (a string)

recommend: Euclidean space

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+The inspiration and high-level understanding of LZJD comes from compression algorithms. Let :math:`C(\cdot)`  represent your favorite compression algorithm (e.g., zip, bz2, etc.), which will take an input :math:`x` and produce a compressed version :math:`C(x)`. Using a decompressor, you can recover the original object or file :math:`x` from :math:`C(x)`. The purpose of this compression is to reduce the size of the file stored on disk. So if :math:`|x|` represents how many bytes it takes to represent the file :math:`x`, the goal is that :math:`|C(x)| < |x|`.
```
mfenner1: (e.g., zip or bz2)    [they're examples, there could be more, don't need etc.]

which will take ---> [I'd strong recommend keeping almost all of the paper in the present tense.  writing in the future gets very wordy and awkward.]

The purpose of this compression is to reduce the size of the file stored on disk.   ---> Compression attempts to reduce the size of the file.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+What if we wanted to compare the similarity of two files, :math:`x` and :math:`y`? We can use compression to help us do that. Consider two files :math:`x` and :math:`y`, with absolutely no shared content. Then we would expect that if we concatenated :math:`x` and :math:`y` together to make one larger file, :math:`x \Vert y`, then compressing the concatenated version of the files should be about the same size as the files compressed separately, :math:`|C(x \Vert y)| = |C(x)| + |C(y)|`. But what if :math:`|C(x \Vert y)| < |C(x)| + |C(y)|`? For that to be true, there must be some overlapping content between :math:`x` and :math:`y` that our compressor :math:`C(\cdot)` was able to reuse in order to achieve a smaller output. The more similarity between :math:`x` and :math:`y`, the greater difference in file size we should see. In which case, we could use the ratio of compressed file lengths to tell us how similar the files are. We could call this a "Compression Distance Metric" :cite:`Keogh:2004:TPD:1014052.1014077` as shown in Equation :ref:`cdm`, where CDM(:math:`x,y`) returns a smaller value the more similar :math:`x` and :math:`y` are, and a larger value if they are different.
```
mfenner1, line 49: In which case, we could use the ratio of compressed ...  ----> We use the ratio of compressed ....

We call this a <ital>Compression Distance Metric</ital> [ref] (Equation 1)    [in almost all cases except one (noted below), I'd recommend italics instead of quotes

, where CDM(:math:`x,y`) returns a smaller value the more similar :math:`x` and :math:`y` are, and a larger value if they are different ---->  .  CDM(x,y) returns a smaller value when x and y are more similar.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. math::
+    :label: cdm
+
+    \text{CDM}(x,y) = \frac{C(x \Vert y)}{|C(x)| + |C(y)|}
+
+
+The CDM distance we just described gives the intuition behind LZJD, that is, we can use compression to measure the similarity between arbitrary files. CDM has been used to perform time series clustering and classification :cite:`Keogh:2004:TPD:1014052.1014077`. A large number of compression based distance measures have been proposed :cite:`Sculley:2006:CML:1126009.1126054` and used for tasks such as DNA clustering :cite:`Li2004`, image retrieval :cite:`doi:10.1117/12.704334`, and malware classification :cite:`Borbely2015`.
```
mfenner1: first sentence is a bit awkward.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. raw:: latex
+
+	\subsection{Mechanics of LZJD} 
+
+While the above strategy has seen much success, it also suffers from drawbacks. Using a compression algorithm for every similarity comparison makes prior methods
+slow, and the mechanics of standard compression algorithms are not optimized for machine learning tasks. Equation :ref:`cdm` also does not have the properties of a true distance metric [#]_, which can lead to confusing behavior and prevents using tools that rely on these properties. LZJD rectifies these issues by converting a specific compression algorithm, LZMA, into a dedicated distance metric :cite:`raff_lzjd_2017`. By doing so, LZJD is fast enough to use for larger datasets, and maintains the properties of a true distance metric. LZJD works by first creating the compression dictionary of the Lempel Ziv algorithm :cite:`Lempel1976`.
```
mfenner1: By doing so, LZJD is fast enough to use for larger datasets, and maintains ---> LZJD is fast enough to use for larger datasets and maintains [no comma]

papers/pylzjd/main.rst
mfenner1: you could include a comment:  # code for string case only

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+The :code:`lzset` method shows the Lempel compression dictionary creation process. Since LZJD cares about similarity as a direct goal, we do not put in the extra work or code normally required to make an effective compressor. Instead, we simply create a Python set of many different sub-strings of the input sequence :code:`b`. Because the :code:`lzset` method gives us a set of objects, we use the well-known Jaccard similarity to measure how close the two sets are. This is defined in the :code:`sim` method above, and mathematically in Equation :ref:`jaccard`. 
+
+.. math:: 
+    :label: jaccard
+
+    J(A, B)=\frac{|A \cap B|}{|A \cup B|}=\frac{|A \cap B|}{|A|+|B|-|A \cap B|}
+
+
+Defining a distance as :math:`d(A,B) = 1-J(A,B)` is a valid metric, and thus provides all the tools necessary to measure the similarity between arbitrary sequences or files. If :math:`a` and :math:`b` represent different sequences, their LZJD would be computed as:
```
mfenner1: Defining a distance as :math:`d(A,B) = 1-J(A,B)`   ---> The distance :math:`d(A,B) = 1-J(A,B)`

their LZJD would be computed as --> their LZJD is computed as

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. code-block:: python
+
+    dist = 1.0-sim(lzset(a),lzset(b))
+
+
+While the procedure above will implement the LZJD algorithm, it does not include the speedups that have been incorporated into PyLZJD. Following :cite:`raff_lzjd_2017` we use Min-Hashing :cite:`Broder:1998:MIP:276698.276781` to convert a set :math:`A` into a more compact representation :math:`A'`, which is of a fixed size :math:`k` (i.e., :math:`|A'|=k`) but guarantees that :math:`J(A, B) \approx J(A', B')` [#]_. :cite:`raff_lzjd_digest` reduced computational time and memory use further by mapping every sub-sequence to a hash, and performing :code:`lzset` construction using a rolling hash function to ensure every byte of input was only processed once. To handle class imbalance scenarios, a stochastic variant of LZJD allows over-sampling to improve accuracy :cite:`raff_shwel`. All of these optimizations were implemented with Cython :cite:`behnel2010cython` in order to make PyLZJD as fast as possible.
```
mfenner1: by mapping every sub-sequence to a hash, and performing :code:`lzset` ---> by mapping every sub-sequence to a hash and performing :code:`lzset`  [no comma]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. [#] The bottom-:math:`k` approach is used by default, where one hash :math:`h(\cdot)` is applied to every item in the set, and the bottom-:math:`k` values according to :math:`h(\cdot)` are selected. 
+
+.. raw:: latex
+
+	\subsection{Vectorizing Inputs}
+
+The LZJD algorithm as discussed so far provides only a distance metric. This is valuable for search and information retrieval problems, many clustering algorithms, and :math:`k`-nearest-neighbor style classification, but does not avail ourselves to all the algorithms that would be available in Scikit-Learn. Prior work proposed one method of vectorizing LZSets :cite:`raff_shwel` based on feature hashing :cite:`Weinberger2009a`, where every item in the set is mapped to a random position in a large and high dimensional input (they used :math:`d=2^{20}`). For new users, we want to avoid such high dimensional spaces to avoid the "curse of dimensionality" :cite:`Bellman1957`, a phenomena that makes obtaining meaningful results in higher dimensions difficult.
```
mfenner1: style classification, but it does not [add it]

the "curse of dimensionality" --> prefer italics

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. code-block:: python
+
+    if b_s not in s 
+      and random.uniform() > false_seen_prob:
+
+
+By doing so, the set of sub-strings returned will be altered. However, the altered set is still true to the data in that every string in the set is a real and valid sub-string from the corpus. This works because the Lempel Ziv dictionary creation is sensitive to small changes in the input, so a few small alterations can propagate forward and cause a number of differences in the entire process. By making the condition random, we can repeat the process several times and get  different results each time, providing additional diversity that can help train a model. When :code:`false_seen_prob` = 0, we get the standard LZJD output. To perform oversampling, we recommend using small values like :code:`false_seen_prob` :math:`\leq 0.05`.
```
mfenner1: will be altered. --> is altered.

mfenner1: get  different results each time, providing additional diversity that can help train a model. ----> get  different results each time.  This provides additional diversity that can help train a model.   [perhaps say "example diversity" to clarify]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+      and random.uniform() > false_seen_prob:
+
+
+By doing so, the set of sub-strings returned will be altered. However, the altered set is still true to the data in that every string in the set is a real and valid sub-string from the corpus. This works because the Lempel Ziv dictionary creation is sensitive to small changes in the input, so a few small alterations can propagate forward and cause a number of differences in the entire process. By making the condition random, we can repeat the process several times and get  different results each time, providing additional diversity that can help train a model. When :code:`false_seen_prob` = 0, we get the standard LZJD output. To perform oversampling, we recommend using small values like :code:`false_seen_prob` :math:`\leq 0.05`. 
+
+
+Using PyLZJD
+-------------
+
+Now that we have given the intuition and described how LZJD works, we show by example how the implementation PyLZJD can be used to do machine learning on a number of different problems, without having to specify a feature processing pipeline. PyLZJD, along with complete versions of these examples, can be found at https://github.com/EdwardRaff/pyLZJD.
```
mfenner1: we show by example how the implementation PyLZJD can be used to do machine learning on a number of different problems, without having to specify a feature processing pipeline. ----> we show three examples of how  PyLZJD performs machine learning, without having to specify a feature processing pipeline.     [quick comment, sometime PyLZJD is written as pyLZJD , particularly in the last section, i think]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+
+Using PyLZJD
+-------------
+
+Now that we have given the intuition and described how LZJD works, we show by example how the implementation PyLZJD can be used to do machine learning on a number of different problems, without having to specify a feature processing pipeline. PyLZJD, along with complete versions of these examples, can be found at https://github.com/EdwardRaff/pyLZJD. 
+
+To use PyLZJD, at most three functions need to be imported, as shown below.
```
mfenner1, line 134: To use PyLZJD, we import three functions:

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. code-block:: python
+
+    from pyLZJD import digest, sim, vectorize
+
+
+These three functions work as follows:
+
+
+- :code:`digest(b, hash_size=1024, mode=None, processes=-1, false_seen_prob=0.0)`: takes in a string as data to convert to a digest, or a path to a file, and converts the file's content to an LZJD digest. If a list is given as input, each element of the list will be processed to return a list of digests. [#]_
```
mfenner1: takes in a string as data to convert to a digest, or a path to a file, and converts the file's content to an LZJD digest. ----> takes in (1) a string as data to convert to a digest or (2) a path to a file and converts the file's content to an LZJD digest.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+- :code:`sim(A, B)`: takes two LZJD digests, and returns the similarity score between two files. 1.0 indicating they are exactly similar, and 0.0 indicating no similarity. 
+
+.. [#] :code:`mode` controls which version of min-hashing is used. :code:`None` for the standard hash, or :code:`"SuperHash"` to use the approach that is compatible with vectorization. 
+
+
+The above is all that is needed for practitioners to use pyLZJD in their code. Below we will go through three examples of how to use these functions in conjunction with Scikit-Learn to get decent results on these problems. For new users, we recommend considering LZJD as a first-pass easy-to-use algorithm so long as the length of the input data is 200 bytes/characters or more. This recommendation comes from the fact that LZJD is compression based, and it is difficult to compress very short sequences. Another way to consider if LZJD may work for your data is to try manually compressing data points with your favorite compression algorithm. If the files compress well, LZJD may work. If the files do not compress well, LZJD is less likely to work.
```
mfenner1: is 200 bytes/characters or more. This recommendation comes from the fact that LZJD is compression based, and it is difficult to compress very short sequences. ---> is 200 bytes/characters or more:  it is difficult to compress very short sequences.

Another way to consider if LZJD may work for your data is to try manually compressing data points with your favorite compression algorithm. --->  A quick test of LZJD's appropriateness, is to manually compress your data points (as files) with your favorite compression algorithm.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. raw:: latex
+
+	\subsection{T5 Corpus Example}
+
+The first example we will use is a dataset called "T5", which has historically been used for computer forensics :cite:`Roussev2011`. It contains 4,457 files that are of one of 8 different file types: html, pdf, text, doc, ppt, jpg, xls, or gif. As a simple first step to using pyLZJD, we will attempt to classify a file as one of these 8 file types. Our code starts by collecting the paths to each file into a list :code:`X_paths`. Creating a LZJD digest for each of these files is simple, and calls the :code:`digest` function, as shown below:
```
mfenner1: The first example we will use is a dataset called "T5", ---> The first example we use is a dataset called <ital>T5</ital>,

mfenner1: Creating a LZJD digest for each of these files is simple, and calls the :code:`digest` function, as shown below: ----> Creating a LZJD digest for each of these files is simple; call `digest`:

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. code-block:: python
+
+    X_hashes = digest(X_paths, processes=-1)
+
+
+The processes argument is optional. By setting it to -1, as many processor cores as are available will be used. If set to any positive value :math:`n`, then :math:`n` cores will be used. A list of digests will be returned with the same corresponding order as the input. The :code:`digest` function will automatically load every file path from disk, and perform the LZJD process outlined above.
```
mfenner1: as many processor cores as are available will be used. --> as many processor cores as are available are used.
[future tense --> present tense in rest of paragraph]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. code-block:: python
+
+    X_hashes = digest(X_paths, processes=-1)
+
+
+The processes argument is optional. By setting it to -1, as many processor cores as are available will be used. If set to any positive value :math:`n`, then :math:`n` cores will be used. A list of digests will be returned with the same corresponding order as the input. The :code:`digest` function will automatically load every file path from disk, and perform the LZJD process outlined above. 
+
+For this first example, we will stick to using LZJD as a similarity tool and distance metric. When you want to use distance based algorithms, you want to use the :code:`digest` and :code:`sim` functions instead of :code:`vectorize`. :code:`vectorize` will be less accurate and slower when computing distances.
```
mfenner1: will be less accurate and --> is less accurate and

papers/pylzjd/main.rst
mfenner1: distance function between data stored as a list of vectors (i.e., a matrix :math:`X`). ---> distance function between examples as a matrix :math:`X`.

and needs to know how to properly measure distances.   [I was unclear what this referred to]

easy way to do this [#], which is   [is this a broken reference?]

mfenner1: Then we can can create a distance function which uses the index, and returns the correct value. ---> Then we create a distance function which uses the index and returns the correct value.  [no comma]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+For this first example, we will stick to using LZJD as a similarity tool and distance metric. When you want to use distance based algorithms, you want to use the :code:`digest` and :code:`sim` functions instead of :code:`vectorize`. :code:`vectorize` will be less accurate and slower when computing distances. 
+
+To use LZJD's digest with Scikit-Learn, we need to massage the files into a form that it expects. Scikit-Learn needs a distance function between data stored as a list of vectors (i.e., a matrix :math:`X`). However, our digests are not vectors in the way that Scikit-Learn understands them, and needs to know how to properly measure distances. An easy way to do this [#], which is compatible with other specialized distance a user may want to leverage, is to create a 1-D list of vectors. Each vector will store the index of its digest in the created :code:`X_hashes` list.  Then we can can create a distance function which uses the index, and returns the correct value. While wordy to explain, it takes only a few lines of code:
+
+.. code-block:: python
+
+    #This will be the vetor given to Scikit-Learn 
+    X = [ [i] for i in range(len(X_hashes))]
```
mfenner1, line 173: should this be X_hash =?   [which is used in lzjd_dist immediately below]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+    	digest_a = X_hashes[int(a[0])]
+    	digest_b = X_hashes[int(b[0])]
+    	#Now that we have the digests, compute a 
+    	#distance measure. 
+    	return 1.0-sim(digest_a, digest_b)
+    	
+    	
+.. [#] This approach is how the Scikit-learn developers recomend using other non-standard distance metrics. For example, the Scikit-learn `FAQ <https://scikit-learn.org/stable/faq.html#how-do-i-deal-with-string-data-or-trees-graphs>`_ shows how to use this approach for doing edit-distance over strings. 
+
+This is all we need to use the tools already built into Scikit-learn. For example, we can perform :math:`k`-nearest-neighbor classification with cross validation to get an idea about how accurately we are able to predict a file's type.
```
mfenner1: the tools already built into Scikit-learn. ---> the tools built into Scikit-learn.

classification with cross validation to get an idea about how accurately we are able to predict ---> classification with cross-validation to see how accurately we predict

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+    knn_model = KNeighborsClassifier(n_neighbors=5,
+        algorithm='brute', metric=lzjd_dist)
+    
+    scores = cross_val_score(knn_model, X, Y)
+    print("Accuracy: %0.2f (+/- %0.2f)" 
+        % (scores.mean(), scores.std() * 2))
+
+
+The above code returns a value of 91\% accuracy, where a majority-vote baseline would return 25\%. This was all done without us having to specify anything about the associated file formats, how to parse them, or any feature engineering work. We can also leverage other distance metric based tools that Scikit-Learn provides. For example, we can use the t-SNE :cite:`Maaten2008` algorithm to create a 2D embedding of our data that we can then visualize with matplotlib. Using Scikit-Learn, this is only one line of code:
```
mfenner1: where a majority-vote baseline would return 25\%  --> where a majority-vote baseline returns 25\%

other distance metric based tools that Scikit-Learn provides. ---> other distance metric tools from Scikit-Learn.

mfenner1: we can then visualize with ---> we can visualize with

papers/pylzjd/main.rst
mfenner1: A plot of the result is shown in Figure 1,  , where we see that ---> The resulting plot is shown in Figure 1.  We see that 

While a tutorial on effective use t-SNE is beyond the scope of this paper, LZJD allows us to leverage this popular tool ---> While a tutorial on effective t-SNE use is beyond our scope, LZJD allows us to leverage t-SNE

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+    Example of t-SNE visualization created using LZJD. Best viewed digitally and in color.
+
+
+A plot of the result is shown in Figure 1, where we see that the groups are mostly clustered into separate regions, but that there is a significant collection of points that were difficult to organize with their respective groups. While a tutorial on effective use t-SNE is beyond the scope of this paper, LZJD allows us to leverage this popular tool for immediate visual feedback and exploration. 
+
+.. raw:: latex
+
+	\subsection{Spam Image Classification}
+
+The prior example used files of varying types, which is similar to the problem domain that LZJD was developed for. In this example, we will change the type of data and how we approach the problem. Here, our goal will be to predict if an email image attachment is a "spam" image (i.e., undesirable) or a "ham" image (i.e., desirable - or at least, more desirable than spam). This dataset was collected in 2007 :cite:`imageSpam2007`, with 3298 spam and 2021 ham images.
```
mfenner1: In this example, we will change the type of data and how we approach the problem.  ----> We change the type of data and how we approach the problem.

Here, our goal will be to predict if an email image attachment is a "spam" image ----> Here, our goal is to predict if an email image attachment is a "spam" image  [also prefer italics to quotes]

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+
+.. figure:: spam_ham_example.png
+    :align: center
+    :figclass: h
+    
+    Example of ham (left) and spam (right) images from the dataset's `website <https://www.cs.jhu.edu/~mdredze/datasets/image_spam/>`_.
+
+In this example, we will use the :code:`vectorize` function to create feature vectors for each data point. This may be desirable in order to build models that avoid the nearest neighbor search, which can be slow and cumbersome to deploy. The trade off is we spend more time during the training phase of the algorithm. Doing this with pyLZJD is simple, and the below code snippet handles the work of creating the labels, loading the files, and creating feature vectors, again, without us having to specify anything about the input.
```
mfenner1: In this example, we will use the :code:`vectorize` function  --->We use :code:`vectorize`

This may be desirable ..... [awkward]

pyLZJD ---> PyLZJD

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+    ham_paths = glob.glob("personal_image_ham/*")
+    
+    all_paths = spam_paths + ham_paths
+    yBad = [1 for i in range(len(spam_paths))]
+    yGood = [0 for i in range(len(ham_paths))]
+    y = yBad + yGood
+    X = vectorize(all_paths)
+
+
+Now that we have feature vectors, we can train a Logistic Regression model to predict if a new image is a spam or not. The code to do that and evaluate it (by several metrics) is shown below.
```
mfenner1: The code to do that and evaluate it (by several metrics) is shown below.  ---> The code to train and evaluate it (by several metrics) is:

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+    print("Precision: %f" %
+      metrics.precision_score(y_test, predicted))
+    print("Recall: %f" % 
+      metrics.recall_score(y_test, predicted))
+    print("F1-Score: %f" % 
+      metrics.f1_score(y_test, predicted))
+    print("AUC: %f" % auc)
+
+
+This should produce an accuracy of about 94.6\%, and an AUC of 98.7\%. In the above code snippet, we included the :code:`class_weight` parameter in an effort to aid the model with the class imbalance that is present in the data. There are more examples of spam images, which can bias a model toward calling most inputs "spam" by default. Using a 'balanced' class weight re-weights the data as if there was an equal number of examples of each class. With pyLZJD, you can perform a special type of over-sampling to help further reduce this impact and improve accuracy. In the below code segment, we show a simple version of using this ability.
```
mfenner1: This should produce an accuracy of about 94.6\%, and an AUC of 98.7\%. ---> This produces an accuracy of about 94.6\% and an AUC of 98.7\%.

parameter in an effort to aid the model with the class imbalance that is present in the data. ---> parameter to address class imbalance in the data.

In the below code segment, we show a simple version of using this ability.  ---> Here is a simple version of this ability:

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+      train_test_split(all_paths, y, 
+        test_size=0.2, random_state=42)
+    
+    X_train_clean = vectorize(paths_train) 
+    X_train_aug = vectorize(paths_train*10, 
+      false_seen_prob=0.05)
+    X_test = vectorize(paths_test)
+
+
+In this code, :code:`X_train_clean` constructs the training data in the normal manner. Alternatively, :code:`X_train_aug` has over-sampled both the spam and ham training data 10 times. Normally, this would create 10 copies of the same vectors, and have no impact on the solution learned. But, we added the :code:`false_seen_prob` flag, which alters how the :code:`lzset` is constructed: this flag turns on the stochastic component described above that makes you get a different result every time you call the function, so that we can get a variety of different (but realistic) examples for each datapoint. If we train a new logistic regression model on this data, we get improved results, which are shown in Table :ref:`spamImgResults`.
```
mfenner1: Normally, this would create 10 copies of the same vectors, and have no impact on the solution learned. ---> Normally, this would create 10 copies of the same vectors and have no impact on the solution learned. [no comma]

constructed: this flag turns on the stochastic component described above that makes you get a different result every time you call the function, so that we can get a variety of different (but realistic) examples for each datapoint. If we train a new logistic regression model on this data, we get improved results, which are shown in Table :ref:`spamImgResults`. ---> constructed: this flag turns on the stochastic component and you get a different result every call.  We get a variety of different (but realistic) examples for each datapoint. If we train a new logistic regression model on this data, we get improved results (Table :ref:`spamImgResults`).

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+	\multicolumn{1}{c}{Metric} & Score    & Over-sampled Score \\ \hline
+	Accuracy                   & 0.946 & 0.957           \\
+	Precision                  & 0.950 & 0.954           \\
+	Recall                     & 0.966 & 0.979           \\
+	F1-Score                   & 0.958 & 0.966           \\
+	AUC                        & 0.987 & 0.992           \\ \hline
+	\end{tabular}
+    \end{table}
+
+LZJD won't always be effective for images, and convolutional neural networks (CNNs) are a better approach if you need the best possible accuracy. However, this example demonstrates that LZJD can still be useful, and has been used successfully to find slightly altered images :cite:`Faria-joao`. This example also shows how to build a more deployable classifier with pyLZJD and tackle class-imbalance situations.
```
mfenner1: LZJD can still be useful, and has been used successfully to find slightly altered images ---> LZJD can still be useful and it found slightly altered images

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+	\end{tabular}
+    \end{table}
+
+LZJD won't always be effective for images, and convolutional neural networks (CNNs) are a better approach if you need the best possible accuracy. However, this example demonstrates that LZJD can still be useful, and has been used successfully to find slightly altered images :cite:`Faria-joao`. This example also shows how to build a more deployable classifier with pyLZJD and tackle class-imbalance situations. 
+
+.. raw:: latex
+
+	\subsection{Text Classification}
+
+As our last example, we will use a text-classification problem. While other methods will work better, the purpose it to show that LZJD can be used in a wide array of potential applications. For this, we will use the well-known 20 Newsgroups dataset, which is available in Scikit-Learn. We use this dataset because LZJD works best with longer input sequences. For simplicity we will stick with distinguishing between the newsgroup categories of 'alt.atheism' and 'comp.graphics'. An example of an email from the later group is shown below.
```
mfenner1: For this, we will use the well-known 20 Newsgroups dataset, which is available in Scikit-Learn. We use this dataset because LZJD works best ---> We will use the well-known 20 Newsgroups dataset, which is available in Scikit-Learn.  LZJD works best

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+	linear.
+	With 1*c_1 + 2*c_2 it works, but we have to find the best
+	compinations -- there's 10 levels, but 16 choises; best 10 must be
+	chosen. Different compinations for the same level, varies a bit, but
+	the levels keeps their order.
+
+	Readers should verify what I wrote... :-)
+
+
+When a string is not a valid path to a file, pyLZJD will processes the string itself to create a digest. This makes working with data stored as strings simple, and getting results is as easy as the code snippet below:
```
mfenner1: This simplifies working with strings, and getting results is as easy as:

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+    
+    clf = LogisticRegression()
+    clf.fit(X_train, newsgroups_train.target)
+    
+    pred = clf.predict(X_test)
+    metrics.f1_score(newsgroups_test.target, 
+        pred, average='macro')
+
+
+With the above code, we get an :math:`F_1` score of 83\%. It is important to note that in this case, using Scikit-Learn's TfidfVectorizer one can get 89\% :math:`F_1`. The point here is that with pyLZJD we can get decent results without having to think about what kind of vectorization is being performed, and that any string encoded data can be feed directly into the :code:`vectorize` or :code:`digest` functions to get immediate results.
```
mfenner1: With the above code, we get an :math:`F_1` score of 83\%.  Using Scikit-Learn's TfidfVectorizer achieves an :math:`F_1` of 89\% . Our point is that PyLZJD we can get decent results without thinking about what kind of vectorization is being performed, and any string encoded data can be feed directly into the :code:`vectorize` or :code:`digest` functions to get forms immediately usable by Scikit-Learn.

papers/pylzjd/main.rst
```diff
@@ -0,0 +1,347 @@
...
+    metrics.f1_score(newsgroups_test.target, 
+        pred, average='macro')
+
+
+With the above code, we get an :math:`F_1` score of 83\%. It is important to note that in this case, using Scikit-Learn's TfidfVectorizer one can get 89\% :math:`F_1`. The point here is that with pyLZJD we can get decent results without having to think about what kind of vectorization is being performed, and that any string encoded data can be feed directly into the :code:`vectorize` or :code:`digest` functions to get immediate results. 
+
+Conclusion
+----------
+
+We have shown, by example, how to use pyLZJD on a number of different datasets composed of raw binary files, images, and regular ASCII text. In all cases, we did not have to do any feature specifications to use pyLZJD, making application simpler and easier. This shortcut is particularly useful when feature specification is hard, such as raw file types, but can also make it easier for people to get into applying Machine Learning.
```
mfenner1: specification wasn't really used earlier in the paper ... perhaps prefer engineering or extraction

mfenner1 (commented):
I've proposed many grammatical edits primarily focused on improving clarity and simplifying sentence structure.  I'll give a more general review in the discussion.

mfenner1:
So, I love this paper.  Feature extraction/construction/engineering is an incredible important topic and I've been working with it for a solid 15 years.  I'm always happy to see new approaches and particularly approaches that aim to be domain agnostic and easy on the user.  I think you use just enough code, mathematics, and verbal descriptions to get your ideas across clearly.  I think you have a good overall balance of background, implementation, and examples.  

To reproduce your results, I feel like I would have a bit of struggle:  primarily in tying together the LZJD description with SuperMinHashing.  Even if public code is available, I would love to see an overall Python or pseudo-code algorithm this ties all of the pieces together.  [I don't know the page limit off the top of my head -- that could be an issue]  Is it feasible to include some pseudo-code for the SuperMinHashing?

Have you done (or do you plan to do) any larger scale studies on a variety of problems (for example, pick 20 datasets and compare some generic baseline learners against PyLZJD)?

I placed specific grammatical stuff in individual line level comments.

EdwardRaff:
@mfenner1 Thank you so much for the kind words and the extensive style edits/notes! I'll set aside some time to go through and adjust. 

While we still have 2 pages free, SuperMingHash is fairly complicated to specify. Adding just that to the paper would take at least 1.5 pages, and wouldn't leave enough space to make a large / complete specification of the whole process in pseudo-code. I think it would make it very difficult for non ML experts to follow along. 

I would love to do a study like that, and its kind-of on our radar, everything we've explored has been predominantly malware related - where LZJD is competitive with state-of-the-art for a number of tasks in part because of how difficult it is to get features. It is unfortunately a lot of work to go hunting for datasets like that, more than I have freely available. 

On our radar is to do something similar to the CDM paper, which showed how to use SAX to apply the CDM to time series classification/clustering problems, which worked quite well. Combining their approach with this [large archive of time series datasets](https://www.cs.ucr.edu/~eamonn/time_series_data_2018/) would allow us to do a large study without too much pain.

mfenner1:
Would there be a way to strike a balance where you tie the pieces together with high level pseudo code and give the SuperMinHash algorithm its own very high level (even just conceptual) breakdown as a sub-procedure?

I can definitely get behind leaving the details of SuperMinHash out, if it will just be confusing (and given that readers can go to those references if they want those details).  So, I guess I'm back to how to tie all the pieces together ... even if SuperMinHash is left as a black-box.

Best,
Mark

papers/pylzjd/main.rst
EdwardRaff: I like mfenner1's proposal, switched to that!

papers/pylzjd/main.rst
EdwardRaff: Fixed!

papers/pylzjd/main.rst
EdwardRaff: Fixed!

papers/pylzjd/main.rst
EdwardRaff: Fixed!

papers/pylzjd/main.rst
EdwardRaff: Fixed!

papers/pylzjd/main.rst
EdwardRaff: I would prefer to keep this section as originally written. Its intentionally in a style as if we are walking through the problem of devising the CDM metric "together" with the reader. I feel like the suggested verbiage makes it sound a bit more like "this is how we do/use it", which is actually not the case, and so I don't want to pass off that implication.

papers/pylzjd/main.rst
EdwardRaff: Adjusted

papers/pylzjd/main.rst
EdwardRaff: Adjusted!

papers/pylzjd/main.rst
EdwardRaff: Adjustments made!

papers/pylzjd/main.rst
EdwardRaff: Adjusted

papers/pylzjd/main.rst
EdwardRaff: Tried adjusting.

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed

EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed. Also standardized to capital PyLZJD (except for the code examples and URL link, where it needs to be lower case to work)

papers/pylzjd/main.rst
EdwardRaff: I've left as is, as you don't need to import all three functions in order to use PyLZJD.

papers/pylzjd/main.rst
EdwardRaff: Adjusted

papers/pylzjd/main.rst
EdwardRaff: I prefer the original text for the 200 bytes as it is more explicit into why, which I feel is valuable in this sentence. I do prefer your second wording suggestion though, and switched to it.

papers/pylzjd/main.rst
EdwardRaff: Fixed

EdwardRaff: Adjusted

papers/pylzjd/main.rst
EdwardRaff: Adjusted

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: It was a broken footnote, fixed! Also reworded sentence to try and make it clear. I prefer to keep the "as a list of vectors" verbiage because that is how the code actually operates, even if it is conceptually the same thing as a matrix (and could be converted to one).

EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: No, it should not. X tells you which index in X_hashes to use.

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed

EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Adjusted

papers/pylzjd/main.rst
EdwardRaff: adjusted & re-worded

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: I prefer the original wording here. I feel like the proposed wording doesn't make it clear that we are referring to an example of someone else's work with LZJD.

papers/pylzjd/main.rst
EdwardRaff: In the context of the larger paragraph, I think the original wording does a better job explaining to the reader why choices are being made, and so prefer to keep it over the proposed wording.

papers/pylzjd/main.rst
EdwardRaff: Fixed

papers/pylzjd/main.rst
EdwardRaff: Adjusted

papers/pylzjd/main.rst
EdwardRaff: Fixed

EdwardRaff:
I've tried to address all the individual comments in the latest commit

I'm not sure I fully understand your "tying it all together" concern. Perhaps it's due to a different set of expectations. 

I'm assuming the standard SciPy reader is not a Machine Learning researcher / does no have experience implementing algorithms from paper descriptions. Based on the other/previous papers, I was under the understanding that the target audience was predominately practitioners / how & why to use the software. 

The entire process could potentially be fit into the paper, but I think it will be jaring for almost all readers and not easy to follow unless they are already experienced ML researchers. At the same time though, I would expect those readers to be able to pick up from the references in this paper for the deeper detail to independently reproduce.  There is the additional issue that the rst format doesn't seem to have a method for me to write pseudo code.

deniederhut:
Hi everyone! I have not read this paper yet, but perhaps I can chime in with some perspective.

To address the easy question first, rst allows both preformatted text and text with code highlighting. You can see examples in the demo paper here -> https://github.com/scipy-conference/scipy_proceedings/blob/2019/papers/00_vanderwalt/00_vanderwalt.rst 

As for the target audience, SciPy tends to be split about 50/50 between academia and industry, but even this is becoming a less useful distinction as more companies become engaged in R&D, and more grad students become founders. The intention behind the conference proceedings is that papers should be accessible to a scientifically literate person who is not in your field (e.g. you can imagine that you are presenting this to a group of computationally savvy researchers from the Geological Survey), and enlightening to someone who is already experienced in your field (e.g. one of the TPOT developers). The examples are intentional - they were both at SciPy last year.

In a lot of ways, these are competing goals, and striking a good balance is hard. I might think about it this way:

1. Would the NGS researcher understand at a surface level what you did and why it matters?
2. Would the TPOT developer see enough detail to reproduce your experiment? (This would not necessarily need to be *in* the paper itself if you can link to source elsewhere)

EdwardRaff:
Hi @deniederhut I'm not sure I understand towhat you are referring to for the vanderwalt example. I do have code highlighting used in my current paper, but I don't think that would benefit writing more pseudo-code like @mfenner1 is asking for. Or, I'm not understanding how that is achieved/done in the vanderwalt example.

deniederhut:
Sorry, I thought you had said that rst does not support formatting for psudo-code -- I was pointing you to an example of how to write pseudo-code in rst.

EdwardRaff:
I'm confused. I don't see any psudo-code in that link? I'm thinking something like this:
Screen Shot 2019-06-24 at 9 50 23 PM

deniederhut:
Yeah, you won't get syntax highlighting like that, but you can still use the fixed-width formatting.

deniederhut:
Hi @mfenner1 @bmcfee and @ankurankan ! Do you feel this paper is now ready for inclusion in the proceedings?

bmcfee:
I've given it another once-over, and LGTM!

deniederhut:
It looks like our other reviewers are being quiet. I'm marking this as ready.

RaffEdwardBAH:
Thank you @deniederhut , and thank you to reviewers @bmcfee @ankurankan & @mfenner1 ! I think your feedback was valuable and help improved the paper. I'm looking forward to attending and seeing the other cool work, and hopefully getting some people interested in pyLZJD :)

mfenner1:
Sorry that I lost track of this, but yes ... should be good to go!

PR #464, opened 2019-05-21
Paper: Better and faster hyper-parameter optimization with Dask

stsievert:
Co-authors are Tom Augspurger and Matthew Rocklin. The paper will be built on http://procbuild.scipy.org/download/stsievert-dask-ml-model-selection


stsievert:
This paper includes some experimental results. As per https://github.com/dask/dask-ml/pull/221#issuecomment-493797071, I plan to re-run these before the conference to squash a bug fixed in https://github.com/dask/dask-ml/pull/497 (but rerunning very likely won't change the plots much).

NickleDave:
Hey @stsievert I'm one of the reviewers for this.
Nice work! Looks like Hyperband will be a very useful addition to Dask-ML.

Some high level things:
- Would it make sense to move the paragraph introducing Dask from the start of the related work section to the start of the "Adaptive Model Selection in Dask"? It's a little confusing to explain what Dask is and not mention it again for a page, and I think it helps to start the "related work" section with the 10k feet view of hyperparameter optimization
- Could you say a little more to give an intuitive sense of how Hyperband works when you first introduce it? You say "Hyperband trains many models in parallel and decides to stop models at  particular  times  to  preserve  computation" and then move on to a sketch of the proof; but as I understand it, *some* models get stopped at every "iteration" of Hyperband and only the ones whose validation loss is still decreasing at some rate considered promising are kept for the next iteration (right?). I think it would help to explain that only *some* models stop early while others that seem to be doing better are carried on for more training.
- I don't find Hyperband docs at the links in the footnotes ... but I'm guessing that's because the PR hasn't been merged yet, and those will work in the future?
- I notice the plots show loss, which I understand demonstrates how the algorithm works in terms of its derivation, but would it be helpful to show plots with error on a test set? Just to know for sure that the models actually perform better. I saw they do this in the original paper (but that was for massive experiments on many datasets)
- Would it be possible to provide one more example with a different algorithm? 
- and possibly some toy problem that someone could run somewhat easily when it's local on a laptop? I understand if both of those are  a big ask, given that this an algorithm aimed at reducing really expensive hyperparameter search, I just think some more generic example (e.g. a toy convnet classifying Fashion MNIST) might help make the demo more intuitive

After a first pass I would say this 90-95% of the way to meeting all the criteria.
@deniederhut I can provide a more detailed review later this week.

stsievert:
Thanks for the review @NickleDave! I've pushed a couple changes that make the paper easier to interpret IMO.

Before the individual responses, here are my TODOs:

- [x] move Dask paragraph in "prior work" section
- [x] add paragraphs better explaining Hyperband
- [x] in experiments: make validation dataset larger; have validation score be more stable
- [ ] in experiments: (possibly) compare with Scikit-learn's `RandomizedSearchCV` (or similar) on one machine.
- [ ] (for documentation) smaller example in dask-examples
- [x] (prior to start of conference) get #221 merged for doc fix

---

> move the paragraph introducing Dask ... to start of the "Adaptive Model Selection in Dask"?

Good suggestion. Done.

> intuitive [explanation] of how Hyperband works when you first introduce it?

I've edited the paragraphs just under the "Hyperband" section title, and added some words too. I think it better introduces the algorithm. Can you give some more feedback? Thank you.

> I don't find Hyperband docs at the links in the footnotes ... but I'm guessing that's because the PR hasn't been merged yet, and those will work in the future?

#221 should be merged shortly. I'll update the link then.

> I notice the plots show loss, ... show plots with error on a test set? Just to know for sure that the models actually perform better.

Loss is all I can show because this a regression problem (thanks – some typos fixed). I show validation loss.

Showing test loss instead of validation loss would require significant implementation. I'd rather the validation set be very large to approximate the true error.

> Would it be possible to provide one more example with a different algorithm?

Are you asking to compare one more model selection algorithm alongside `HyperbandSearchCV`, `HyperbandSearchCV(..., patience=True)` and `RandomizedSearchCV`?

I can try an equivalent version of `RandomizedSearchCV` on one machine, but I'm not sure it'll work.

> I just think some more generic example (e.g. a toy convnet classifying Fashion MNIST) might help make the demo more intuitive

I think future work will be to implement that example in [dask-examples][1]. I'm seeing this as a suggestion for documentation improvements, not as an edit for the paper. Am I reading that right?

(I've also added the appendix with the complete code, so now `# complete definition in Appendix` means something).

[1]:https://github.com/dask/dask-examples

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+-------------------------------------------
+Better and faster model selection with Dask
+-------------------------------------------
+
+.. class:: abstract
+
+   Nearly every machine learning estimator has parameters assumed to be given.
+   Finding the optimal set of these values is a difficult and time-consuming
+   process for most modern estimators and is called "model selection". A recent
+   breakthrough model selection algorithm Hyperband addresses this problem by
```
deniederhut: I think we are missing some commas here, i.e. "selection algorithm, Hyperband, addresses"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+   machine learning, model selection, distributed, dask
+
+Introduction
+============
+
+Training any machine learning pipeline requires data, an untrained model or
+estimator and parameters that change the model and data, a.k.a.
+"hyper-parameters". These hyper-parameters greatly influence the performance of
+the model but are typically assumed to be given. A good example is with
```
deniederhut: For readers unfamiliar with this process, I might be more explicit in saying that hyperparameters are needed to use the model, but you don't really know if you chose the right hyperparameters until after you are done using the model.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+adapting the ridge regression or LASSO models to the amount of noise in the
+data with the regularization parameter :cite:`marquardt1975`
+:cite:`tibshirani1996`.
+
+Model performance strongly depends on the hyper-parameters provided, even for
```
deniederhut: I feel like I've read something (a conference paper maybe?) in the last few years arguing that many of these 'state of the art' papers are not introducing better models, but better hyperparameter tuning.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+the simple examples above. This gets much more complex when more
+hyper-parameters are required. For example, a particular visualization tool
+(t-SNE) requires (at least) three different hyper-parameters
+:cite:`maaten2008visualizing`. The first section of a study on how to use this
+tool effectively is titled "Those hyper-parameters really matter"
+:cite:`wattenberg2016`.
+
+These hyper-parameters are typically assumed to be given. This requires
```
deniederhut: You're repeating the phrase "assumed to be given" quite a bit - I might think about using some different phrasing to keep things sounding less formulaic.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+the model but are typically assumed to be given. A good example is with
+adapting the ridge regression or LASSO models to the amount of noise in the
+data with the regularization parameter :cite:`marquardt1975`
+:cite:`tibshirani1996`.
+
+Model performance strongly depends on the hyper-parameters provided, even for
+the simple examples above. This gets much more complex when more
+hyper-parameters are required. For example, a particular visualization tool
+(t-SNE) requires (at least) three different hyper-parameters
+:cite:`maaten2008visualizing`. The first section of a study on how to use this
```
deniederhut: It might help motivate your argument if you were to highlight some examples of model performance differences that are solely explained by hyperparameter tuning. Are we talking .1% change in accuracy? Or 10%?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+:bibliography: refs
+
+-------------------------------------------
+Better and faster model selection with Dask
+-------------------------------------------
+
+.. class:: abstract
+
+   Nearly every machine learning estimator has parameters assumed to be given.
+   Finding the optimal set of these values is a difficult and time-consuming
```
deniederhut: What makes this difficult? Human effort? Clock time? Wall clock time?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+Model performance strongly depends on the hyper-parameters provided, even for
+the simple examples above. This gets much more complex when more
+hyper-parameters are required. For example, a particular visualization tool
+(t-SNE) requires (at least) three different hyper-parameters
+:cite:`maaten2008visualizing`. The first section of a study on how to use this
+tool effectively is titled "Those hyper-parameters really matter"
+:cite:`wattenberg2016`.
+
+These hyper-parameters are typically assumed to be given. This requires
+searching over the possible values to find the best value by some measure
```
deniederhut: This sentence feels a bit more abstract than necessary. What do you mean by "searching"? What about "some measure"?

To make my question a little more concrete: for example, it is more concrete to say that we are generally looking for the smallest prediction error on a holdout set than to say we are using "some measure".

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+Typically models are scored on unseen data through a "cross-validation" search.
+These searches are part of a "model selection" process because hyper-parameters
+are considered part of the model. Even in the simple ridge regression case
+above, a brute force search is required :cite:`marquardt1975`.
```
deniederhut: It sounds like you are trying to convey that these hyperparameters cannot successfully be set by heuristics a priori. I might go further, and emphasize to the reader that not only is tuning brute force, but that we are searching over the cartesian product of possible parameter values (so this can become a big task).

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+This gets more complex with many different hyper-parameter values to input, and
+especially because there's often an interplay between hyper-parameters. A good
+example is with deep learning, which has specialized techniques for handling
+many data :cite:`bottou2010large`. However, these optimization methods can't
+provide basic hyper-parameters because there are too many data. For example,
+the most basic hyper-parameter "learning rate" or "step size" is
+straightforward with few data but infeasible with many data
```
deniederhut: I might be more explicit in stating that by "many data", I think you mean some amount too large to be trained in a single batch.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+:cite:`maren2015prob`.
+
+Contributions
+=============
+
+A hyper-parameter search a.k.a "model selection" is required if high
+performance is desired. In practice, it's a burden for machine learning
```
deniederhut: Burden feels emotionally strong, but otherwise not very descriptive. Maybe something like "time-consuming"?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+researchers and practitioners. Ideally, model selection algorithms return high
+performing models quickly and are simple to use.
+
+Returning high performing models quickly will allow the user (e.g., a data
+scientist) to more easily use model selection. Having a better search method
+will allow them to easily find high performing models and remove the need for
```
deniederhut: The first half of this sentence feels like a repeat of the previous sentence.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+repeating model selection searches to obtain a better model. Returning this
+high performing model quickly would lower the barrier to performing model
```
deniederhut: What do you mean by barrier? Cost in dollars? Cost in time?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+selection.
+
+This work
+
+* provides implementation of a particular model selection algorithm, Hyperband
+  in Dask, a Python library that provides advanced parallelism. Hyperband
+  returns models with a high validation score with minimal training.  A Dask
+  implementation is attractive because Hyperband is amenable to parallelism.
+* makes a simple modifications to increase Hyperband's amenability to
```
deniederhut, line 103: "a simple modification

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+  parallelism.
+* provides an simple heuristic to determine the parameters to Hyperband, which
+  only requires knowing how many examples the model should observe and a rough
+  estimate on how many parameters to sample
+* provides validating experiments
+
+Hyperband treats computation as a scarce resource [#scarce]_ and has parallel
+underpinnings. Hyperband can only return high performing models with minimal
```
deniederhut: Why is this necessarily true?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+training because it evaluate models in parallel.
+
+In the experiments, Hyperband returns high performing models fairly quickly,
```
deniederhut: I feel like we've been repeating the statement that "Hyperband returns high performing models quickly" a lot. That's a great first sentence. The following sentences should be talking about how much faster it finds the best model than other approaches, how it accomplishes that feat, and what new capabilities or practices this speed will unlock.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+with the simple heuristic for determining the input parameters to Hyperband.
+The implementation can be found on the machine learning for Dask, Dask-ML. The
+documentation for Dask-ML is available at https://ml.dask.org.
+
+This paper will review other existing work for model selection before
+detailing the Hyperband implementation in Dask. A realistic set of experiments
+will be presented before mentioning ideas for future work.
+
+.. [#scarce] If computation is not a scarce resource, there is no benefit from
```
deniederhut: I have a vague memory of reading an article arguing that the combined deep learning experiments of 2019 burned as much energy as all of the personal vehicles in the U.S. I'm not sure it's correct to say that a smarter model selection process would have "no benefit", even if computation is easy to come by.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+A commonly used method for hyper-parameter selection is a random
+selection of hyper-parameters followed by training each model to completion.
+This offers several advantages, including a simple implementation that is very
+amenable to parallelism. Other benefits include sampling "important
+parameters" more densely over unimportant parameters :cite:`bergstra2012random`
+This randomized search is implemented in many places, including in Scikit-Learn
+:cite:`pedregosa2011`.
+
+These implementations do not adapt to previous training, and are by definition
+`passive`. `Adaptive` algorithms can return a higher quality solution in less
```
deniederhut: I don't believe you've defined what "passive" means in your technical conception.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+   score.
+
+The rest of this paper will focus on steps 2 and 3, which is where most of the
+work happens in model selection.
+
+A commonly used method for hyper-parameter selection is a random
+selection of hyper-parameters followed by training each model to completion.
+This offers several advantages, including a simple implementation that is very
+amenable to parallelism. Other benefits include sampling "important
+parameters" more densely over unimportant parameters :cite:`bergstra2012random`
```
deniederhut: "more densely than"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+These implementations do not adapt to previous training, and are by definition
+`passive`. `Adaptive` algorithms can return a higher quality solution in less
+time by adapting to previous training and choosing which hyper-parameters to
+sample. This is especially useful for difficult model selection problems with
+many hyper-parameters and many values for each hyper-parameter.
+
+Bayesian algorithms are popular as adaptive model selection algorithms. These
+algorithms treat the model as a black box and the model scores as a noisy
+evaluation of that black box. These algorithms try to tune a set of
```
deniederhut: All of these algorithms "try to tune a set of hyper-parameters over time". What makes the Bayesian algorithms different? What do they do at each iteration? Why is that important?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+:cite:`kleinbayesopt17` through AutoML [#automl]_. This package also includes
+Fabolas, a method that takes dataset size as input and allows for some
+computational control :cite:`klein2016`.
+
+.. [#automl] https://github.com/automl/
+
+Hyperband
+---------
+
+Hyperband is an adaptive model selection algorithm :cite:`li2016hyperband`, and
```
deniederhut: The "early stopping" part seems more relevant -- I might headline that.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+a principled early stopping scheme [#resources]_ for randomized
+hyper-parameter selections. At the most basic level, it partially trains
+models before stopping models with low scores, then
+repeats. By default, it stops training the bottom 33% of the available models
```
deniederhut: By iteration do you mean one step in the partial fit of your group of models? Or one step in the complete fitting of a group of models?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+on each iteration. This means that the number of models decay over time, and
+the surviving models have high scores.
+
+The amount of training to do before stopping models depends on the
+relative importance of training time and hyper-parameters. If training time
```
deniederhut: I misread this sentence twice. I might be more explicit that we are talking about the  magnitude of the effect that training time has on model performance, and not about the importance of training time to the impatient researcher.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+can be combined by using the Hyperband bracket framework `sequentially` and
+progressively tuning a Bayesian prior to select parameters for each bracket
+:cite:`falkner2018`. This work is also available through AutoML.
+
+There is little to no gain from adaptive searches if the passive search
+requires little computational effort. Adaptive searches spends choosing which
+models to evaluate to minimize the computational effort required; if that's not
+a concern there's not much value the value in any adaptive search is limited.
+
+Dask
```
deniederhut, line 282: It feels a bit discontinuous to have two sections about dask in the middle of your discussion of hyperband.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+.. TODO: add link to Hyperband docs
+
+Hyperband architecture
+----------------------
+
+There are two levels of parallelism in Hyperband, which result in two
+embarrassingly parallel for-loops:
+
+* an "embarrassingly parallel" sweep over the different brackets of the
```
deniederhut, line 337: I don't think we need to repeat "embarrassingly parallel" twice here.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+                               (5, 243)]
+       final_models = [sha(n, r) for n, r in brackets]
+       return top_k(final_models, k=1)
+
+In this pseudo-code, the train set and validation data are hidden, which ``train``
+and ``top_k`` rely on. ``top_k`` returns the ``k`` best performing
+models on the validation data and ``train`` trains a model for a certain number
+of calls to ``partial_fit``.
+
+Each bracket indicates a value in the tradeoff between hyper-parameter and
```
deniederhut: The bracket discussion feels incomplete. What is the first number in the tuple? What is the second? Does hyperband set these automatically? If so, how does it determine the number of partial fits in each step from the total number of models? Can I tune the superparameter of how aggressive the pruning method is? Do you have rules of thumb for how aggressive pruning can be before model accuracy deteriorates noticeably?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+training time importance. With ``max_iter=243``, the least adaptive bracket runs
+5 models until completion and the most adaptive bracket aggressively prunes off
+81 models.
+
+This architecture with many embarrassingly parallel for-loops and nested
```
deniederhut: We've already stated everything in this paragraph at some point earlier in the paper. What new idea are you trying to convey?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+this algorithm and train models from different brackets concurrently.
+
+Dask Distributed is required because of the nested parallelism and the decision
+to stop training low-performing models. This means the computational graph is
+dynamic and depends on other nodes in the graph.
+
+Input parameters
+----------------
+
+Hyperband is fairly easy to use as well. It only requires two input parameters:
```
deniederhut: As well as what?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+1. the number of ``partial_fit`` calls for the best model (via
+   ``max_iter``)
+2. the number of examples that each ``partial_fit`` call sees (which is
+   implicit and referred to as ``chunks``, which can be the "chunk size" of the
+   Dask array).
+
+These two parameters rely on knowing how long to train the model
+[#examples]_ and having a rough idea on the number of parameters to evaluate.
+Trying twice as many parameters with the same amount of computation requires
```
deniederhut, line 423: Are the chunks randomly drawn from the dataset? If not, what happens when the product of partial fits and chunk size is larger than the dataset?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+In comparison, random searches require three inputs:
+
+1. the number of ``partial_fit`` calls for `every` model (via ``max_iter``)
+2. how many parameters to try (via ``num_params``).
+3. the number of examples that each ``partial_fit`` call sees (which is
+   implicit and referred to as ``chunks``, which can be the "chunk size" of the
+   Dask array).
+
+Trying twice as many parameters with the same amount of computation requires
+doubling ``num_params`` and halving either ``max_iter`` or ``chunks``, which
```
deniederhut, line 438: Why did you decide to make chunk size the free parameter and not the number of values for each hyperparameter?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+means every model will see half as many data. An balance between training time
```
deniederhut: "A balance"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+and hyper-parameter importance is implicitly being decided upon. Hyperband has
+one fewer input because it sweeps over this balance's importance.
```
deniederhut: How is it sweeping over these values? I thought you stated above that pruning strength was a static parameter that could be set directly by the user?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+.. [#tolerance] Tolerance (typically via ``tol``) is a proxy for ``max_iter``
+   because smaller tolerance typically means more iterations are run.
+
+Dwindling number of models
+--------------------------
+
+At first, Hyperband evaluates many models. As time progresses, the number of
+models decay because Hyperband is a (principled) early stopping scheme.
+Hyperband varies how aggressively to stop model training per bracket. Each
+bracket performs something like a binary search but varies the amount of
```
deniederhut: How is this like a binary search?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+training between each decision. The least aggressive bracket lets a few models
+run without any stopping.
+
+This means towards the end of the computation, a few models can be training
```
deniederhut: Why do we want the last models to be pruned the least aggressively?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+while most of the computational hardware is free. This is especially a problem
+when computational resources are not free (e.g., with cloud platforms like
+Amazon AWS or Google Cloud Engine).
+
+Hyperband is a principled early stopping scheme, but doesn't protect against at
```
deniederhut: We have been repeating "Hyperband is a principled early stopping scheme" quite a bit by this point in the paper.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+least two common cases:
+
+1. when models have converged before training completes (i.e., the score stays
+   constant)
+2. when models have not converged and poor hyper-parameters are chosen (so the
+   scores are decreasing).
+
+These common use cases happen when the user specifies a poor set of
+hyper-parameters or that training continue for too long. Regardless,
+the scores of the models above will not increase too much with high
```
deniederhut: I'm not clear on the meaning of this sentence.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+.. [#pytorch] https://pytorch.org
+.. [#skorch] https://github.com/skorch-dev/skorch
+
+Problem
+-------
+
+This section will walk through an image denoising task. The inputs and desired
+outputs are given in Figure :ref:`fig:io+est`. This is an especially difficult
+problem because the noise variance varies slightly between images, which
+requires a model that's at least a little complex.
```
deniederhut: Could we be more explicit than "at least a little complex"?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+Model architecture & Parameters
+-------------------------------
+
+To address that complexity, let's use an autoencoder. These are a type of neural
+network that reduce the dimensionality of the input before expanding to the
+original dimension. This can be thought of a lossy compression. Let's create
```
deniederhut: "as a lossy compression"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+.. This autoencoder has two layers that compress
+
+Of course, this is a neural network so there are many hyper-parameters to tune.
+Only one effects the global optimum:
+
+* ``estimator__activation``: which specifies the activation the neural network
+  should use.
+
+This hyper-parameter is varied between 4 different choices, all different types
```
deniederhut: Which different variants of relu?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+of the rectified linear unit (ReLU). The other hyper-parameters control
+reaching the global optimum:
```
deniederhut: This sentence would seem to be in direct contradiction to "Only one effects the global optimum"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+:ref:`fig:io+est`.
+
+.. figure:: imgs/io+est.png
+   :align: center
+
+   The rows show in the ground truth, input and output respectively for the
+   denoising problem. The output is shown for the best model that Hyperband
+   finds. :label:`fig:io+est`
+
+``HyperbandSearchCV`` beat hand-tuning by a considerable margin. While manually
```
deniederhut: I think a more relevant comparison would be another algorithmic method, like bayesian optimization or randomized search.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+Performance
+-----------
+
+Let's compare three algorithms with the same model, parameters and validation
+data. The comparisons are shown in Figures :ref:`fig:calls`, :ref:`fig:time`
+and :ref:`fig:activity` and the legends for these plots is shown in Table
+:ref:`table:legend`. In these experiments, 25 workers are used with Dask,
+meaning that 25 tasks can complete in parallel.
+
+I will compare against a basic stop on plateau algorithm with particular
```
deniederhut: I think this is my favorite part of the paper, but I also think it could use some more situational context. What is the best accuracy we could hope to achieve? E.g. by exhaustive search? How long would that work take on 25 cores? I don't think these necessarily need to be plotted at each training iteration like you've done for the hyperband examples, but having anchor points would help make your results more interpretable.

deniederhut (changes requested):
Hi Scott! It looks like some of our reviewers were not able to complete their review of your paper by the deadline, so I've stepped in to add a few comments.

NickleDave:
Hi again @stsievert. Thank you for your quick reply to my comments.
The to-do list is a good idea; I'll be sure to copy that for future PRs.

I'll add a few more comments about specific language in the paper to those from @deniederhut, but I want to go back to the question about validation loss and an example with another algorithm.

I'm not asking you to show test loss (not quite sure what that would mean since you don't want to train on a test set). I'm also not asking you to compare with some other algorithm for hyperparameter optimization. I'm asking for another experiment in the paper that uses a different learning algorithm, for example supervised classification with a neural network, and that shows the error on a test set, so we know that HyperBand actually translates to an improvement in model performance. It's possible that (a) you don't have time to do that before the 22nd and (b) I'm not understanding the use cases of HyperBand, but I think for an "average" reader showing some other more typical task like Fashion MNIST (or, I don't know, choosing RNN hyperparameters for models that learn sequential permuted MNIST), and showing as they do in the original Hyperband paper that this produces clear improvements in error (not loss!) on a test set would be a good demonstration that the implementation in Dask-ML does the job.

NickleDave:
One more sort of high-level thing I'll say here before commenting on lines is that you seem to conflate hyperparameter optimization and model selection, e.g. repeatedly saying "hyperparameter tuning a.k.a. model selection". But obviously there's a lot that goes into selecting a model (e.g. am I doing linear regression predicting a single x from a single y) before you start applying graduate student descent to hyperparameters. Please rephrase those sentences to be more precise. For example: "Hyperparameter optimization is an important aspect of model selection, especially when dealing with models like neural networks that have millions of parameters and that require days of training on terabytes of data"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+:author: Matthew Rocklin
+:email: mrocklin@gmail.com
+:institution: NVIDIA
+:institution: Relevant work performed while employed for Anaconda, Inc.
+
+:bibliography: refs
+
+-------------------------------------------
+Better and faster model selection with Dask
```
NickleDave: hyperparameter optimization! =/= model selection

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+-------------------------------------------
+
+.. class:: abstract
+
+   Nearly every machine learning estimator has parameters assumed to be given.
+   Finding the optimal set of these values is a difficult and time-consuming
+   process for most modern estimators and is called "model selection". A recent
```
NickleDave: is called hyperparameter optimization! =/= model selection

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+:bibliography: refs
+
+-------------------------------------------
+Better and faster model selection with Dask
+-------------------------------------------
+
+.. class:: abstract
+
+   Nearly every machine learning estimator has parameters assumed to be given.
```
NickleDave: Define hyperparameters, e.g.
"...has a set of parameters set before training that determines how the algorithm learns from data, known as hyperparameters"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+   Finding the optimal set of these values is a difficult and time-consuming
+   process for most modern estimators and is called "model selection". A recent
+   breakthrough model selection algorithm Hyperband addresses this problem by
+   finding high performing estimators with minimal training and has theoretical
+   backing. This paper will explain why Hyperband is well-suited for Dask, the
+   required input parameters and the rationale behind some minor modifications.
+   Experiments with a neural network will be used for illustration and
+   comparison.
```
NickleDave: Conclude the abstract by explaining what the experiment demonstrates, e.g. "We demosntrate that Dask is indeed well-suited for an efficient parallelized implementation of Hyperband"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+.. class:: keywords
+
+   machine learning, model selection, distributed, dask
+
+Introduction
+============
+
+Training any machine learning pipeline requires data, an untrained model or
+estimator and parameters that change the model and data, a.k.a.
```
NickleDave: "parameters that dictate how the model fits the data, which are chosen before training, and hence are known as hyperparameters"

papers/scott_sievert/hyperband.rst
NickleDave: I also wouldn't say "hyperparameters are typically assumed to be given".

There's no a collective unconscious of hyperparameters shared by machine learning researchers, right? Most (good) papers state in the methods how they arrived at their hyperparameters, e.g., "hyperparameters were chosen by random search (obligatory citation of Bergstra and Bengio)" or grid search or at least "we tried different hyperparameters and found no qualitative difference".

You should use this paragraph by framing the core problem, and use the first sentence as a topic sentence that does so.

Example.
 "Many methods exist for hyperparameter optimization with different pros and cons. The simplest would be brute force grid search, as has been applied to ridge regression (your citation here). A  more efficient alternative to grid search with good theoretical grounding is random search (Bergstra and Bengio). In the past few years a Bayesian approach has been developed, as discussed in more detail in Related Work, but it remains unclear how effective this approach is. One alternative is to speed up random search; a recent algorithm known as Hyperband (citation) does this by framing hyperparameter optimization as a multi-arm bandit problem." (segue to Contribution)

papers/scott_sievert/hyperband.rst
stsievert: Whoop! It's not necessarily true, only with high probability. Thanks for the catch. This paragraph should be

> Hyperband returns high performing models with high probability, treats computation as a scarce resource [#scarce]_ and has parallel
underpinnings. Each of these will be expanded upon in the following sections.

stsievert:
@deniederhut and @NickleDave, thanks for all the specific comments! I've skimmed them, and I believe the resolution of the specific concerns mentioned will improve the paper clarity.

> I'm asking for another experiment in the paper that uses a different learning algorithm, for example supervised classification with a neural network, and that shows the error on a test set, so we know that HyperBand actually translates to an improvement in model performance. ... this produces clear improvements in error (not loss!) on a test set

I think I see what you're saying @NickleDave. I'm interpreting this as "it'd be nice to have a hard numerical metrics for `HyperbandSearchCV` and `IncrementalSearchCV` as the search proceeds. Accuracy will provide this because it's what users care about." Is that correct?

stsievert:
> before the 22nd

@NickleDave what's on the 22nd? The next deadline I see is "June 25: Final Recommendation and Comprehensive Review Deadlines" on the [SciPy timeline for 2019][1].

[1]:https://github.com/scipy-conference/scipy_proceedings#timeline-for-2019

deniederhut:
>  I'm asking for another experiment in the paper that uses a different learning algorithm, for example supervised classification with a neural network, and that shows the error on a test set, so we know that HyperBand actually translates to an improvement in model performance. 

My understanding is that the goal of Hyperband is to reduce training cost, not to increase the accuracy of the model. That said, it should not unduly damage the accuracy of the model, so I agree that it would be nice to have a data point for comparison.

stsievert:
@deniederhut, thanks for stepping in and volunteering to do the review!

> My understanding is that the goal of Hyperband is to reduce training cost, not to increase the accuracy of the model

This is correct in the informal theorem I've presented.

However, Li et. al. show that Hyperband is much more general, and agnostic to how models are trained and scored. Their theory relies on the "multi-armed bandit" framework, which treats every model as an iterative black box with some scoring function. This scoring function can use any dataset (i.e., train or test) and any metric (e.g., accuracy or negative loss).

papers/scott_sievert/hyperband.rst
stsievert: I've ripped out "assumed to be given" and replaced it with "need to be specified by the user".

NickleDave:
> > before the 22nd
> 
> @NickleDave what's on the 22nd? The next deadline I see is "June 25: Final Recommendation and Comprehensive Review Deadlines" on the [SciPy timeline for 2019](https://github.com/scipy-conference/scipy_proceedings#timeline-for-2019).

Yes I meant the 25th, sorry!

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+Future work
+===========
+
+The biggest area for improvement is using another application of the Hyperband
+algorithm: controlling the dataset size as the scarce resource.  This would
+treat every model as a black box and vary the amount of data provided. This
+would not require the model to implement ``partial_fit`` and would only require
+a ``fit`` method.
+
+Another area of future work is ensuring ``IncrementalSearchCV`` and all of it's
```
NickleDave: "its", not "it's"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+:ref:`table:legend`. In these experiments, 25 workers are used with Dask,
+meaning that 25 tasks can complete in parallel.
+
+I will compare against a basic stop on plateau algorithm with particular
+choices for ``patience`` and ``num_params``. Specifically, I choose a fairly
+aggressive value for ``patience`` and hence choose to evaluate twice as many
+hyper-parameters. This illustrates the choice between hyper-parameter vs.
+training time importance because training models for longer with the same
+computational effort would require a higher value for ``num_params`` and a
+lower and more aggressive of ``patience``.
```
NickleDave: >> lower and more aggressive of ``patience``
Please add the missing word here.

papers/scott_sievert/hyperband.rst
NickleDave: I think @stsievert means that the activation function of the neurons has no direct effect on optimization, unlike the learning rate, etc., but this confused me as well

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,756 @@
...
+
+Let's compare three algorithms with the same model, parameters and validation
+data. The comparisons are shown in Figures :ref:`fig:calls`, :ref:`fig:time`
+and :ref:`fig:activity` and the legends for these plots is shown in Table
+:ref:`table:legend`. In these experiments, 25 workers are used with Dask,
+meaning that 25 tasks can complete in parallel.
+
+I will compare against a basic stop on plateau algorithm with particular
+choices for ``patience`` and ``num_params``. Specifically, I choose a fairly
+aggressive value for ``patience`` and hence choose to evaluate twice as many
```
NickleDave: by "aggressive" you mean "low", right?

Did you define `patience` above? If not, please briefly define here for the unfamiliar reader, and clarify that a low value for patience will stop training sooner so more models can be trained.

And maybe just say "low" in the next sentence--"aggressive patience" sounds a little oxymoronic, at least it was unclear to me upon first reading

papers/scott_sievert/hyperband.rst
stsievert: Thanks. I've clarified this to

``` diff
- However, these optimization methods can't
-provide basic hyper-parameters because there are too many data. For example,
- the most basic hyper-parameter "learning rate" or "step size"  ...
+ However, these optimization methods can't
+ provide basic hyper-parameters because there are too many data to perform optimization efficiently :cite:`bottou2010large`. For example,
+ the most basic hyper-parameter "learning rate" or "step size"
```

papers/scott_sievert/hyperband.rst
stsievert: >  Cost in dollars? Cost in time?

Both. I've added a clarification.

stsievert: ...and the clarification is

``` diff
- would lower the barrier to performing model
+ would lower the expense and/or time barrier to performing model
```

papers/scott_sievert/hyperband.rst
stsievert: Thanks. I've added a paragraph in the "Contribution" section (which is where "quickly" is most common):

> **Contributions**
>
> A hyper-parameter search a.k.a "model selection" is required if high performance is desired. In practice, it's expensive and time-consuming for machine learning researchers and practitioners. Ideally, model selection algorithms return high performing models quickly and are simple to use.
>
> Returning quality hyper-parameters quickly relies on making decisions about which hyper-parameters to devote training time to. This might mean progressively choosing higher-performing hyper-parameter values or stopping low-performing models early during training.
>
> Returning high performing models quickly will allow the user (e.g., a data scientist) ...

papers/scott_sievert/hyperband.rst
stsievert: I've tried to make my intent, defining "passive" in that sentence more clear with

> These implementations are by definition `passive`  because they do not adapt to previous training.

papers/scott_sievert/hyperband.rst
stsievert: Here's a better explanation:

> These algorithms have an estimate of the optimal set of hyper-parameters and use some probabilistic methods to improve the estimate. The choice of which hyper-parameter value to evaluate depends on previous evaluations.

By contrast, Hyperband chooses all the hyper-parameters right away. It decides to stop models based on previous evaluations.

papers/scott_sievert/hyperband.rst
stsievert: Thanks. This varies over time; it's simpler to say "at certain times".

papers/scott_sievert/hyperband.rst
stsievert: I revamped this to separate the early stopping and importance:

> Naturally, model quality depends on two factors: the amount of training and the values of various hyper-parameters. If training time only matters a little, it makes sense to aggressively stop training models. On the flip side, if only training time influence the score, it only makes sense to let all models train for as long as possible and not perform any stopping.  The frequency at which Hyperband stops models early depends on these importances.
>
> Hyperband sweeps over the relative importance of hyper-parameter choice and amount of training.  This sweep over training time importance enables a formal mathematical statement that Hyperband will return a much higher performing model than the randomized search without early stopping returns. This is best characterized by an informal presentation of the main theorem:

papers/scott_sievert/hyperband.rst
stsievert: Thanks. I've added a paragraph. Now it reads

> Each bracket indicates a value in the tradeoff between hyper-parameter and training time importance, and is specified by the list of tuples in the example above. Each bracket is specified so that the total number of ``partial_fit`` calls is approximately the same among different brackets. Then, having many models requires pruning models very aggressively and vice versa with few models. As an example, with ``max_iter=243`` the least adaptive bracket has 5 models and no pruning. The most adaptive bracket has 81 models and fairly aggressive early stopping schedule.
>
> The exact aggressiveness of the early stopping schedule depends one optional input to ``HyperbandSearchCV``, ``aggressiveness``. The default value is 3, which has some theoretical motivation :cite:`li2016hyperband`.  ``aggressiveness=4`` is likely more suitable for initial exploration when not much is known about the model, data or hyper-parameters.

papers/scott_sievert/hyperband.rst
stsievert: Thanks for this catch!

papers/scott_sievert/hyperband.rst
stsievert: Chunks are randomly drawn from the dataset.

papers/scott_sievert/hyperband.rst
stsievert: Can you expand on your question? I made `chunks` (or `max_iter`) the free variable because it's desired to sample "twice as many parameters".

papers/scott_sievert/hyperband.rst
stsievert: Thanks! I've rewritten this with

> Of course, this is a neural network so there are many hyper-parameters to tune.
Only one hyper-parameter affects the model architecture:
> 
> * ``estimator__activation``, [...]
> 
> All other hyper-parameters do not influence the model architecture. They
control finding the optimal model after the architecture is fixed:

papers/scott_sievert/hyperband.rst
stsievert: I'm planning on implementing another (synthetic) example where these questions can be answered (or at least visualized). I'll incorporate this example later this weekend.

I'll try to bring a baseline into the new baseline, and provide some metrics for best possible. I've deferred discussion to https://github.com/stsievert/dask-hyperband-comparison/issues/2.

papers/scott_sievert/hyperband.rst
stsievert: I debated this too. I went with "model selection" because Scikit-learn has a sub-package called `model_selection`. 

I think "hyper-parameter optimization" is a better name.

stsievert: I've changed the paper and talk title with this feedback. :+1:

deniederhut:
Hey @stsievert ! It looks like you've been making some edits. Feel free to @ us when you feel this is ready for a second look 😄

stsievert:
Thanks @deniederhut! I've added some simulations, re-organized a bit and edited with a fine-toothed comb. I've pushed my changes, and I think they're ready for another look. They're visible at http://procbuild.scipy.org/download/stsievert-dask-ml-model-selection.

The largest edit I've made has been adding the simulations as suggested by @NickleDave. This is a good suggestion, and presents one of my results more cleanly. The resulting figure is a good illustration of this:

Screen Shot 2019-06-22 at 1 32 23 PM

This enables cleaner presentation of my two main takeaways:

* Hyperband is a principled early stopping scheme for `RandomizedSearchCV` that finds better hyperparameters with less training
* Hyperband takes less time in the presence of serial or parallel computational resources.

I am working on improving the graphs that show best score. The graph above ~~will be updated later today~~ has been updated, and the graph on best score vs time will be updated by tomorrow (hopefully with error bars too).

stsievert:
@NickleDave some of your comments have been more useful than I initially thought. With the synthetic result I can present a new result/figure and have implemented a pull request implementing with the new feature (https://github.com/dask/dask-ml/pull/527). The figure is in the paper and also https://github.com/dask/dask-ml/pull/527#issuecomment-504714808.

NickleDave:
I'm glad the comments were helpful.
My first impression is these additional figures do help clarify things.
Please do let us know when you are ready for another review.

@deniederhut how fixed is the deadline of the 25th?
I need to finish something that day but will have a good chunk of time to review Wednesday the 26th

deniederhut:
>  implemented a pull request implementing with the new feature

Whoa! This is one of the cooler things I've seen come out of our open review process. Come find me at SciPy and I'll buy both of you a beer 😄

> how fixed is the deadline

We try not to publicize this, but we build in some wiggle room to the deadlines. If Scott is okay getting the feedback on Wed, then it's okay with me

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,948 @@
+:author: Scott Sievert
+:email: scott@stsievert.com
+:institution: University of Wisconsin–Madison
+:institution: Relevant work performed while interning for Anaconda, Inc.
```
deniederhut, line 4: Could I ask you to add a :corresponding: tag here?

stsievert:
> If Scott is okay getting the feedback on Wed, then it's okay with me

Yup, fine by me. I'll try to reply on Wednesday night.

> Please do let us know when you are ready for another review.

I think I'm ready for another review as mentioned in https://github.com/scipy-conference/scipy_proceedings/pull/464#issuecomment-504684771. I'll put in some work to try and update the image denoising figure before Wednesday (though I believe only the image will change).

>  Come find me at SciPy and I'll buy both of you a beer 😄

:+1: :beer: plus GitHub user names will be on the badges this year!

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,948 @@
...
+These hyperparameters need to be specified by the user. There are no good
+heuristics for determining what the values should be.
+These values are typically found through a search over possible values through
+a "cross validation" search where models are scored on unseen holdout data.
+These searches are part of a hyperparameter optimization, and for certain more import. Even in the simple ridge regression case
+above, a brute force search is required :cite:`marquardt1975`. This brute force
+search quickly grows infeasible as the number of hyperparameters grow.
+
+Hyperparameter optimization gets more complex with many hyperparameters, and
+especially because there's often interplay between hyperparameters. A good
```
deniederhut: I might say interaction instead of interplay

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,948 @@
...
+of Hyperband, which trains a couple models to completion. This has a large
+impact on the time to solution with these parallel computational resources but
+doesn't have a large impact on the number of ``partial_fit`` calls.
+
+Performance
+-----------
+
+Anecdotally, ``HyperbandSearchCV`` performs well and beats manual hand-tuning
+by a considerable margin. While manually tuning, I considered any scores about
+:math:`-0.10` to be pretty good, and I obtained scores no higher than
```
deniederhut: What is the scoring metric here?

deniederhut (changes requested):

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,950 @@
...
+every job can be run. To get around this in very parallel environments, the
+worst performing models all have the same priority for each bracket.
+
+
+.. [#random-sampling-hyperband] As much as possible – Hyperband evaluates more
+   hyperparameter values. The random search without early stopping
+   evaluates every hyperparameter value Hyperband evaluates.
+
+Parallel Experiments
+====================
```
stsievert, line 707: I still need to re-run these simulations – the prioritization of jobs is ready for this now in  https://github.com/dask/dask-ml/pull/527

deniederhut:
@NickleDave did you find some time to re-review yesterday?

NickleDave:
I did not, will do at noon today if that's okay

NickleDave:
@stsievert finally getting to this, sorry I am so behind

I am going to suggest a bunch of small edits to the writing. Please know they are not meant to be nitpicky; I think you've done important work here and I'm trying to help you make it as clear to other people as possible.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+:bibliography: refs
+
+--------------------------------------------------------
+Better and faster hyperparameter optimization with Dask
+--------------------------------------------------------
+
+.. class:: abstract
+
+    Nearly every machine learning model requires that the user specify certain
+    parameters before training begins, aka "hyperparameters". Finding the
```
NickleDave: "aka" sounds informal, and hyperparameters doesn't need to be in quotes here.
Definite hyperparameters as clearly as possible right here at the start.
E.g., "...certain parameters that control how the algorithm fits model parameters to data, and are therefore known as hyperparameters"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+    optimal set of hyperparameters is often a time- and resource-consuming
+    process. A recent breakthrough hyperparameter optimization algorithm,
+    Hyperband :cite:`li2016hyperband`, can find high performing hyperparameters with minimal training
+    and has theoretical foundations. This paper will provide intuition for
```
NickleDave: either remove "has theoretical foundations" because it's not clear why this is important, or else explain why, e.g. "a recent breakthrough algorithm, Hyperband, formulates hyperparameter optimization as bandit problem which minimizes training, speeding up random search"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+    Hyperband, explain the slightly modified implementation and why it's
```
NickleDave: You say "the slightly modified implementation" but you haven't introduced the idea that your implementation modifies anything.

Say something like "this paper presents an intuitive introduction to Hyperband, and provides an implementation that is modified to take advantage of parallel processing via Dask, a Python library ..."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+    well-suited for Dask, a Python library that scales Python to larger
+    datasets and more computational resources. Experiments find high performing
+    hyperparameters more quickly in the presence of serial or parallel
+    computational resources with deep learning models.
```
NickleDave: Make clear here the distinction between your results and your conclusions.
E.g., "Experiments with deep learning models demonstrate our implementation rapidly finds high-performing hyperparameters. Additionally, we show that Dask enables taking advantage of parallel computing resources to further speed up hyperparameter optimization."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+
+   distributed computation, hyperparameter optimization, machine learning
+
+Introduction
+============
+
+Training any machine learning pipeline requires data, an untrained model or
+estimator and "hyperparameters", parameters chosen before training begins that
+help with cohesion between the model and data. The user needs to specify values
+for these hyperparameters in order to use the model. A good example is with
```
NickleDave: you don't need "with" here in "A good example is with"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+t-SNE requires (at least) three hyperparameters
+:cite:`maaten2008visualizing` and the first section in a study on how to use this
+tool effectively is titled "Those hyperparameters really matter"
+:cite:`wattenberg2016`.
+
+These hyperparameters need to be specified by the user. There are no good
+heuristics for determining what the values should be.
+These values are typically found through a search over possible values through
+a "cross validation" search where models are scored on unseen holdout data.
+These searches are part of a hyperparameter optimization, and for certain more import. Even in the simple ridge regression case
```
NickleDave: Sorry if I commented on this before and it's already in a commit somehwere;
I think you can lose this whole sentence "These searchers are a part of a hyperparameter...".
If you don't agree, then please fix the end where you say "and for certain more import".
Not at all clear to me what you meant here

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+above, a brute force search is required :cite:`marquardt1975`. This brute force
+search quickly grows infeasible as the number of hyperparameters grow.
+
+Hyperparameter optimization gets more complex with many hyperparameters, and
```
NickleDave: Try not to repeat a word three times in one sentence.
Rewrite as e.g. "Optimization becomes more complex as the number of hyperparameters increase, especially because of interactions between them."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+especially because there's often interaction between hyperparameters. A good
+example of hyperparameter optimization is with deep learning, which has
```
NickleDave, line 72: You lose me in this sentence, I think because you change subject from hyperparameter optimization to plain-old optimization in the middle of it.

Better to split up.
E.g., "Deep learning algorithms in particular have multiple hyperparameters. These algorithms use specialized optimizations for dealing with very large datasets, such as stochastic gradient descent, but large datasets make the problem of hyperparameter optimization even more difficult [Bot10]. Finding the optimal value for just a single hyperparameter for stochastic gradient descent is infeasible with many data [MH15]."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+  attractive because Hyperband is amenable to parallelism.
+* makes a simple modifications to increase Hyperband's amenability to
+  parallelism.
+* provides an simple heuristic to determine the parameters Hyperband requires,
+  which only requires knowing how many examples the model should observe and a
+  rough estimate on how many parameters to sample
+* provides validating experiments that also illustrate common use cases
+
+Hyperband treats computation as a scarce resource [#scarce]_ and has parallel
+underpinnings.  In experiments, Hyperband returns high performing models fairly
```
NickleDave: make clear you mean "experiments in this paper" and "the implementation presented here", not, say, experiments from the original Hyperband paper

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+these optimization methods can't provide basic hyperparameters because there
+are too many data to perform optimization efficiently :cite:`bottou2010large`.
+For example, the most basic hyperparameter "learning rate" or "optimization
+step size" is a quick computation with few data but infeasible for many data
+:cite:`maren2015prob`.
+
+Contributions
+=============
+
+A hyperparameter optimization is required if high performance is desired. In
```
NickleDave, line 82: This section is good

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+
+.. [#scarce] If computation is not a scarce resource, there is little benefit from
+   this algorithm.
+
+.. [#dask-ml] https://ml.dask.org.
+
+Related work
+============
+
+Hyperparameter optimization
```
NickleDave, line 130: Much improved presentation in this section, thank you

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+
+Naturally, model quality depends on two factors: the amount of training
+performed and the values of various hyperparameters. If training time only
+matters a little, it makes sense to aggressively stop training models. On the
+flip side, if only training time influences the score, it only makes sense to
+let all models train for as long as possible and not perform any stopping.
+
+Hyperband sweeps over the relative importance of hyperparameter choice and
+amount of training.
+This sweep over training time importance enables a theorem that Hyperband will return a much higher performing model than the
```
NickleDave, line 200: Ah I get it now! Thank you

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+scheduler has low overhead. It's implemented in pure Python and can scale
+from massive datasets to a cluster with thousands of cores to a laptop running
+single process. In addition, it's designed with interactive computing
+and provides rapid feedback and diagnostics to aid humans.
+
+
+Dask's implementation of Hyperband
+==================================
+
+Dask can scale up to clusters or to massive datasets. Hyperparameter optimization searches
```
NickleDave: I would switch the order of sentences is in this paragraph.
Start with "Dask is a natural fit for Hyperband", remind us about HO and ||ism, and then say "Dask can enable this parallelism and scale up to large datasets by computing on clusters" (or something like that)

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+often require significant amounts of computation and can involve large
+datasets, and Hyperband is amenable to parallelism. Combining Dask
+with Hyperband is a natural fit.
+
+This work focuses on the case when the computation required is not
```
NickleDave: be more direct here, and don't start a sentence with "Then, " -- sounds weird, to my ear at least
"This work focuses on the case when when significant computational resources are required. In such cases, the existing hyperparameter..."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+insignificant. Then, the existing passive hyperparameter optimization algorithms in
+Dask-ML have limited use because they don't adapt to previous training to
+reduce the amount of training required.  [#dasksearchcv]_
+
+An adaptive hyperparameter optimization algorithm, Hyperband is implemented in Dask's
```
NickleDave: This sounds like if you haven't already been telling us about Hyperband for four pages.

Rephrase as "Our core contribution here is to provide an implementation of Hyperband as part of the Dask-ML library. Briefly we introduce aspects of the Hyperband algorithm that make it amenable to implementation in Dask, as well as details of the inputs" etc.

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+training time importance, and is specified by the list of tuples in the example
+above. Each bracket is specified so that the total number of ``partial_fit``
+calls is approximately the same among different brackets. Then, having many
+models requires pruning models very aggressively and vice versa with few
+models. As an example, with ``max_iter=243`` the least adaptive bracket has 5
+models and no pruning. The most adaptive bracket has 81 models and fairly
+aggressive early stopping schedule.
+
+The exact aggressiveness of the early stopping schedule depends one optional
+input to ``HyperbandSearchCV``, ``aggressiveness``. The default value is 3,
```
NickleDave: ah, this is why you were talking about aggressiveness below -- sry

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+   \label{fig:synthetic}
+   \end{figure}
+
+
+
+Model architecture & Hyperparameters
+-------------------------------------
+
+The model used is Scikit-learn's fully-connected neural network, their
+``MLPClassifier``. In this, there are several hyperparameters.  Only one is
```
NickleDave: "which has several hyperparameters."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+effects the architecture of the best model: ``hidden_layer_sizes``, which
```
NickleDave: remove "is", change "effects" (a plural noun) to "affects" (verb)

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+controls the number of layers and number of neurons in each layer.
+
+There are 5 values for the hyperparameter. It is varied so the neural network
+has 24 neurons but varies the network depth and the width of each layer. Two
+choices are 12 neurons in 2 layers or 6 neurons in four layers. One choice
+has 12 neurons in the first layer, 6 in the second, and 3 in third and
+fourth layers.
+
+Six other hyperparameters have to be tuned and control finding the best model,
```
NickleDave: The order of ideas in this sentence is confusing.
How about something like "The six other hyperparameters control how the algorithm finds the best model; 3 are continuous and we provide 50 values for the 3 other discrete hyperparameters"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+Hyperband's early stopping is designed to identify the highest performing model
+with minimal training. Setting ``patience`` to be high avoids interference with
+this scheme, protects against both cases above, and errs on the side of giving
+models more training time. In particular, it also provides a basic early
+stopping mechanism for the least adaptive bracket of Hyperband.
+
+Serial Simulations
+==================
+
+This section focuses on a synthetic classification example between 4 classes.
```
NickleDave: You don't say anything about why there are both serial and parallel simulations.
Pleas start this section by explaining why, e.g. "we begin with experiments where we use only serial resources, to demonstrate that our implementation of the Hyperband algorithm works as intended before we leverage the parallel-computing resources of Dask"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+
+
+.. [#random-sampling-hyperband] As much as possible – Hyperband evaluates more
+   hyperparameter values. The random search without early stopping
+   evaluates every hyperparameter value Hyperband evaluates.
+
+Parallel Experiments
+====================
+
+This section will highlight a practical use of ``HyperbandSearchCV`` and use 25
```
NickleDave: Good intro but don't miss a chance to emphasize your contribution.
"Having demonstrated that our implementation works, we now show that we can leverage the parallelization strengths of Dask, using 25 workers to speed up optimization of an image denoising algorithm"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+.. [#pytorch] https://pytorch.org
+.. [#skorch] https://github.com/skorch-dev/skorch
+
+The inputs and desired outputs are given in Figure :ref:`fig:io+est`. This is
+an especially difficult problem because the noise variance varies slightly
+between images. To protect against this, let's use a shallow neural network
+that's slightly more complex than a linear model.  This means hyperparameter
+optimization is not simple.
+
+Again, some detail is mentioned in the appendix though complete details can be
```
NickleDave: Not sure why "though" -- how about "and" instead?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+   :label:`fig:activity`
+
+
+Model architecture & Hyperparameters
+-------------------------------------
+
+To address that complexity, let's use an autoencoder. These are a type of neural
+network that reduce the dimensionality of the input before expanding to the
+original dimension. This can be thought of as a lossy compression. Let's create
+that model and the images it'll denoise:
```
NickleDave: I would avoid use of very informal contractions like this, how about:
"We create the model and images that it will denoise"

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+of the rectified linear unit (ReLU) :cite:`relu`, including the leaky ReLU
+:cite:`leaky-relu`, parametric ReLU :cite:`prelu` and exponential linear units
+(ELU) :cite:`elu`.
+
+There are 6 other hyperparameters do not influence the model architecture.
+There are 3 discrete hyperparameters (and 160 combinations of all discrete
+variables) and 3 continuous hyperparameters. These hyperparameters all control
+finding the optimal model after the architecture is fixed. These includes
+hyperparameter like the optimizer to use (stochastic gradient descent
+:cite:`bottou2010large` a.k.a SGD or Adam :cite:`adam`), initialization,
```
NickleDave: remove "a.k.a. SGD" unless you plan to use this acronym frequently throughout the paper

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+
+This hyperparameter is varied between 4 different choices, all different types
+of the rectified linear unit (ReLU) :cite:`relu`, including the leaky ReLU
+:cite:`leaky-relu`, parametric ReLU :cite:`prelu` and exponential linear units
+(ELU) :cite:`elu`.
+
+There are 6 other hyperparameters do not influence the model architecture.
+There are 3 discrete hyperparameters (and 160 combinations of all discrete
+variables) and 3 continuous hyperparameters. These hyperparameters all control
+finding the optimal model after the architecture is fixed. These includes
```
NickleDave: Why not make this the topic sentence and just cut the current topic sentence?
E.g., "There are 6 other hyperparameters that control finding the optimal model after the architecture is fixed..."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+``patience=max_iter // 3``. This is most useful for the least adaptive bracket
+of Hyperband, which trains a couple models to completion. This has a large
+impact on the time to solution with these parallel computational resources but
+doesn't have a large impact on the number of ``partial_fit`` calls.
+
+Performance
+-----------
+
+Anecdotally, ``HyperbandSearchCV`` performs well and beats manual hand-tuning
+by a considerable margin. While manually tuning, I considered any scores/negative losses of about
```
NickleDave: This could be just (another) style thing but I would say this a bit more formally, e.g.,
"The best loss values during experiments with manually tuning hyperparameters were in the range of `-0.10` and obtained scores no higher than `-0.098`. For comparision, ``HyperbandSearch`` achieved a score of ..."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+Each bracket indicates a value in the trade-off between hyperparameter and
+training time importance, and is specified by the list of tuples in the example
+above. Each bracket is specified so that the total number of ``partial_fit``
+calls is approximately the same among different brackets. Then, having many
+models requires pruning models very aggressively and vice versa with few
+models. As an example, with ``max_iter=243`` the least adaptive bracket has 5
+models and no pruning. The most adaptive bracket has 81 models and fairly
+aggressive early stopping schedule.
+
+The exact aggressiveness of the early stopping schedule depends one optional
```
NickleDave: I think you're missing an "on" here: "depends **on** one optional..."

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+   #     "module__init",  # 4 choices
+   #     "module__activation",  # 4 choices
+   #     "optimizer__lr",  # cnts
+   #     "optimizer__momentum",  # cnts
+   #     "optimizer__weight_decay"  # cnts
+   # ])
+
+Details are in the appendix.
+
+Usage: plateau specification for non-improving models
```
NickleDave, line 796: The order of presentation here is confusing to me.
Can you have `Table 1` appear before the figures, and explicitly reference it as well as thoroughly define/discuss `IncrementalSearchCV` et al. *before* you point anyone to the those figures?

papers/scott_sievert/hyperband.rst
```diff
@@ -0,0 +1,951 @@
...
+treat every model as a black box and vary the amount of data provided. This
+would not require the model to implement ``partial_fit`` and would only require
+a ``fit`` method.
+
+Another area of future work is ensuring ``IncrementalSearchCV`` and all of its
+children (including ``HyperbandSearchCV``) work well with large models.
+Modern models often consume most of GPU memory, and currently
+``IncrementalSearchCV`` requires making a copy the model. How much does this
+hurt performance and can it be avoided?
+
```
NickleDave, line 920: Maybe add one closing sentence here?
"While these areas of improvement remain, we have demonstrated that our implementation of Hyperband can make hyperparameter optimization more efficient, and specifically leverage the abilities of Dask to deal with large distributed datasets"

NickleDave:
Ok @deniederhut those are all my suggested changes.
@stsievert I think I'm the one who owes you a beer for dumping a bunch of comments on you at the last second

deniederhut:
Hi @stsievert ! Have you had a chance to respond to David's comments?

stsievert:
Thanks for the detailed and specific edits @NickleDave! It looks like they'll help improve the paper :+1:

@deniederhut I haven't responded yet because my energy has been devoted towards rerunning the experiments in the last section. When do you want a response? Should the draft also be ready by then? I *hope* to have a response and completed draft by Tuesday morning.

papers/scott_sievert/hyperband.rst
stsievert: Thanks. I've replaced "mathematical foundations" with "principled early stopping scheme".

papers/scott_sievert/hyperband.rst
stsievert: Oof. I'm not sure how that got past me.

papers/scott_sievert/hyperband.rst
stsievert: Thanks!

papers/scott_sievert/hyperband.rst
stsievert: Thanks – that's a good observation.

papers/scott_sievert/hyperband.rst
stsievert: Thanks for this catch.

deniederhut:
Depends on whether you want your paper to be available for people to read before the conference. If yes, the heard deadline is this Wed when I publish the conference-ready proceedings. If no, then... another two weeks.

stsievert:
I've pushed the changes that show the parallel experiments I ran.

I still have some edits to make. I'll try to make them by Monday.

deniederhut:
Hey Scott! How are things going with the paper?

deniederhut:
Hey @stsievert ! Is this ready for @NickleDave to re-review?

stsievert:
Yup! It took a little longer than expected but it's ready now.

deniederhut:
@NickleDave do you feel this paper is now ready for inclusion in the conference proceedings?

NickleDave:
Yes but where's my beer?!!
Liking that figure 2 @stsievert 😏 
LGTM 👍

stsievert:
The beer will have to wait until next year 😔 I had to leave early yesterday

deniederhut:
😬 Scott couldn't find you at the happy hour so he gave me your beer. It's on me in 2020 😄

PR #472, opened 2019-05-22
Paper: Expert RF Feature Extraction to Win the Army RCO AI Signal Classification Challenge

Teque5:
This pull request is for 1of2 papers from The Aerospace Corporation approved for submission to the 2019 SciPy Conference.

OTR201900733 for the paper Expert RF Feature Extraction to Win the Army RCO
AI Signal Classification Challenge

Everything built OK on my box.


deniederhut:
Hm... I wonder if the server is mad about the slash in the branch name. Let me look into this.

deniederhut:
Looks like that did the trick! You should be able to see your papers on the build server now.

thedatalass:
Paper reviewed by LKahn 5/28 and ready for conference proceedings

yingweiy:
**REVIEW COMMENTS**

In this paper, the authors presented a method to predict the radio signal modulation type and SNR, which won the Army RCO AI signal classification challenge. This paper is well written. The method is technically sound in general, and the results appear to be reproducible. The data source are clearly referenced. 

Here are my questions/suggestions:
1. Is the source code available to the readers? This is required in the review criteria.
2. Figure 1. "cv" refers to "cross validation". It is better add this notation in the figure caption.
3. Eq 1 in page 2: the log loss basically is the "cross entropy", and this needs to be properly cited.
4. The "shallow machine learning classifiers" are defined as "classifiers not build features out of the raw data". However, the deep learner as proposed also takes in the engineered features instead of deriving these features from the raw data. Is it the best practice? What if the DL algorithm directly applied to the raw data?

Teque5:
> Is the source code available to the readers? This is required in the review criteria.

Releasing our entire codebase won't be possible. Instead I've tried to show which modules we relied upon like [allantools](https://pypi.org/project/AllanTools/) and [scipy.signal](https://docs.scipy.org/doc/scipy/reference/signal.html). I've also tried to make as clear as possible how individual features were calculated in latex. Moreover, since there really isn't an 'algorithm' at work here the code is really constrained to the individual features discussed. We may include code snippets on the poster that will be shown at the scipy 2019. Unfortunately getting code cleared for public release is quite difficult due to export control.

> Figure 1. "cv" refers to "cross validation". It is better add this notation in the figure caption.

Ok no problem.

> Eq 1 in page 2: the log loss basically is the "cross entropy", and this needs to be properly cited.

Ok; I can add this as a footnote.

> The "shallow machine learning classifiers" are defined as "classifiers not build features out of the raw data". However, the deep learner as proposed also takes in the engineered features instead of deriving these features from the raw data. Is it the best practice? What if the DL algorithm directly applied to the raw data?

This is addressed in our companion paper, [available here as a separate submission](https://github.com/scipy-conference/scipy_proceedings/pull/473). The DL network produced better results than the engineering features, but combined (ensemble) yielded even better results. Moreover the engineering features based network was very robust against test data that is unlike the training data. A pure DL network stumbled in this regard.

@yingweiy If this is OK, let me know and I'll make the two changes you suggest.

bbotran:
The paper is very well written and it is easy to follow. Signal processing is not my domain and I am not up to date with latest methodologies used in the field. However, the methodology is thorough and well explained. Even without releasing the code, I would definitely accept this paper.
I found a small issue in page 2, where the formula for recall and precision is the same one. Please, correct it.

Teque5:
@bbotran The formulas look similar, but are different. I believe the current version is correct.

yingweiy:
@Teque5 Thanks for your reply. Yes, I can accept this paper with the two minor changes as suggested. Thanks!

bbotran:
> @bbotran The formulas look similar, but are different. I believe the current version is correct.

@Teque5 You are right. Ignore my comment. Congrats for your work. I recommend accepting the paper.

Teque5:
I just pushed the latest with the above changes as discussed.

deniederhut:
It looks like at least one of these papers is missing its DOI. Could I ask you to take a stroll through you bibliography and add these?

Teque5:
@deniederhut DOIs have been added/updated. Good catch.

deniederhut:
This looks good to go. Thanks so much for submitting!

PR #475, opened 2019-05-22
Paper: pyjanitor - clean APIs for data cleaning

ericmjl:


deniederhut:
Hi @ericmjl ! It looks like none of our reviewers were able to complete their reviews for your paper by the deadline, so I'll be stepping in to complete the review myself. Please expect comments from me some time in the next handful of days 🙂.

.gitignore
```diff
@@ -5,3 +5,4 @@ _build
 *.pyc
 *.swp
 *~
+.vscode/settings.json
```
deniederhut: Please don't change files outside of your paper directory.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+.. class:: abstract
+
+    The ``pandas`` library has become the de facto library for data wrangling
+    in the Python programming language. However, inconsistencies in the
+    ``pandas``  application programming interface (API), while idiomatic due to
+    historical use, prevent use of expressive, fluent programming idioms that
+    enable self-documenting ``pandas`` code. Here, we introduce ``pyjanitor``,
+    an open source Python package that extends the ``pandas`` API with
+    expressive and fluent idioms. We describe how ``pyjanitor`` reconciles
```
deniederhut: You are repeating "expressive and fluent" here.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+It is a common experience amongst data scientists that data wrangling can
+occupy up to 80% of their time :cite:`nytimes` :cite:`JSSv059i10`. Part of this
+time is spent defining modelling approaches, and part of this time is writing
+code that executes the sequence of transformations on raw data that wrangle it
+into the necessary shape for downstream modelling work.
+
+In the Python ecosystem, ``pandas`` is the *de facto* tool for data
+manipulation. As reflected by the article with a pun-laden title, "Pandas
+responsible for Python explosion", ``pandas`` was pivotal for the development
+of the Python programming language away from merely a scripting language to
```
deniederhut: Calling Python "merely a scripting language" is inaccurate, unnecessary for the argument you are making, and frankly feels inflammatory. I'm also not sure that Pandas has contributed much at all to the base Python language, unless you count the buffer protocol (which comes from NumPy). 

Perhaps what you meant to say was that Pandas provides functionality for a particular use case (tabular data manipulation) that is required for analytic work, but absent from the standard library.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+Data preprocessing, or data wrangling, is an unavoidable task in data science.
+It is a common experience amongst data scientists that data wrangling can
+occupy up to 80% of their time :cite:`nytimes` :cite:`JSSv059i10`. Part of this
+time is spent defining modelling approaches, and part of this time is writing
+code that executes the sequence of transformations on raw data that wrangle it
+into the necessary shape for downstream modelling work.
+
+In the Python ecosystem, ``pandas`` is the *de facto* tool for data
+manipulation. As reflected by the article with a pun-laden title, "Pandas
```
deniederhut: I think the citation for this article is missing. Now that I'm thinking about it, a better option might be the stack overflow developer survey from this year.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+responsible for Python explosion", ``pandas`` was pivotal for the development
+of the Python programming language away from merely a scripting language to
+one in which data science workflows could be constructed and executed.
+
+``pandas`` provided the key primitive object for data science's key modelling
```
deniederhut: I don't know that you want to make the argument that tables are "the key" primitive object (this accolade might go to arrays instead). Perhaps they are "a key" abstraction. Also, you repeat the words "key" and "modelling" here.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+.. code-block:: python
+
+    import pandas as pd
+    import numpy as np
+
+    df = pd.DataFrame(...)
+    # standardize column names.
+    df.columns = [
+        i.lower().replace(' ', '_') for we in df.columns
```
deniederhut: This doesn't seem like it will execute the way you are expecting. I would also add that you can do this without resorting to list comprehensions by using the vectorized string methods defined on the Column object itself.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+    ]
+    # remove unnecessary columns
+    del df['column_name_14']
```
deniederhut: There is also the .drop method, if you prefer functional approaches.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+    # transform a column by taking the log
+    df['column_name_13'] = np.log10(df['column_name_13'])
+    df = df[df['column_name_12'] < 3]
+    df = df.dropna()
+
+This coding style has become idiomatic over the years. However, the mixing of
+imperative and functional programming styles, and the lack of
+*informative operation names*, yields a non-fluent expression of what is
```
deniederhut: This feels like a value judgment presented with little evidence. Which names are uninformative? Why?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+intended to be accomplished. This, in turn, hampers readability, and hence
```
deniederhut: Several of your examples share syntax with both NumPy and R, and should feel at least roughly familiar to people coming from other numeric approaches. So when you say it harms readability, which group of people is it harming?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+    import pandas as pd
+    import numpy as np
+    import janitor
+
+    df = (
+        pd.DataFrame(...)
+        .clean_names()
+        .remove_column('column_name_14')
+        .transform_column('column_name_13', np.log10)
+        .query('column_name_12 < 3')
```
deniederhut: The query method is also available in Pandas, so it feels a bit disingenous to list it here as a feature provided only by pyjanitor.

papers/eric_ma/manuscript.rst
deniederhut: Repeating "fluent and expressive" here

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+        .dropna()
+    )
+
+This is the API design that ``pyjanitor`` aims to provide to ``pandas`` users.
+By using a fluent API design, ``pyjanitor`` explicitly targets a
```
deniederhut: Repeating "fluent and expressive" here.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+``pandas``-compatible API that enables data scientists to construct their data
+processing code in an expressive fashion. By providing data processing
+routines, we also save time for data scientists and engineers like ourselves,
```
deniederhut: This feels like a value judgment without evidence. How much time are you saving? How did you measure this?

The "like ourselves" feels superfluous here.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+History of ``pyjanitor``
+------------------------
+
+``pyjanitor`` started as a Python port of the R package ``janitor``, which
+provides the same functionality to R users. The initial goal was to explicitly
+copy the ``janitor`` function names while engineering it to be compatible with
+``pandas.DataFrames``, following Pythonic idioms, such as the method chaining
+provided by ``pandas``. As the project evolved, the scope broadened, to provide
+a defined and expressive DSL for data processing, centered around the DataFrame
```
deniederhut: Why is having a doman-specific language an important thing for data analysis?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+    )
+
+This is the API design that ``pyjanitor`` aims to provide to ``pandas`` users.
+By using a fluent API design, ``pyjanitor`` explicitly targets a
+``pandas``-compatible API that enables data scientists to construct their data
+processing code in an expressive fashion. By providing data processing
+routines, we also save time for data scientists and engineers like ourselves,
+allowing us to accomplish our work more efficiently.
+
+History of ``pyjanitor``
```
deniederhut, line 164: This section feels rather brief. Were you intending to speak more about the motivation? The community of practice around the library?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+------------------------
+
+``pyjanitor`` started as a Python port of the R package ``janitor``, which
+provides the same functionality to R users. The initial goal was to explicitly
+copy the ``janitor`` function names while engineering it to be compatible with
+``pandas.DataFrames``, following Pythonic idioms, such as the method chaining
+provided by ``pandas``. As the project evolved, the scope broadened, to provide
+a defined and expressive DSL for data processing, centered around the DataFrame
+object as a first-class citizen.
```
deniederhut: Someone might read this to mean that Pandas DataFrames are only given the privileges that other objects have in your library, and not in Python (e.g. https://en.wikipedia.org/wiki/First-class_function). Rewording this to say that your library provides an extension on Pandas DataFrames (and not Series, or labeled arrays) might make your intention more clear.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+Architecture
+------------
+
+``pyjanitor`` relies completely on the ``pandas`` extension API :cite:`pandas`,
```
deniederhut: It might make more sense to link to the documentation here than to repeat your citation for the library.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+intended to be accomplished. This, in turn, hampers readability, and hence
+code maintainability.
+
+In terms of API design, we might want a block of code that reads as follows:
+
+.. code-block:: python
+
+    import pandas as pd
+    import numpy as np
+    import janitor
```
deniederhut, line 134: Is there a reason that the module is named 'janitor' while the package is named 'pyjanitor'?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+    def data_cleaning_function(df, **kwargs):
+        ...
+        # data cleaning functions go here
+        ...
+        return df
+
+In order to reduce the amount of boilerplate required, ``pyjanitor`` also makes
+heavy use of ``pandas_flavor`` :cite:`pf`, which provides an easy-to-use function
+decorator that handles class method registration. As such, to extend the
+``pandas`` API with more class-method-like functions, we just have to decorate
```
deniederhut: Do you mean instance method here?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+        # data cleaning operations go here
+        ...
+        return df
+
+Underneath each data cleaning function, we are free to use both the imperative
+and functional APIs. What is exposed, though, is a functional and fluent API
+:cite:`fluent` for the end-user.
+
+Thanks to the ``pandas.DataFrame.query()`` API, symbolic evaluations are
+generally available in ``pyjanitor`` for filtering data. This enables us to
```
deniederhut: What is the benefit of using pyjanitor over using `query` directly?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+.. code-block:: python
+
+    def data_cleaning_function(df, **kwargs):
+        ...
+        # data cleaning functions go here
+        ...
+        return df
+
+In order to reduce the amount of boilerplate required, ``pyjanitor`` also makes
```
deniederhut: Are there tradeoffs here around implicitly mutating the class object? E.g. discoverability? Are there any edge cases where the normal behavior of a DataFrame is modified?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+and functional APIs. What is exposed, though, is a functional and fluent API
+:cite:`fluent` for the end-user.
+
+Thanks to the ``pandas.DataFrame.query()`` API, symbolic evaluations are
+generally available in ``pyjanitor`` for filtering data. This enables us to
+write functions that do filtering of the DataFrame using a verb that may
+match end-users' intuitions better. One such example is the
+``.filter_on('criteria')`` method, illustrated in the opening example.
+
+At the time of writing, there has been nascent development of bioinformatics,
```
deniederhut: This feels like it belongs in a 'future directions' section.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+cheminformatics, and finance submodules. This enables convenient construction
+of fluent data preprocessing workflows for each of those particular subfields.
+The dependencies required for their usage are optional at install-time, and we
+provide instructions for end-users to install the relevant packages if they are
+not already installed locally.
+
+Design
+------
+
+Inspired by the ``dplyr`` world, ``pyjanitor`` functions are named with verb
```
deniederhut: Earlier in the paper, you stated that pyjanitor was inspired by janitor. Does janitor not use verbs?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+expressions. This helps achieve the DSL-like nature of the API. Hence, if we
+want to "clean names", the end user can call on the ``.clean_names()``
+function. If the end user wants to "remove all empty rows and columns", they
+can call on ``.remove_empty()``. As far as possible, function names are
+expressed using simple English verbs that are understandable cross-culturally,
```
deniederhut: Isn't there a tradeoff here as well? Using plain English like this may risk imprecision, no? E.g. what if I have a variable called "names"? What if my row index consists of names? What if my column index is temporal offsets?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+        .clean_names(
+            preserve_original=True,
+            remove_special=True,
+            ...
+        )
+    )
+
+In order to keep the code functional, no operations that change the original
+DataFrame are allowed. Hence, if the internal implementation of a function
+results in a mutation of the original DataFrame, we explicitly make a copy of
```
deniederhut: Many Pandas operations already copy data before modifying it. How often does your copying result in a double-copy? What's the worst case complexity for performing a single transformation? Is it ever worse than 2mn?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+        df
+        .clean_names(
+            preserve_original=True,
+            remove_special=True,
+            ...
+        )
+    )
+
+In order to keep the code functional, no operations that change the original
+DataFrame are allowed. Hence, if the internal implementation of a function
```
deniederhut: How do you handle DataFrames that store objects?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+the DataFrame first.
+
+Finally, we explicitly do not allow overriding or duplicating existing
+DataFrame class methods. The goal here is to extend ``pandas``, rather than
+replace its API.
+
+Documentation
+-------------
+
+Full API Documentation for `pyjanitor` is available on ReadTheDocs
```
deniederhut: This feels like it should be a footnote, not a section.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+:cite:`docs`.
+
+An examples gallery, which contains Jupyter notebooks that showcase how to use
+``pyjanitor``, is also part of the documentation.
+
+Development
+-----------
+
+Additionally, development takes place on GitHub :cite:`github`.
```
deniederhut: I'm not sure what this sentence adds to the substance of the paper.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+The reception to ``pyjanitor`` has been encouraging thus far. Newcomer
+contributors to open source have made their maiden contributions to
```
deniederhut: "maiden" feels unnecessarily gendered.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+An examples gallery, which contains Jupyter notebooks that showcase how to use
+``pyjanitor``, is also part of the documentation.
+
+Development
+-----------
+
+Additionally, development takes place on GitHub :cite:`github`.
+
+The reception to ``pyjanitor`` has been encouraging thus far. Newcomer
```
deniederhut: This feels like a value judgment presented without evidence. How many active contributors are there? Do you have a CoC? Community guidelines?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+Documentation
+-------------
+
+Full API Documentation for `pyjanitor` is available on ReadTheDocs
+:cite:`docs`.
+
+An examples gallery, which contains Jupyter notebooks that showcase how to use
+``pyjanitor``, is also part of the documentation.
+
+Development
```
deniederhut, line 294: This section feels fairly light on details. If there's not more to say than "we have a community of developers", I might consider omitting it from the paper.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+As with most open source software development, maintenance and new feature
+development are entirely volunteer driven. Users are invited to post feature
+requests on the source repository issue tracker, but are more so invited to
+contribute an implementation themselves to share. To date, 31 contributors have
+made pull requests into ``pyjanitor``.
+
+Evolving Contributions
+----------------------
+
+In the spirit of being beginner-friendly, new contributions to the pyjanitor
```
deniederhut: I'm not sure what the point of this sentence is.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+generalize the implementation.
+
+As an example, the commit history for ``clean_names()`` follows this pattern.
+The initial implementation manually listed out every character to be replaced
+by an underscore, in a DataFrame with a single column level. A later pull
+request extended the implementation to multi-level columns, and the current
+improved version uses regex string replacement. Each of these contributions
+were made by first-time open source contributors.
+
+The presence of a robust test suite with high test coverage enables us to
```
deniederhut: This feels superfluous.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+have also been welcome. There is a salient lesson here: with open source tools,
+savvy users can help steer development in a direction that they need.
+
+As with most open source software development, maintenance and new feature
+development are entirely volunteer driven. Users are invited to post feature
+requests on the source repository issue tracker, but are more so invited to
+contribute an implementation themselves to share. To date, 31 contributors have
+made pull requests into ``pyjanitor``.
+
+Evolving Contributions
```
deniederhut: Is there a reason you have two different sections on development?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+originator" for verb-based data processing syntax. ``janitor`` the R package
+extends  ``dplyr``. It is available for use by Python users through ``rpy2``;
+however, its primary usage audience is R users.
+
+**pandas-ply** :cite:`pandas-ply`: This is a tool developed by Coursera, and
+aims to provide the ``dplyr`` syntax to ``pandas`` users. One advantage that it
+has over ``pyjanitor`` is that symbolic expressions can be used inside
+functions, which  automatically get parsed into an appropriate lambda function
+in Python. However, the number of verbs available is restricted to the
+``dplyr`` set. As of 24 November 2018, development was last seen 3 years ago,
```
deniederhut: I think more relevant than the number of contributors is a comparison with your library. Why should someone prefer pyjanitor? Does it solve a different use case? Same use case but better functionality? Same functionality but faster?

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+    df = (
+        pd.DataFrame(...)
+        .mutate(
+            expression="column_name_12 + column_name_13",
+            new_column_name="summation"
+        )
+    )
+
+As of now, because symbolic parsing is unavailable, this fluent and declarative
```
deniederhut: Have you looked into patsy?

deniederhut (changes requested):
Thanks for the paper submission! Pyjanitor certainly sounds like a useful library, and I do enjoy the chained method syntax.

I've left a few inline comments where I had something specific to say about the content of the paper. I think my general impression is that your submission tells the reader that pyjanitor is useful, but could be more convincing if it showed the reader exactly why. For instance, I feel it's a bit light on details about:

1. the capabilities of the library
2. the implementation of those capabilities
3 the benefit of your implementation over Pandas, or another Pandas interfacing library

I look forward to seeing your revisions!

ericmjl:
@deniederhut, thanks for taking time to do the review. I suspect I may have accidentally resolved a pointer on my phone, when I didn’t intend to. I’ll go hunting for the review point when back on my computer!

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+manipulation. As reflected by the article with a pun-laden title, "Pandas
+responsible for Python explosion", ``pandas`` was pivotal for the development
+of the Python programming language away from merely a scripting language to
+one in which data science workflows could be constructed and executed.
+
+``pandas`` provided the key primitive object for data science's key modelling
+tools of statistical modelling and machine learning: the ``DataFrame``.
+However, there are inconsistencies in the `pandas` application programming
+interface (API) that, while now idiomatic due to historical use, prevent use of
+expressive, fluent programming idioms that enable self-documenting data science
```
stargaser: Later in the paper, I find a reference to the 2005 post by Martin Fowler in which he coined the term "fluent interface". I haven't encountered this phrase as formal terminology before. It would be helpful to add more description of what is meant by "fluent" and why it is important, earlier in the paper. In the Wikipedia page https://en.wikipedia.org/wiki/Fluent_interface I find a succinct "with the goal of making the readability of the source code close to that of ordinary written prose, essentially creating a domain-specific language within the interface" which makes the term more understandable.

papers/eric_ma/manuscript.rst
```diff
@@ -0,0 +1,403 @@
...
+
+    @pf.register_dataframe_method
+    def data_cleaning_function(df, **kwargs):
+        ...
+        # data cleaning operations go here
+        ...
+        return df
+
+Underneath each data cleaning function, we are free to use both the imperative
+and functional APIs. What is exposed, though, is a functional and fluent API
```
stargaser: This is the part of the paper where I first realized that "fluent" meant something formal, especially by seeing a reference. I would like the term to be defined more explicitly, sooner in the paper (per comment in the Introduction).

papers/eric_ma/manuscript.rst
stargaser: I noticed this too. Perhaps "first open-source contribution".

papers/eric_ma/manuscript.rst
stargaser: I am also concerned about citing last development, and number of contributors. The information probably won't age well. One could argue that number of stars or number of forks is a better or at least equally important statistic. 

I am also left wondering: when deciding to write pyjanitor, did you take these contribution indicators as the basis for deciding to make a new package? Or is it a post-package justification for having written pyjanitor? It seems to me that the paper is mostly arguing for pyjanitor on the basis of providing a fluent API.

stargaser (commented):
@ericmjl thank you for this interesting submission. I read it before seeing @deniederhut's review comments. I agree with those and I added a few more in-line.

papers/eric_ma/manuscript.rst
ericmjl: Thanks for the improved phrasing, @deniederhut!

papers/eric_ma/manuscript.rst
ericmjl: As I have removed reference to this article.

papers/eric_ma/manuscript.rst
ericmjl: I included this section having reflected on (1) my prior experience as a beginner contributor to open source packages, and (2) the most recent sprint I led at PyCon, where I used those prior experiences to guide new contributors. There were individuals who had concerns about whether they should try to work out every single edge case for a new contribution, or to just contribute what they could first. I encouraged them to just contribute what they had first, to make sure it worked, and then we could generalize the function later (which is what other contributors did for some of my own functions, `clean_names` being the case study shown in the text.

I guess I’m trying to use the this little section of the paper to communicate to two types of readers:

1. Sprint leaders: they can and should encourage atomic contributions that are easy to say yes to.
1. Newcomer contributors should not feel burdened to make comprehensive contributions. Atomic, easy-yes-type contributions are more than welcome!

That said, if this seems out of scope for a Proceedings paper, I am happy to take the entire section out and move it to a blog post of some kind on my own blog. Having experienced first-hand the joys of leading a sprint with many first-timer contributions, I thought this section of the manuscript might be one more outlet to nudge our community in a better direction.

papers/eric_ma/manuscript.rst
ericmjl: Agreed. I have removed it from the manuscript.

papers/eric_ma/manuscript.rst
ericmjl: In retrospect, it looks like the two sections should have been merged. Either that, or I need sub-section headers. I am more inclined towards the former, but open to change. Will push; please review the new style.

papers/eric_ma/manuscript.rst
ericmjl: > when deciding to write pyjanitor, did you take these contribution indicators as the basis for deciding to make a new package?

To address this question, the answer is less clear than some might assume. I made the first prototype port of pyjanitor on a whim having seen janitor from my colleague. That said, having:

1. Searched and not finding any *actively developed* packages with a large community of contributors that provided the fluent API,
2. Encountered further use cases at work where a fluent API could help with maintainability of my code (e.g. repeated copy/pasta’d code that should have just been wrapped in a chain-able function)
3. Found that the custom code I was writing may be generally useful, but not necessarily aligning well with the existing dplyr verbs (e.g. code for getting `(X, y)` matrices for ML purposes),

I decided that these were reasons enough to develop a new library that covered these use cases.

I’m glad for the question, @stargaser, because that prompted also an answer to @deniederhut’s questions. In the current state, I’m not sure this particular section conveys the ideas I just listed above. Perhaps the section needs to be reworked? What would be your advised course of action here?

papers/eric_ma/manuscript.rst
ericmjl: Thanks for the syntax catch.

Regarding vectorized string methods, the syntax would still be imperative rather than functional/chained, wouldn’t that be right? Either way, the chaining would make for a more readable block of code, wouldn’t it?

papers/eric_ma/manuscript.rst
ericmjl: @deniederhut I think it’s less about the names, and more about how the code reads as a whole. A series of imperative statements reads less naturally than a series of verbs with documented meaning. At the same time, I’m not aware of user studies that have been conducted to bolster my aforementioned claim, though I can go digging around to see if there are any.

At this moment in time, I’m at a bit of a loss on how to rephrase. Might you have some suggestions? If not, I’ll need to take some time to think about this.

papers/eric_ma/manuscript.rst
ericmjl: I hope this is a misunderstanding on my part; the claim here is on hampering, not harming, which I think is different in a nuanced way. My thoughts here are that reading off a block of verbs is much more makes it easier to identify the sequence of manipulations, compared to reading off a block of imperative statements. Hampering, in this context, would mean “not making as efficient as could be”.  If we have a sequence of verbs we’re reading off, it’s like reading a data transformation story: “clean_names, then remove_column, then transform_column with log10, then query for values in column_name_12 that are less than 3, then drop all nulls.” This is harder to grok from the sequence of imperative statements that have become idiomatic in the `pandas` world.

papers/eric_ma/manuscript.rst
ericmjl: I disagree that this is disingenuous. The goal here is to show that the APIs that pyjanitor provides are mix-and-matchable with pandas’ method chainable syntax.

ericmjl: If need be, I am happy to annotate which are pyjanitor APIs, and which are pandas APIs.

papers/eric_ma/manuscript.rst
ericmjl: The use of “non-fluent expression” is intended to contrast against “fluent and expressive”.

papers/eric_ma/manuscript.rst
ericmjl: > How much time are you saving? How did you measure this?

I took it to be quite obvious that by providing a function library that wraps multiple lines of code into a single function call, we would be saving users time, by definition: no need to memorize multiple lines, less typing, etc. I’m not sure it’s necessary to back this statement up?

papers/eric_ma/manuscript.rst
ericmjl: I think I was thinking about sub-fields, e.g. the worlds of bioinformatics and cheminformatics that I touch,  and the world of finance that some of pyjanitor users have encountered. For example, no finance user would do a `join_fasta` (unless they were, perhaps, analyzing the science behind a biotech firm), but this function exists to help Bioinformaticians wrangle their FASTA files into a data frame. Would it be worth expanding on this point somewhere else in the paper?

papers/eric_ma/manuscript.rst
ericmjl: > Were you intending to speak more about the motivation?

Yes, but I was not sure whether this would be appropriate in the Proceedings paper. What is your opinion on this?

> The community of practice around the library?

I’m not quite sure what one would write about re: community practice. Could you enlighten me on what you’re thinking about?

papers/eric_ma/manuscript.rst
ericmjl: Thanks for the better wording, you grokked my intent here correctly.

papers/eric_ma/manuscript.rst
ericmjl: Yes. I wanted to name it `janitor`, consistent with the R package,  but that name was already taken on PyPI. Hence, `janitor` is the intended name, but `pyjanitor` is the name used on PyPI out of convenience.

papers/eric_ma/manuscript.rst
ericmjl: Yes, thank you for the note on this.

papers/eric_ma/manuscript.rst
ericmjl: `filter_XYZ` is provided as a more specific version of `query`; some users (including myself) think of “I want to filter on column X based on this condition” rather than “I would like to query data frame for column X values of some condition”. It is hence provided for convenience.

ericmjl: In the paper, we mention the phrase “may match end-users’ intuitions better” to express this idea.

papers/eric_ma/manuscript.rst
ericmjl: Yes, agreed.

ericmjl: I will remove this section, actually, because it overlaps with the “Extensions beyond `pyjanitor`” section.

papers/eric_ma/manuscript.rst
ericmjl: Yes, it does, and we consider `janitor` to be part of the `dplyr` world.

papers/eric_ma/manuscript.rst
ericmjl: > Using plain English like this may risk imprecision, no?

Yes, it may, if the documentation were not also available on what each function does precisely. On the other hand, this documentation does exist, so I think we are in an okay regime here.

ericmjl: Once again, my experiences with trying to be inviting and inclusive for contributors is the main motivation for putting this statement here.

papers/eric_ma/manuscript.rst
ericmjl: We generally try to avoid double-copying in the internal implementation of each function. Though it may happen by accident in the future, I have not yet accepted a PR that results in double-copying on a function.

There was a long thread we had on our issue tracker regarding whether or not to mutate the original data frame; we ended up deciding not to, in the interest of following pandas development trajectory (they also are moving away from mutation in-place).

ericmjl: As for worst-case complexity: using `transform_column` as an example: we take advantage of `df[‘column_name’].apply(np.log10)`, which only does a single pass over the DataFrame. 

Side note, as you can see, `transform_column` can be thought of as a convenience wrapper for `df[‘column_name’].apply(func)`.

papers/eric_ma/manuscript.rst
ericmjl: Functions must be able to operate on those objects. Otherwise, those functions’ error messages will be raised.

papers/eric_ma/manuscript.rst
ericmjl: I’m trying to think of the counterfactual here: if “first contribution” was masculine, would that be less or more inclusive?

On a broader note, “maiden” something is a widespread and common use of English, for which I think it is unnecessary to de-gender. That said, if the principle here is not “gendered” issues but “issues of accessibility to non-English readers”, I am happy to change this to “first open source contribution”, since “maiden contribution” is indeed *not* idiomatic for non-native speakers of English.

papers/eric_ma/manuscript.rst
ericmjl: Apologies here, but I’m not grokking: How is it a value judgement? 

In terms of active contributors:

- Four of us, unfortunately all guys. I’m actively on the look-out for minority contributors whom we can partner with to sustain the package.
- No CoC explicitly at the moment, but it doesn’t exist only because I have yet found the time to put one in. The one in my head matches the SciPy/PyCon CoC very closely.
- Community guidelines: these are alluded to in the README, I think.

papers/eric_ma/manuscript.rst
ericmjl: Yes, thanks for pointing this out, @stargaser; the current omission of this point is this paper’s main weakness, IMO. This will be addressed.

papers/eric_ma/manuscript.rst
deniederhut: Hm. One way to make this compelling might be to describe a series of transformations in english, then show:

1. the transformations written in pyjanitor; and,
2. the transformations written in pandas; where

the hope is that the pyanitor syntax will look a lot more like the english than pandas does.

This also might be a good place to pull in "Readability counts" from the zen of python.

papers/eric_ma/manuscript.rst
deniederhut: I think I meant that if you are intending to show that pyjanitor is less imperative than pandas, the fair comparison is to use pandas's most functional syntax for the comparison.

papers/eric_ma/manuscript.rst
deniederhut: So when I hear DSL, what comes to mind is something like SQL or postscript. I didn't even think to make the connection with something like `join_fasta` and biology. 

Yeah, it might be good to talk more about how you are simplifying biotech specifically with special functions in pyjanitor. I seem to recall this is mentioned in a paragraph somewhere already?

papers/eric_ma/manuscript.rst
deniederhut: Sure, so you are drawing explicit parallels between your library and dplyr. One of the things that made Wickham's work successful was a large community of people who used these, and were in regular contact both with each other and with Wickham (the tidyverse). When writing new libraries, for example, he has sourced names for functions from the community (sometimes in public on Twitter). So the software he develops is close to what people actually do in R - this ongoing discussion around "what people actually do" is the community of practice.

papers/eric_ma/manuscript.rst
deniederhut: > already taken on PyPI

☹️  that does happen

papers/eric_ma/manuscript.rst
deniederhut: Awesome. Could you include this in the paper? It feels like an important (and interesting) implementation detail.

papers/eric_ma/manuscript.rst
deniederhut: I'm thinking about dataframes storing references to mutable Python objects, and avoiding mutating those. Come to think of it, I don't remember whether Pandas does a deepcopy on PyObjects first?

papers/eric_ma/manuscript.rst
deniederhut: Awesome. Could I ask you to put those examples in the paper?

papers/eric_ma/manuscript.rst
deniederhut: Hm. What about this: we modify the abstract/introduction to state that the theses of this paper are twofold. 

1. To describe the design, implementation, and practical use of pyjanitor re: code fluency
2. To enumerate the methods by which pyjanitor has been successful in including first-time contributors

papers/eric_ma/manuscript.rst
ericmjl: Understood, I get this point. Will modify.

papers/eric_ma/manuscript.rst
stargaser: I suggest including these three points at the outset, and perhaps renaming the section to "Comparison with related packages". Of course, your reasons applied at the time you decided to develop pyjanitor; circumstances can change and competing packages can arise, that would lead one to develop the package differently if starting today. (Example: astropy.table was developed before pandas burst on the scene; nowadays Astropy devs like me are asked, why didn't you just base astropy.table on pandas? And then we have to explain the history...)

Maybe after acknowledging that the landscape today may be different than it was when first developing pyjanitor, you could list the other packages with more emphasis on functionality / fluency, and less emphasis on the time of last development.

papers/eric_ma/manuscript.rst
ericmjl: Yes, indeed, near the end of the paper.

As time has progressed, so has the package, and as such I think I will elaborate more on the domain-specific functionality that has been added.

papers/eric_ma/manuscript.rst
ericmjl: Done.

papers/eric_ma/manuscript.rst
ericmjl: I'm not sure myself either, but it's not a case we have had to worry about very much, at most in the cheminformatics module. Going to mark this as resolved for now, but happy to come back to it later.

papers/eric_ma/manuscript.rst
ericmjl: This has been reworked.

papers/eric_ma/manuscript.rst
ericmjl: Perhaps ending with time of last development for _every_ package led to that masking over the functionality/fluency points listed. I have removed the statements on time of last development, and instead replaced it with a summary.

deniederhut:
@ericmjl we're a bit past the deadline here -- is this paper ready for @stargaser and I to take a second look?

ericmjl:
Oh my, I'm really sorry about the delay. I'm trying to squeeze out time for the final change - addressing @stargaser's comment on fluent programming. Will try to be done by tomorrow (Saturday) EOD. Apologies again for the delay here, work has been heavy-load recently, for some reason.

deniederhut:
> work has been heavy-load recently 

I sympathize 😅

papers/eric_ma/manuscript.rst
ericmjl: Not yet, thanks for the suggestion!

papers/eric_ma/manuscript.rst
ericmjl: We try to not mutate the normal behaviour of a DataFrame (i.e. don’t override existing attributes). pandas-flavor helps us discover, at runtime, whether we have clobbered an existing attribute. Come to think of it, this technical point should go into the paper.

papers/eric_ma/manuscript.rst
ericmjl: Going to resolve this for the time being. I am finding it challenging at this point in time to see how to change the paper.

ericmjl:
@deniederhut, @stargaser thank you for your review comments.

Within the constraints of time, I’ve done my best to address your comments. Nonetheless, I am happy to continue editing where necessary.

You will notice lots of modifications to the plain text organization - it is because I used the switch to [semantic line breaks](https://sembr.org/) as an excuse to go over the text with a fine-toothed comb, hopefully catching and correcting minor inconsistencies in formatting. Hope you don’t mind, it was borne out of me trying to be conscientious with the little things.

I have done my best to address the comment on fluent programming - namely, introducing it in a bit more detail earlier. I tried a few ways, such as adding a paragraph, and rewriting the introduction from scratch, but in the end, a footnote that described what fluent programming aims to achieve seemed to be the best way to explain fluent interfaces without disrupting the flow. Nonetheless, I’m still open to suggestions on how to change this for the purpose of clarity.

Looking forward to your further comments.

papers/eric_ma/manuscript.rst
```diff
@@ -45,36 +46,49 @@ Introduction
...
+This API was noticeably missing from the Python standard library and NumPy, 
+which, prior to ``pandas`` emergence, 
+were the primary tools for data analysis in Python.
+Hence, through the ``DataFrame`` object and its interfaces, 
+``pandas`` provided a key API 
+that enabled statisticians, data scientists, and machine learners 
+to wrangle their data into a usable shape. 
+That said, there are inconsistencies in the `pandas` API
+which, though now are idiomatic due to historical use, 
+prevent the use of expressive, fluent :cite:`fluent` programming idioms [#fluent]
```
deniederhut: I don't think this footnote is behaving the way you would like.

papers/eric_ma/manuscript.rst
```diff
@@ -84,24 +98,29 @@ To do this with the pandas API, one might write the following code.
     import numpy as np
 
     df = pd.DataFrame(...)
+
     # standardize column names.
     df.columns = [
```
deniederhut: I don't think that deliberately choosing the imperative syntax from Pandas makes a fair comparison against method chaining in Pyjanitor. If you want this to be apples-to-apples (instead of comparing the best example of one library against the worst from another), you would want to show something similar to:

```python
(df
  .rename(...)
  .assign(...)
  .drop(...)
  .query(...)
  .dropna(...)
  .groupby(...)
  .mean(...)
  .join(...)
)
```

for the Pandas case.

papers/eric_ma/manuscript.rst
```diff
@@ -169,40 +208,50 @@ the custom function, as illustrated in the following code sample:
...
-
-Where domain-specific verbs are used, we strive to match the
-mental models and vocabulary of domain experts. One example comes from the
-``biology`` submodule, where the ``join_fasta`` function allows a
-bioinformatics-oriented user to add in a column of sequences based on FASTA
-accession numbers that are keys for sequence values in a FASTA-formatted file
-:cite:`fasta`.
+Inspired by the ``dplyr`` world, 
+``pyjanitor`` functions are named with verb expressions. 
+This as mentioned earlier, this helps with readability.
```
deniederhut: This, as mentioned earlier, helps with readability

papers/eric_ma/manuscript.rst
```diff
@@ -169,40 +208,50 @@ the custom function, as illustrated in the following code sample:
...
+Hence, if we want to "clean names", 
+the end user can call on the ``.clean_names()`` function/class method. 
+If the end user wants to "remove all empty rows and columns", 
+they can call on ``.remove_empty()``. 
+As far as possible, function names are expressed using simple English verbs 
+that are understandable cross-culturally
+and well-documented, 
+to ensure that this API is inclusive and accessible
```
deniederhut, line 247: 👍👍👍

papers/eric_ma/manuscript.rst
```diff
@@ -215,116 +264,143 @@ before cleaning, one can add the ``preserve_original`` keyword argument to the
...
 
-In order to keep the code functional, no operations that change the original
-DataFrame are allowed. Hence, if the internal implementation of a function
-results in a mutation of the original DataFrame, we explicitly make a copy of
-the DataFrame first.
+In order to adhere to a functional programming paradigm, 
+no operations that change the original DataFrame are allowed. 
+Hence, if the internal implementation of a function results in a mutation of the original DataFrame, 
+we explicitly make a copy of the DataFrame first, 
+though we also generally try to avoid double-copying as well.
```
deniederhut, line 276: 👍👍👍

papers/eric_ma/manuscript.rst
```diff
@@ -215,116 +264,143 @@ before cleaning, one can add the ``preserve_original`` keyword argument to the
...
-extends  ``dplyr``. It is available for use by Python users through ``rpy2``;
+The initial implementation manually listed out every character to be replaced by an underscore, 
+in a DataFrame with a single column level. 
+A later pull request extended the implementation to multi-level columns, 
+and the current improved version uses regex string replacement
+to concisely express the cleaning operation. 
+Most notably, each of these contributions were made by first-time open source contributors.
+
+For the long-term health of the package, 
+we are on the lookout for minority contributors who would like to help maintain the package as well.
```
deniederhut: Not to be too pedantic, but could we replace "minority" with "underrepresented" here?

papers/eric_ma/manuscript.rst
```diff
@@ -215,116 +264,143 @@ before cleaning, one can add the ``preserve_original`` keyword argument to the
...
+A later pull request extended the implementation to multi-level columns, 
+and the current improved version uses regex string replacement
+to concisely express the cleaning operation. 
+Most notably, each of these contributions were made by first-time open source contributors.
+
+For the long-term health of the package, 
+we are on the lookout for minority contributors who would like to help maintain the package as well. 
+A code of conduct document, 
+and a community guidelines document, 
+are also on our development roadmap.
```
deniederhut, line 333: Awesome. This is one of the things that [ROpenSci requries during package review](https://devguide.ropensci.org/collaboration.html#friendlyfiles) and I think it's a really good idea.

deniederhut (changes requested):
Thanks for your work on this! I think your main arguments have become more clear. I have a few last change requests, some typo fixes, and a build issue in the comments.

Oh, also this is showing up in the rendered PDF:

```
 system-message
WARNING/2 in <string>, line 96
Explicit markup ends without a blank line; unexpected unindent.backrefs:
```

papers/eric_ma/manuscript.rst
ericmjl: I have to admit to not really trying the first review partially because the intent behind the example was to show what was idiomatic, rather than what was already built-in functional - i.e. what the community knows how to do might not necessarily be the best way. In other words, the example is intentionally designed to show how a package user might attempt the problem, having maybe done one or two searches using StackOverflow.

However, since you’ve insisted :smile:, I have put in a more hearty attempt at chaining. There is one place where jumping out of the chain looks unavoidable - that is, the groupby-aggregation-joinback step. Once again, having tried multiple ways, I can’t see a way to get this to work without first jumping out of the chain, creating the necessary series, then joining it back into the original dataframe, with the original dataframe being the left-side frame.

papers/eric_ma/manuscript.rst
ericmjl: I think it’s resolved now.

papers/eric_ma/manuscript.rst
ericmjl: Resolving for now, happy to continue revisiting this.

stargaser (approved):
I am happy with the revisions addressing:
* defining fluent - the footnote is fine
* the discussion of related tools

deniederhut:
This looks good to me! Thanks everyone!

deniederhut (approved):

Issue #355, opened 2018-03-11
ORCID support


mpacer:
We should have some way for authors to designate themselves with an ORCID that we link to in the pdf and the html, this will require changing the templates as well as the information that we extract from the rst's header.

@FatihAkici This is a more complicated feature to implement than in #353, so you should tackle that one first. Completing this should give you a better idea about the templating system and the way that metadata is collected from the documents and stored.

mepa:
@stargaser, @deniederhut, should this issue be closed now that MR #532 has implemented the capability for authors to link their ORCID in the html? Or is this waiting on linking in the pdf also?

deniederhut:
Good question - I don't remember if there were links on the HTML renders or not.

Related question, was this also implemented for slides and posters? No one submitted an orcid with their slides, but that doesn't mean that it wasn't possible. 
ivan-marroquin <notifications@github.com> wrote:
“@stargaser, @deniederhut, should this issue be closed now that MR #532 has implemented the capability for authors to link their ORCID in the html? Or is this waiting on linking in the pdf also?”

stargaser:
For the papers, as I recall, we decided not to implement linking in the PDFs, and that a link on the HTML page for each paper was sufficient.

I don't know if the ORCID infrastructure will work for slides and posters.

deniederhut:
Okay so it sounds like there is still work to do on this ticket for the slides and posters.
ivan-marroquin <notifications@github.com> wrote:
“For the papers, as I recall, we decided not to implement linking in the PDFs, and that a link on the HTML page for each paper was sufficient.

I don't know if the ORCID infrastructure will work for slides and posters.”

PR #513, opened 2019-07-16
Poster for The PyData Weaver

henrykironde:
Submitting Poster pdf


presentations/posters/pydataweaver/info.json
```diff
@@ -0,0 +1,27 @@
+{
+  "authors": [
+    {
+      "name": "Henry Senyondo",
+      "orcid": "0000-0001-7105-5808",
```
deniederhut: Sorry, we don't currently support orcid (see https://github.com/scipy-conference/scipy_proceedings/issues/355). Could I ask you to update your info.json so that it looks like the example?

deniederhut (changes requested):

PR #514, opened 2019-07-16
Pydataweaver lightning

henrykironde:
Submitting  lightning slides


presentations/posters/pydataweaver/info.json
```diff
@@ -0,0 +1,27 @@
+{
+  "authors": [
+    {
+      "name": "Henry Senyondo",
```
deniederhut: Could I ask you to update this so it matches the example info.json?

deniederhut (changes requested):

PR #515, opened 2019-07-17
Fix grant number.

bdice:
@deniederhut Hi, we noticed a mistake in a grant number on our paper. Would it be possible to correct this before the final proceedings are published? Thanks!


deniederhut (approved):
👍

PR #527, opened 2020-01-24
Update README and scipy_proc.json for 2020

### LINKED ISSUE #525: Update README and scipy_proc.json for 2020 conference
The dates of the conference and deadlines will need to be updated everywhere. In scipy_proc.json, we can update the conference chair names and affiliations, the conference titles, and the volume number. Everything else will need to wait until after the conference.

mepa:
Updates dates, names, and volume number for 2020 conference.

~Note that some names and affiliations will need to be filled and/or updated sometime closer to the conference dates.~

Updates organizer names and affiliations to match https://www.scipy2020.scipy.org/organisers.

Closes #525

### DIFF
--- a/scipy_proc.json
+++ b/scipy_proc.json
@@ -19,38 +19,38 @@
     },
     "title": {
       "conference": "Python in Science Conference",
-      "full": "Proceedings of the 18th Python in Science Conference",
-      "short": "PROC. OF THE 18th PYTHON IN SCIENCE CONF. (SCIPY 2019)",
+      "full": "Proceedings of the 19th Python in Science Conference",
+      "short": "PROC. OF THE 19th PYTHON IN SCIENCE CONF. (SCIPY 2020)",
       "acronym": "SciPy",
-      "ordinal": "18th"
+      "ordinal": "19th"
     },
-    "citation_key": "proc-scipy-2019",
+    "citation_key": "proc-scipy-2020",
     "editor_email": [
+      "agarwal.meghann@gmail.com",
       "cbc@unc.edu",
-      "dalippa@gmail.com",
       "dillon.niederhut@gmail.com",
       "dave.shupe@gmail.com"
     ],
-    "dates": "July 8 - July 14",
+    "dates": "July 6 - July 12",
     "editor": [
+      "Meghann Agarwal",
       "Chris Calloway",
-      "David Lippa",
       "Dillon Niederhut",
       "David Shupe"
     ],
     "location": "Austin, Texas",
-    "volume": 4,
+    "volume": 5,
     "number": 1,
     "isbn": "noisbn",
-    "year": "2019"
+    "year": "2020"
   },
   "organization": [
     {
       "name": "Conference Chairs",
       "members": [
         {
-          "org": "Arizona State University",
-          "name": "Serge Rey"
+          "org": "",
+          "name": "Jonathan Guyer"
         },
         {
           "org": "Enthought, Inc.",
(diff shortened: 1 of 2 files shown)


scipy_proc.json
```diff
@@ -19,38 +19,38 @@
...
+    "year": "2020"
   },
   "organization": [
     {
       "name": "Conference Chairs",
       "members": [
         {
-          "org": "Arizona State University",
-          "name": "Serge Rey"
+          "org": "",
```
deniederhut: You can find the organizers' affiliations here: https://www.scipy2020.scipy.org/organisers

scipy_proc.json
```diff
@@ -117,17 +129,17 @@
...
           "org": "Argonne National Laboratory",
           "name": "Scott Collis"
         },
         {
           "org": "Novartis Institutes for Biomedical Research",
           "name": "Eric Ma"
+        },
+        {
+          "org": "",
+          "name": ""
```
deniederhut: Are we missing Nadia here?

scipy_proc.json
```diff
@@ -154,45 +166,49 @@
         {
           "org": "University Corporation for Atmospheric Research",
           "name": "Ryan May"
-        },
-        {
```
deniederhut, line 158: Are we missing Juan and Dharhas?

scipy_proc.json
```diff
@@ -154,45 +166,49 @@
...
         {
-          "org": "University of Illinois",
-          "name": "Nathan Goldbaum"
+          "org": "IBM Research Africa",
+          "name": "Celia Cintas"
         },
         {
           "org": "Federal University of Santa Catarina",
-          "name": "Melissa Mendonca"
+          "name": "Melissa Weber Mendonca"
```
deniederhut, line 189: 👍 thanks for catching this!

scipy_proc.json
```diff
@@ -219,107 +235,18 @@
...
+        {
+          "org": "Enthought, Inc.",
+          "name": "Kristen Leiser"
         }
       ]
     },
     {
       "name": "Proceedings Reviewers",
       "members": [
         {
```
deniederhut, line 256: Hm... we might need this to be an empty list for now. I'm worried that a list with an empty object might break the builds with a KeyError somewhere.

deniederhut:
Thanks for taking a stab at this! I've left a few suggestions for changes, but I think one piece of information you might be missing (because I did not think to tell you 😬) is that the organizers are also listed on the main conference page -- https://www.scipy2020.scipy.org/organisers

scipy_proc.json
mepa: The dict isn't empty (please see below), so I think it is fine to leave as is?

scipy_proc.json
```diff
@@ -219,107 +235,18 @@
...
-        },
-        {
-          "name": "Tania Lorido Botran"
-        },
-        {
-          "name": "Yingwei Yu"
-        },
-        {
-          "name": "Zach Golkhou"
+          "name": ""
```
mepa, line 257: @deniederhut, the list contains a dict with `key: value` of `"name": ""` (way down here in the diff!)

mepa:
Thanks, @deniederhut! I updated the organizer names and affiliations from the main conference page (I was referring to a Google doc and somehow didn't think to check the page :-D)

I left the Proceedings Reviewers section same since I think it should work fine as is for the build (I haven't tested explicitly, though I should figure out how to test that). But please lmk if you think otherwise.

scipy_proc.json
deniederhut: Hmm yeah I'm not sure what I was looking at

deniederhut (approved):
Thanks Meghann!

Issue #375, opened 2018-05-05
Update instructions-for-authors to include "who can be an author"


deniederhut:
[Instructions for authors](https://github.com/scipy-conference/scipy_proceedings#instructions-for-authors) should be updated to include information on group authorship. Specifically, to inform authors that we:

1. Allow projects to be listed as authors; and,
2. The project can be the sole author, first author, or affiliation of the authors; however, 
3. The corresponding author email must belong to a specific person

Issue prompted by discussion in #374.

mpacer:
I think that there should be able to be more than one corresponding author (as suggested by @eseiver) but our current build server does not support this well. I'll open an issue about this separately, but I want to modify 3 to read "The corresponding author(s) email(s) must belong to specific people."

PR #528, opened 2020-02-22
ENH: adds authorship guidelines

deniederhut:
Closes https://github.com/scipy-conference/scipy_proceedings/issues/375

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ have final say in whether to accept or reject a paper.
   your PR needs to have a title that begins with "Paper:". If you do not do
   this, the co-chairs will change your title on your behalf.
 - Authors may include a project or consortium (e.g. [The Jupyter Project](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2018/papers/project_jupyter/paper.rst))
-- There must be at least one corresponding author, and this must be a specific person
+- There must be at least one corresponding author, and this must be a specific person with a valid email address
 - All citations that have DOIs should include those DOIs in the paper's
   references section, see [`mybib.bib`](./papers/00_bibderwalt/mybib.bib).
 - All figures and tables should have captions.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -172,6 +172,8 @@ have final say in whether to accept or reject a paper.
 - For your paper to be found by the build system at http://procbuild.scipy.org
   your PR needs to have a title that begins with "Paper:". If you do not do
   this, the co-chairs will change your title on your behalf.
+- Authors may include a project or consortium (e.g. [The Jupyter Project](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2018/papers/project_jupyter/paper.rst))
+- There must be at least one corresponding author, and this must be a specific person
```
stargaser: perhaps add "with a valid email address"?

README.md
```diff
@@ -172,6 +172,8 @@ have final say in whether to accept or reject a paper.
 - For your paper to be found by the build system at http://procbuild.scipy.org
   your PR needs to have a title that begins with "Paper:". If you do not do
   this, the co-chairs will change your title on your behalf.
+- Authors may include a project or consortium (e.g. [The Jupyter Project](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2018/papers/project_jupyter/paper.rst))
```
stargaser, line 175: This is a good idea. I feel like these will be footnotes, and that perhaps we should say so? Unless this is intended to be something else, like an extra affiliation for each author.

stargaser (commented):
These seem like good additions. I'm wonder if we need to add a little more detail, per the inline comments.

README.md
deniederhut: The idea is to be able to list a consortium as an author. I believe Jupyter, for example, lists the project as first author.

README.md
deniederhut: Done!

PR #532, opened 2020-04-03
Add ORCIDs to papers

stargaser:
Add ORCIDs to papers.

TODO:
- [x] Add orcids to example papers
- [x] Make html links out of each ORCID

### DIFF
--- a/publisher/writer/__init__.py
+++ b/publisher/writer/__init__.py
@@ -37,7 +37,7 @@ class Translator(LaTeXTranslator):
         self.author_institutions = []
         self.author_institution_map = dict()
         self.author_emails = []
-        self.author_orcids = []
+        self.author_orcid_map = dict()
         self.corresponding = []
         self.equal_contributors = []
         self.paper_title = ''
@@ -70,6 +70,7 @@ class Translator(LaTeXTranslator):
     def visit_author(self, node):
         self.author_names.append(self.encode(node.astext()))
         self.author_institution_map[self.author_names[-1]] = []
+        self.author_orcid_map[self.author_names[-1]] = ''
         raise nodes.SkipNode
 
     def depart_author(self, node):
@@ -94,7 +95,7 @@ class Translator(LaTeXTranslator):
         if self.current_field == 'email':
             self.author_emails.append(text)
         elif self.current_field == 'orcid':
-            self.author_orcids.append(text)
+            self.author_orcid_map[self.author_names[-1]] = text
         elif self.current_field == 'corresponding':
             self.corresponding.append(self.author_names[-1])
         elif self.current_field == 'equal-contributor':
@@ -260,7 +261,7 @@ class Translator(LaTeXTranslator):
                                'author_email': self.author_emails,
                                'author_institution': self.author_institutions,
                                'author_institution_map' : self.author_institution_map,
-                               'author_orcids': self.author_orcids,
+                               'author_orcid_map': self.author_orcid_map,
                                'abstract': self.abstract_text,
                                'keywords': self.keywords,
                                'copyright_holder': copyright_holder,
--- a/publisher/writer/__init__.py
+++ b/publisher/writer/__init__.py
@@ -94,6 +94,7 @@ class Translator(LaTeXTranslator):
         if self.current_field == 'email':
             self.author_emails.append(text)
         elif self.current_field == 'orcid':
+            self.validate_orcid(text)
             self.author_orcid_map[self.author_names[-1]] = text
         elif self.current_field == 'corresponding':
             self.corresponding.append(self.author_names[-1])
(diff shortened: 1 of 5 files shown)


publisher/writer/__init__.py
```diff
@@ -92,6 +93,8 @@ def visit_field_body(self, node):
 
         if self.current_field == 'email':
             self.author_emails.append(text)
+        elif self.current_field == 'orcid':
+            self.author_orcids.append(text)
```
deniederhut: What happens if only some authors on a paper have an ORCID? Will we end up assigning the wrong ID to any of them?

stargaser: Indeed, this will be a problem. I changed the list to a map. An alternative would be to include a null value in the list, but this doesn't look like it can be easily coded. Author emails are treated as a list but perhaps that is required for every author?

stargaser:
@deniederhut I've got a first addition of ORCIDs to the article BibTeX entry, but it is a dictionary. I was trying to make a comma-separated list in the proper order but I haven't hit upon the right coding solution for that. What I've got now is:
```
@InProceedings{ 00_vanderwalt-proc-scipy-2020,
  author    = { {G}aius {C}aesar and {M}ark {A}nthony and {J}arrod {M}illman and {B}rutus },
  title     = { {A} {N}umerical {P}erspective to {T}erraforming a {D}esert },
  booktitle = { {P}roceedings of the 19th {P}ython in {S}cience {C}onference },
  pages     = { 1 - 3 },
  year      = { 2020 },
  editor    = { {M}eghann {A}garwal and {C}hris {C}alloway and {D}illon {N}iederhut and {D}avid {S}hupe },
  orcids    = { {'Mark Anthony': '0101-0101-0101-0102', 'Jarrod Millman': '0101-0101-0101-0103', 'Brutus': '', 'Gaius Caesar': '0101-0101-0101-0101'} },
  doi       = {  }
}
```
Do you have a preferred format for the orcids entry?

publisher/writer/__init__.py
deniederhut: 😅 or we've been listing the wrong emails under author names

deniederhut:
Hm. I've done a bit of googling and I'm not finding any examples of people linking ORCIDs to authors in the bibfile -- it's all in the text document as hyperrefs. Is that the standard practice?

stargaser:
Having done some more reading up, mostly from orcid.org, it seems to me that publishers providing ORCIDs in an official way, to allow publications to be associated with an ORCID automatically, requires using the ORCID API and having authors authenticate. That's certainly beyond what we can do for SciPy 2020. We could have the hyperlinks in the PDF to each author's ORCID record but that would be kind of unofficial since the author would be reporting their own ORCID.

stargaser:
The issue #355 does request just the self-reported ORCIDs and a way to display them in the PDF and html, so we could do that. From my research into .bib files, it is permitted to add whatever entries one likes beyond the mandatory items, so we could leave the map of ORCIDs in there for anyone who clicks through to the .bib for each article.

stargaser:
@mepa @deniederhut If we followed what Journal of Open Source Software displays, we could display the ORCID as a hyperlink on the HTML page for the proceedings papers, e.g. [this example](http://conference.scipy.org/proceedings/scipy2019/aero_nn_vila.html). What do you think?

mepa:
@stargaser @deniederhut That sounds reasonable to me to display the ORCID as a hyperlink on the HTML page for the proceedings papers, similar to the JOSS example. 

Regarding using the ORCID API, I agree that it's beyond what we can do for SciPy 2020. We could  manually check the self-reported ORCIDs for this year's proceedings, and add an issue to update our approach to adhere to [ORCID best practices](https://orcid.org/organizations/publishers/best-practices). This at least allows authors the option to include their ORCIDs this year. Wdyt?

deniederhut:
Hey @stargaser ! I generated a crossref submission file for the test paper with DOIs and it looks like it passes the schema validation. Yay! Thanks for your work on this.

So if I remember correctly, the only remaining effort for this PR is to remove the ORCIDs from the bibfile. Is that correct? Or were we also adding html links to the pdfs?

stargaser:
Thanks for trying this out @deniederhut ! As I recall, we concluded that putting html links in the pdfs obscures them and is not very helpful, and that putting the html link on the articles html page is sufficient.

This leaves:

- remove the ORCIDs from the bibfile (easy)
- validating the ORCIDs -- @arfon passed along [this code](https://github.com/openjournals/whedon/blob/master/lib/whedon/orcid_validator.rb) from JOSS. I feel we should do the same, yes?

-

deniederhut:
Yeah that sounds good to me! The example is in ruby though, which we probably will want to translate to python 🐍🐍🐍

stargaser:
I've tested the ORCID validation with some real ORCIDs. The ones in the example are made up but they do have the right checksums.

publisher/writer/__init__.py
```diff
@@ -477,6 +478,37 @@ def visit_PartLaTeX(self, node):
         self.out.append('\n' + node['latex'] + '\n')
         raise nodes.SkipNode
 
+    def validate_orcid(self, orcid):
+        """ Validate the ORCID string provided for an author
+
+        The code has been ported from this Ruby code from JOSS:
+        https://github.com/openjournals/whedon/blob/master/lib/whedon/orcid_validator.rb
```
deniederhut, line 485: Attribution! Love it!

deniederhut (approved):
Looks like this is behaving as expected. Thanks Dave!

PR #548, opened 2020-05-28
Paper: "Falsify your Software: validating scientific code with property-based testing"

Zac-HD:
(still sorting out a few tweaks, but it's at least ready to see if the PDF builds)


Zac-HD:
Seems like the build server doesn't like draft PRs... will try again.

PR #549, opened 2020-05-28
Paper: Falsify your Software: validating scientific code with property-based testing

Zac-HD:
PDF link: http://procbuild.scipy.org/download/Zac-HD-zac-hatfield-dodds

> **Abstract:** *Where traditional example-based tests check software using manually-specified input-output pairs, property-based tests exploit a general description of valid inputs and program behaviour to automatically search for falsifying examples. Given that Python has excellent property-based testing tools, such tests are often easier to work with and routinely find serious bugs that all other techniques have missed.
> I present four categories of properties relevant to most scientific projects, demonstrate how each found real bugs in Numpy and Astropy, and propose that property-based testing should be adopted more widely across the SciPy ecosystem.*

### DIFF
--- a/papers/zac_hatfield-dodds/falsify-your-software.rst
+++ b/papers/zac_hatfield-dodds/falsify-your-software.rst
@@ -1,0 +1,448 @@
+:author: Zac Hatfield-Dodds
+:email: zac.hatfield.dodds@anu.edu.au
+:institution: Australian National University
+:orcid: 0000-0002-8646-8362
+
+:bibliography: references
+
+
+-----------------------------------------------------------------------------
+Falsify your Software: validating scientific code with property-based testing
+-----------------------------------------------------------------------------
+
+.. class:: abstract
+
+    Where traditional example-based tests check software using manually-specified
+    input-output pairs, property-based tests exploit a general description of
+    valid inputs and program behaviour to automatically search for falsifying
+    examples.  Given that Python has excellent property-based testing tools,
+    such tests are often *easier* to work with and routinely find serious bugs
+    that all other techniques have missed.
+
+    I present four categories of properties relevant to most scientific projects,
+    demonstrate how each found real bugs in Numpy and Astropy, and propose that
+    property-based testing should be adopted more widely across the SciPy ecosystem.
+
+
+.. class:: keywords
+
+   methods, software, validation, property-based testing
+
+
+
+Introduction
+------------
+
+Much research now depends on software for data collection, analysis, and reporting;
+including on software produced and maintained by researchers.  This has empowered
+us enormously: it is hard to imagine an analysis that was possible *at all* a
+generation ago which could not be accomplished quickly by a graduate student today.
+
+Unfortunately, this revolution in the power and sophistication of our software has
+largely outstripped work on validation.  While it would be unthinkable to publish
+research based on custom-built and unvalidated physical instruments, this is routine
+in software.  As an effect, :cite:`Soergel2015` estimates that
+
+    Any reported scientific result could very well be wrong if data have passed
+    through a computer, and these errors may remain largely undetected [#]_. It is
+    therefore necessary to greatly expand our efforts to validate scientific software.
+
+I argue that property-based testing :cite:`PraisePBT` is more effective for validation
+of Python programs than using only traditional example-based tests [#]_, and support
+this argument with a variety of examples from well-known scientific Python projects.
+
+This is a recent development: while the concept of property-based testing dates back
+to 1999 :cite:`QuickCheck`, early tools required deep computer-science expertise.
+Since 2015, Hypothesis :cite:`MacIver2019` has made state-of-the-art testing technology
+available *and accessible to non-experts* in Python, and has added multiple features
+designed for testing scientific programs [#]_ since 2019.
+
+
+.. [#] and indeed :cite:`NMRbug` reported such a bug, affecting around 160 papers.
+.. [#] i.e. common workflows using ``pytest``.  If you have no automated tests at all,
+        fix that first, but your second test could reasonably use Hypothesis.
+.. [#] e.g. numeric-aware error reporting, first-class support for array shapes
+        including broadcasting and gufunc signatures, dtypes and indexers, etc.
+
+
+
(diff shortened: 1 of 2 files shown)


deniederhut:
Hey! I took a quick look at the build errors, and it looks like bibtex tripping on your use of backslashes in journal names in the bibfile for this paper. If you can handle those, the rest of this should be okay.

.gitignore
```diff
@@ -1,5 +1,6 @@
 .DS_Store
 .ipynb_checkpoints
+.hypothesis
```
deniederhut: We prefer that paper PRs not modify anything outside of their own directory (this makes managing the git history a lot easier for us). Would you mind reverting this change to .gitignore? If it's annoying for you, we'd be happy to have you submit that change as a separate PR targeting the `dev` branch.

deniederhut: Thanks!

papers/zac_hatfield-dodds/falsify-your-software.rst
```diff
@@ -0,0 +1,447 @@
...
+clear design feedback: where example-based tests map one input to one output regardless
+of complexity, every special case or interacting feature has to be addressed.  Clean
+abstractions which allow you to say "for all ...", without caveats, stand in clear
+contrast and are a pleasure to test.
+
+Tests which use random data are usually property-based, but using a library designed
+for the task has several advantages:
+
+- a concise and expressive interface for describing inputs
+- tests are never flaky - failing examples are cached and
```
anirudhacharya, line 86: this requires more explanation. how do libraries that do Property Based Testing prevent flakiness in tests? If anything, adding random data generators into tests, adds flakiness into the testing framework. 

A more general comment - how can we reproduce test failures with PBT? With random number generators, we can seed the random number generator to reproduce results. Do the various PBT frameworks also do something similar under the hood?

anirudhacharya (commented):
Can property based testing replace the traditional example-based unit testing?

papers/zac_hatfield-dodds/falsify-your-software.rst
Zac-HD: The basic trick is that we cache the *output from the PRNG* between runs, and automatically replay any previous failures on the next run.

There are a bunch of options to explicitly report and then force particular seeds or insert a buffer into the cache, but in normal development it's entirely automatic.

This means that we can get the benefit of a different set of examples each run, while also having any failures replayed every time.  (or if you *want* the same set of examples each time, you can set the seed manually too)

Zac-HD:
> Can property based testing replace the traditional example-based unit testing?

*Mostly*, yes.  A good example might be my `hypothesis-jsonschema` project: of around 500 tests, about six are traditional example-based tests, 80 are parametrised tests (so ~10 tests averaging eight cases each), and the remaining 400 tests are all property-based and run hundreds of examples each.

Traditional example-based tests are still nice to pin down specific or weird edge cases, where asserting the exact result is easy and a property would be fiddly or error-prone, but almost all of the tests I write are property-based.

deniederhut:
Hey @Zac-HD  ! Thanks to some awesome behind the scenes work by @stargaser, we now have experimental support for publishing papers with ORCIDs. If you have one and would like to add it to your paper, it's as simple as adding an `:orcid:` tag under your name in the paper header. [You can see an example of how to do this in the `00_vanderwalt` example paper](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2020/papers/00_vanderwalt/00_vanderwalt.rst). If you don't have one, no worries! You can still publish your paper as is.

Zac-HD:
I do indeed have one, and took it out of my early draft when I noticed that it wasn't supported yet.  Thanks for your persistence @stargaser!

anirudhacharya:
> > Can property based testing replace the traditional example-based unit testing?
> 
> _Mostly_, yes. A good example might be my `hypothesis-jsonschema` project: of around 500 tests, about six are traditional example-based tests, 80 are parametrised tests (so ~10 tests averaging eight cases each), and the remaining 400 tests are all property-based and run hundreds of examples each.
> 
> Traditional example-based tests are still nice to pin down specific or weird edge cases, where asserting the exact result is easy and a property would be fiddly or error-prone, but almost all of the tests I write are property-based.

Wouldn't it also make sense to have unit tests for Test Driven Development to ensure basic sanity of the software as it is being developed. It would seem property-based tests are not so good for Test Driven Development.

deniederhut:
Hi @Zac-HD ! Do you have thoughts about using hypothesis with TDD? Or do you feel this is out of scope for the paper?

Zac-HD:
(sorry I missed this before!)

I think TDD is out of scope for this paper, but for what it's worth I also think property-based tests are just as applicable for TDD as for any other way testing fits into your development cycle.  Personally, I'd rather do my basic sanity-checks with properties than specific examples - @anirudhacharya if you've found this problematic I'd be interested to hear what difficulties you ran into!

deniederhut:
Thanks for the response! @anirudhacharya do you feel that this paper is now ready for inclusion in the proceedings?

anirudhacharya:
> Thanks for the response! @anirudhacharya do you feel that this paper is now ready for inclusion in the proceedings?

@deniederhut Yes I do.

papers/zac_hatfield-dodds/references.bib
```diff
@@ -0,0 +1,164 @@
...
+  publisher = {The Open Journal},
+  volume = {4},
+  number = {43},
+  pages = {1891},
+  author = {David MacIver and Zac Hatfield-Dodds and Many Contributors},
+  title = {Hypothesis: A new approach to property-based testing},
+  journal = {Journal of Open Source Software}
+}
+
+% TODO: to be replaced by formal citation as soon as one is available
```
deniederhut, line 16: Is this a note to yourself to update this reference before your paper gets merged?

Zac-HD: Yes, but with the caveat that it won't have a DOI minted until mid-July!

So I'm happy to merge as-is, and update it later if that's possible?

deniederhut: Works for me!

PR #550, opened 2020-05-28
Paper: Awkward Array: JSON-like data, NumPy-like idioms

jpivarski:


deniederhut:
That commit history is too relatable 😂

bal827:
Hi!

I am one of the reviewers of this paper and I will send you my first comments before the 18th.

Thanks!
Isha

hongsupshin:
Hi there, I'm also one of the reviewers of the paper. I will finish up the review by the end of this weekend. Sorry for the delay!

deniederhut:
Hey @jpivarski  ! Thanks to some awesome behind the scenes work by @stargaser, we now have experimental support for publishing papers with ORCIDs. If you have one and would like to add it to your paper, it's as simple as adding an `:orcid:` tag under your name in the paper header. [You can see an example of how to do this in the `00_vanderwalt` example paper](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2020/papers/00_vanderwalt/00_vanderwalt.rst). If you don't have one, no worries! You can still publish your paper as is.

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+In this paper, we will describe Awkward Array [2]_ [3]_, a generalization of NumPy's core functions to the nested records, variable-length lists, missing values, and heterogeneity of JSON-like data. In doing so, we'll focus on the internal representation of data structures as columnar arrays, very similar to (and compatible with) Apache Arrow [4]_. The key feature of the Awkward Array library, however, is that calculations, including those that restructure data, are performed on the columnar arrays themselves, rather than instantiating objects. Since the logical data structure is implicit in the arrangement of columnar arrays and most operations only rearrange integer indexes, these functions can be precompiled for specialized data types, like NumPy.
```
hongsupshin: "The key feature of the Awkward Array library, however, is that calculations, including those that restructure data, are performed on the columnar arrays themselves, rather than instantiating objects."

Could you be more specific about the pros of your package compared to numpy?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+Motivation from particle physics
+--------------------------------
```
hongsupshin: I do understand that this was how the development of the package has started but I don't know how easy to understand the particle physics example without any domain knowledge, meaning without clear idea about the data itself. I think you didn't want to keep this part too long, which makes sense. But without any explanation about what these particles mean, why it makes sense to read the peak in the histogram, and so on, I am not sure whether it is worth mentioning this detail unless you are aiming physicists as your main audience. You could either provide more information to include more audience or just reduce domain-specific information but add more explanation of the heterogeneity of the data with specific examples.

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+Since kaons have a well-defined mass, the :math:`m_{K_S}` values that correspond to real kaons form a peak in a histogram: see Figure :ref:`physics-example`. Not knowing which pions are due to kaon decays, every pair of pions in a collision event must be searched, in principle.
+
+.. figure:: figures/physics-example.pdf
+   :align: center
+   :scale: 13%
+   
+   Example of a particle physics problem requiring combinatorial search: all pairs of pions in a collision event must be tested for compatibility with decay from a kaon. :label:`physics-example`
+
+Physicists employ many other techniques, but most of them involve combinatorial searches like this one. Since the numbers of particles of each type differ from event to event and they carry dozens of attributes that must be referenced in the combinatorial pairs, this would be a hard analysis to do in NumPy.
```
hongsupshin: "this would be a hard analysis to do in NumPy" -> What would entail if one wants to use NumPy? Could you briefly explain?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+   m_{K_S} = \sqrt{(E_{\pi^+} + E_{\pi^-})^2 - \left|\vec{p}_{\pi^+} + \vec{p}_{\pi^-}\right|^2}
+
+Since kaons have a well-defined mass, the :math:`m_{K_S}` values that correspond to real kaons form a peak in a histogram: see Figure :ref:`physics-example`. Not knowing which pions are due to kaon decays, every pair of pions in a collision event must be searched, in principle.
+
+.. figure:: figures/physics-example.pdf
+   :align: center
+   :scale: 13%
+   
+   Example of a particle physics problem requiring combinatorial search: all pairs of pions in a collision event must be tested for compatibility with decay from a kaon. :label:`physics-example`
```
hongsupshin: I think the figure requires more annotation in general. What is CMS? What are the numbers for Run/Event? Are these indices? What is K0s? What are the lines mean and what do the colors indicate?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+Physicists employ many other techniques, but most of them involve combinatorial searches like this one. Since the numbers of particles of each type differ from event to event and they carry dozens of attributes that must be referenced in the combinatorial pairs, this would be a hard analysis to do in NumPy.
+
+With Awkward Array, however, it would be
+
+.. code-block:: python
+
+   kaon_masses = ak.combinations(pions[good], 2).mass
+
```
hongsupshin: - Maybe it's better to import the module here too for `ak`
- What is `good`? What is the type?
- `.mass`: this makes me think that there might be functions in your module that are specifically for particle physics. Is this the case? If yes, I think it would be nice to mention it somewhere in the beginning of the paper.

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+.. code-block:: python
+
+    import urllib.request
+    import json
+
+    url = "https://raw.githubusercontent.com/Chicago/" \
+          "osd-bike-routes/master/data/Bikeroutes.geojson"
+    bikeroutes_json = urllib.request.urlopen(url).read()
+    bikeroutes_pyobj = json.loads(bikeroutes_json)
```
hongsupshin, line 85: It might be better to give an overview of this json file in terms of structure so that when you refer the key names later the audience can follow easily.

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+Importing this JSON object as an Awkward Array splits its record-oriented structure into a contiguous buffer for each field, making it ready for columnar operations. Heterogeneous data are split by type, such that each buffer in memory has a single numerical type.
+
+.. code-block:: python
+
+    import awkward1 as ak
+    bikeroutes = ak.Record(bikeroutes_pyobj)
+
```
hongsupshin, line 93: What does `ak.Record` do?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+.. code-block:: python
+    
+    longitude = bikeroutes["features", "geometry",
+                           "coordinates", ..., 0]
+    latitude  = bikeroutes["features", "geometry",
+                           "coordinates", ..., 1]
+
+The :code:`longitude` and :code:`latitude` arrays both have type :code:`1061 * var * var * float64` (expressed as a Datashape): 1061 routes with a variable number of variable-length polylines.
+
+To compute lengths of each route, we can use NumPy universal functions (like :code:`np.sqrt`) and reducers (like :code:`np.sum`), which are overridden by Awkward-aware functions using NumPy's NEP-13 [6]_ and NEP-18 [7]_ protocols. Distances between points can be computed with :code:`a[:, :, 1:] - a[:, :, :-1]` even though each inner list :code:`a[:, :]` may have a different length.
```
hongsupshin: - Why specifically NEP-13 and -18 protocols? How are they different?
- Is `a[]` an arbitrary ak array?
- why the indices `1` and `-1`?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+.. code-block:: python
+
+    km_east = (longitude - np.mean(longitude)) * 82.7
+    km_north = (latitude - np.mean(latitude)) * 111.1
```
hongsupshin, line 135: What do 82.7 and 111.1 represent?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+                last = (km_east, km_north)
+
+            route_length.append(sum(segment_length))
+        total_length.append(sum(route_length))
+
+.. figure:: figures/bikeroutes-scaling.pdf
+   :align: center
+   :scale: 45%
+
+   Scaling of Awkward Arrays and pure Python loops for the bike routes calculation shown in the text. :label:`bikeroutes-scaling`
```
hongsupshin, line 195: What are the hardware specs for these runs?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+Columnar representation, columnar implementation
+------------------------------------------------
+
+Like Arrow, Awkward data structures are not localized in memory. Instead of concentrating all data for one array element in nearby memory (as an "array of structs"), all data for a given field are contiguous, and all data for another field are elsewhere contiguous (as a "struct of arrays"). This favors a pattern of data access in which only a few fields are needed at a time.
+
+Additionally, Awkward operations are performed on columnar data without returning to the record-oriented format. To illustrate, consider a list of variable-length lists, such as longitude points along a bike route,
+
+.. code-block:: python
+
+    [[1.1, 2.2, 3.3], [4.4], [5.5, 6.6], [7.7, 8.8, 9.9]]
```
hongsupshin, line 239: Could you explain what each list represent?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+    stops:   3, 4, 6, 9
+    content: 1.1, 2.2, 3.3, 4.4, 5.5, 6.6, 7.7, 8.8, 9.9
+
+This new array represents
+
+.. code-block:: python
+
+    [[     2.2, 3.3], [   ], [     6.6], [     8.8, 9.9]]
+
+and we could reuse the original :code:`content` to construct it. Since :code:`content` is untouched, the slice can be a precompiled routine that treats the :code:`content` as an opaque pointer. The :code:`content` might contain other lists or records, like the example in Figure :ref:`example-hierarchy`. Similarly, :code:`a[:, :-1]` is computed by subtracting :code:`1` from the original :code:`stops`, and it is up to the :code:`-` operation to align the :code:`content` of its arguments before applying :code:`np.subtract`.
```
hongsupshin: - Without exactly what the "content" represent, I think it's a bit difficult to follow your example.
- What is exactly `a`?
- For `a[:, 1:]`, you said we're adding 1 to `starts` and for `a[:, :-1]` we're subtracting 1 from `stops`. Based on these two statements, the indexing seems to be a bit confusing. How does it know whether to use `starts` or `stops`? Also, it lookslike `a` has two dimensions. What is its 1st dimension?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+Awkward 1.x
+-----------
+
+The first widely used version of Awkward Array (0.x) was released in September 2018 as a pure Python module, in which all of the columnar operations were implemented using NumPy. This library was successful, but limited, since some data transformations are difficult or impossible to write without explicit loops.
+
+In August 2019, we began a half-year project to rewrite the library in C++ (1.x), isolating all of the array manipulations in a "CPU kernels" library that can be swapped for "GPU kernels." Apart from the implementation of the "GPU kernels," this project is complete, though users are still transitioning from the original "Awkward 0.x" to the new "Awkward 1.x," which are both available as separate libraries in PyPI.
```
hongsupshin: I think some of this information can be included in the introduction.

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+* an alternate implementation of array navigation was written for Python functions that are compiled by Numba,
+* array manipulation algorithms (without memory management) are independently implemented as "CPU kernels" and "GPU kernels" plugins. The kernels' interface is pure C, allowing for reuse in other languages.
+
+.. figure:: figures/awkward-1-0-layers.pdf
+   :align: center
+   :scale: 45%
+
+   Components of Awkward Array, as described in the text. :label:`awkward-1-0-layers`
+
+Most array operations are shallow, affecting only one or a few levels of the hierarchy, but even in the worst case, an operation initiated by a Python call steps over at most all of the nodes of an array, which is no more than thousands in particle physics use-cases. The number of elements in the array can be billions (multi-GB memory). The loops over array elements are strictly contained in the kernels, so performance optimizations can focus on this layer.
```
hongsupshin: If the GPU kernels are not actually implemented yet, it would make sense to address that in the figure or in the caption for clarification.

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+    #                             values
+    # entry subentry subsubentry
+    # 0     0        0               1.1
+    #                1               2.2
+    #       2        0               3.3
+    # 2     0        0               4.4
+    #       1        0               5.5
+    #                1               6.6
+    # 3     0        0               7.7
+    # 4     0        0               8.8
```
hongsupshin, line 485: - It looks like the empty lists are dropped but still potentially can be identified by checking missing entry or subentry indices. It looks like it might be still possible to include the empty lists as nan or something else. I am curious to know about the decision that went into dropping the empty lists.
- It is possible that each list in a list of lists can contain different data types. In this case, if we return a single column like this, pandas might just infer the column type as object and the information about the original data types (for each list) might be lost. Is there any way to mitigate this issue?

hongsupshin (commented):
@jpivarski Hi, Jim! Thanks for such an interesting paper. I didn't know about the Awkward Array until now, but it does seem very useful and quite versatile for many domains (also pretty straightforward to use). I think the manuscript is already in a good shape and so my comments are mostly for clarification and better communication. Hope they help!

jpivarski:
> I think the manuscript is already in a good shape and so my comments are mostly for clarification and better communication. Hope they help!

@hongsups Thank you for your detailed review! I'm in the process of addressing your comments now, first by editing the paper, then by responding to each of your points.

The biggest change is that I'll be replacing the motivation from physics with more detail on the bike routes example. I also started to have second thoughts about the physics section while writing and recording my talk last week—the examples that had been helpful when I was always speaking to a physics audience seem to be unhelpful when I'm not, since the domain-specific background is hard to introduce in a short space and takes focus away from the main point of longer. Many of your specific criticisms are about the physics section, which coincides with my growing unease about it. Instead of going 50/50 with a physics example and a non-physics example, I should more fully discuss the example that doesn't require domain knowledge.

You also had comments beyond the physics section, and I'm not forgetting those. Your idea about including the history and status of the project in the introduction is a good one.

papers/jim_pivarski/paper.rst
bal827: I agree with the comment above! Maybe, one of the way to convey the message of motivation from particle physics would be to just retain the first graph here in the section or make it a bit more comprehensive and then talk directly about the bike routes example.

bal827:
I had no idea about Awkward Arrays before! I ended up reading in detail about NumPy and Arrow to learn more about how Awkward Arrays adds value in this ecosystem. This paper is already written so well, however I do feel it caters more to Physicists/Physics Audience. Most of the suggestions have already been conveyed by @hongsups. I am mostly just reiterating some of those suggestions. I did have difficulty in understanding the overall performance (time) of Awkward Arrays.

papers/jim_pivarski/paper.rst
bal827: Reiterating the comment above, I have trouble understanding this particular example.

jpivarski:
@bal827 Thanks for all of the feedback! I'm taking it all in consideration and will have a new draft shortly. (In fact, I can wait until you're done with your comments to be sure to address them all.) I've been working on it most of the day, but I haven't been committing because getting the updates in bits and pieces would make it more confusing for you as reviewers. (I.e. "Which version should we take to be the update?") So I'm incorporating everything that I'm reading here into a single update, which I'll post when I'm done and you're ready.

papers/jim_pivarski/paper.rst
bal827: A good idea would be to put the snapshot of this json file showing high level structure of the file.

papers/jim_pivarski/paper.rst
bal827: Same as above, where is this record function from?

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,558 @@
...
+
+The C++ layer is therefore not motivated by performance, since that is the responsibility of the kernels. It exists because many particle physics libraries are written in C++ and having a full object model for the arrays in C++ allows us to pass data to and from C++ libraries as Awkward Arrays, avoiding unnecessary conversions. The C++ implementation freely uses dynamic dispatch (virtual methods) and atomic reference counting (shared pointers) to match Python's object model.
+
+Numba for just-in-time compilation
+----------------------------------
+
+Some expressions are simpler in "vectorized" form, such as :code:`pions[good]` to select :code:`pions` with a broadcastable array of booleans :code:`good`. However, some operations are more difficult to express in this form, particularly iterative algorithms.
```
bal827: I had to search what is "pion" here, again re-iterating : maybe a good idea would be to use more generalized examples or use more known familiar physics terms wherever possible (this is assuming the target audience is more general).

bal827 (commented):
Thanks! I am quite done with my review! The paper is already in pretty much good shape!

papers/jim_pivarski/paper.rst
jpivarski: I have expanded this section with a clarification that

> Our goal in developing Awkward Array is not to replace NumPy, but to extend the set of problems to which it can be applied.

Specifically, it's the ability to apply NumPy's own functions to non-rectilinear data.

papers/jim_pivarski/paper.rst
jpivarski: Based on both of your comments, I eliminated the physics section entirely, giving space to explain the bike routes (non-physics) example more fully.

But the "hard to do in NumPy" is because neither the data structures nor the algorithms are expressible in NumPy. (Without serious contortions, which is what Awkward 0.x was; an abstraction layer that performed this mapping, but it's non-trivial.)

papers/jim_pivarski/paper.rst
jpivarski: Yes, physics plot. These things are boilerplate; some of them are required by the collaboration that provided the data (CMS), but it's a significant distraction here.

papers/jim_pivarski/paper.rst
jpivarski: It's a physics example in compressed space; the names were meant to be suggestive of their types, but that fell flat. For years, I've been using examples like this in a specialized community for which they are clarifying, rather than obfuscating, but I need to rethink my assumptions.

papers/jim_pivarski/paper.rst
jpivarski: I have now done this. (In parallel with your suggestions, actually.) Now that I have space, I included a Datashape expression of the array type in this file, which reads like a schema.

papers/jim_pivarski/paper.rst
jpivarski: I think it might be clearer now that I refer to them as the `__array_ufunc__` and `__array_function__` protocols with links to their definitions as NumPy Enhancement Proposals. Even if you haven't heard of these special methods, the double-underscores suggest that they're methods to override (for an audience with strong Python background, something I can assume of SciPy).

As for the array, I've explicitly stated that we're inventing `a` and I explain the logic of the `a[1:]`, `a[:-1]` slicing more fully.

papers/jim_pivarski/paper.rst
jpivarski: I added the CPU model, clock speed, cache size, and number of threads in a parenthesized comment.

papers/jim_pivarski/paper.rst
jpivarski: As stated elsewhere, I removed this section to give more room to expand the others, particularly the bike routes example, which is (I think) an accessible introduction. The text also had to be adjusted to accommodate this—the physics motivation is mentioned in several places, but specific physics examples are not.

papers/jim_pivarski/paper.rst
jpivarski: I have added explanation in the text: `ak.Array` and `ak.Record` are the two primary types in the library, and this constructor was used because the GeoJSON dataset is a single JSON object (like a record).

papers/jim_pivarski/paper.rst
jpivarski: I have added explanation in the text: it's a conversion from degrees to kilometers at the latitude of Chicago. The bike routes example is one that I've been reworking in several presentations, and I forgot that this was left unexplained here.

papers/jim_pivarski/paper.rst
jpivarski: It's a toy example, now specified as such in the text.

papers/jim_pivarski/paper.rst
jpivarski: I added quite a bit more explanation to this section. I've been using words like `content`, `starts`, and `stops` as jargon for a while, and so I precede their introduction with prose. The words are still used, but they could just be taken as labels for the array buffers described in the text.

papers/jim_pivarski/paper.rst
jpivarski: It is now. Thanks!

papers/jim_pivarski/paper.rst
jpivarski: I suppose indicating empty lists with an entry/subentry/subsubentry of NaN value would be a valid representation of the data, though the fact that there are empty lists here is not lost because the entry numbers skip: the presence of an `[]` at entry `1` can be inferred from the fact that the entry number goes from `0` to `2`. However, that doesn't work for empty lists after the last non-empty list!

Choosing a representation like that, with NaNs at the appropriate entry number would have consequences for subsequent analysis that I haven't fully thought through yet, but there's likely to be trouble. Dropping the empty lists is a natural consequence of broadcasting: we're broadcasting entry numbers onto the data in the list and that's a natural thing to do. Adding an entry with a NaN at the empty list would make this operation differ from broadcasting in other contexts, where this is the right thing to do. For example,

```python
>>> import awkward1 as ak
>>> a = ak.Array([[1, 2, 3], [], [4, 5]])
>>> b = ak.Array([100, 200, 300])
>>> a + b
<Array [[101, 102, 103], [], [304, 305]] type='3 * var * int64'>
```

In this context, we need the empty one to go away to get what a user clearly wants: to match the three scalar values of `b` onto all the values in each corresponding list in `a`. That requires the `200` to disappear, to get associated with nothing.

"Exploding" the Awkward Array into equivalentish Pandas structures is a lossy operation, though one that has been useful so far. The alternative is to use the Awkward Arrays as Pandas columns, which does not lose any information:

```python
>>> import pandas as pd
>>> a = ak.Array([[1, 2, 3], [], [4, 5]])
>>> b = ak.Array([100, 200, 300])
>>> pd.DataFrame({"a": a, "b": b})
           a    b
0  [1, 2, 3]  100
1         []  200
2     [4, 5]  300
```

To prove that these are still Awkward Arrays (and not converted into NumPy `dtype=object` arrays, for instance), I can apply a ufunc to them:

```python
>>> pd.DataFrame({"a": a, "b": b}) * 2
           a    b
0  [2, 4, 6]  200
1         []  400
2    [8, 10]  600
```

This is fully information-preserving (the Awkward Array is simply wrapped in a `pd.Series`), but it has been less useful because there aren't many Pandas functions that can be applied to data of this type. The ufuncs are an exception because Pandas and Awkward both speak fluent ufunc.

In the end, Awkward Arrays and Pandas DataFrames are rather different ways of approaching what is _sometimes_ the same data. The scope of a single Awkward Array is broader than the scope of a single Pandas DataFrame, but a relational set of DataFrames, linked by foreign keys, are as expressive as an Awkward Array. Awkward operations are more focused on local transformations: things you do with logically "nearby" parts of the structure, while Pandas has much better support for long-distance things like JOINs. Physics analyses on statistically independent events really didn't need JOINs, but they did need a lot of combinatoric operations like _local_ cross-joins ([ak.cartesian](https://awkward-array.readthedocs.io/en/latest/_auto/ak.cartesian.html) with `axis > 0`) that Pandas doesn't provide, so one couldn't be replaced by the other.

papers/jim_pivarski/paper.rst
jpivarski: The words "pion" and "kaon" have been fully exorcised from the text. (An incremental search confirms it.)

papers/jim_pivarski/paper.rst
jpivarski: You're right—that would make a lot more sense. I moved this information to the introduction, which required adjustments throughout the text, but the history and current status of the project are now stated at the beginning. In fact, it's useful to refer to it later on.

jpivarski:
@hongsups and @bal827, I believe all of your comments have been addressed, and all but one of the conversations has been resolved. (I left that one open because it doesn't correspond to a change in the text; it made more sense to have that discussion outside of the text.) Thank you for your input!

papers/jim_pivarski/paper.rst
```diff
@@ -42,45 +42,37 @@ Awkward Array: JSON-like data, NumPy-like idioms
...
+Origin and development
+----------------------
 
-The most basic operation reconstructs particles that decayed before they could be detected, using kinematic constraints on their visible decay products. For example, if a neutral kaon (:math:`K_S`) decays into two charged pions (:math:`\pi^+`, :math:`\pi^-`), the energy (:math:`E`) and momenta (:math:`\vec{p}`) of the observed pions can be combined to reconstruct the unobserved kaon's mass (:math:`m`):
+Awkward Array was intended as a way to enable particle physics analyses to take advantage of scientific Python tools. Particle physics problems are inherently structured, heavy on nested loops over variable-length lists. They also involve big data, typically tens to hundreds of terabytes per analysis. Traditionally, this required physicists to do data analysis in Fortran (with custom libraries for data structures [hydra]_ before Fortran 90) and C++, but many physicists are now moving to Python for end-stage analysis [phypy]_. Awkward Array provides the link between scalable, interactive, NumPy-based tools and the nested, variable-length data structures that physicists need.
```
hongsupshin: This new intro is excellent!

papers/jim_pivarski/paper.rst
```diff
@@ -42,45 +42,37 @@ Awkward Array: JSON-like data, NumPy-like idioms
...
 
-.. math::
+Since its release in September 2018, Awkward Array has become one of the most popular Python libraries for particle physics, as shown in Figure :ref:`awkward-0-popularity`. The Awkward 0.x branch was written using NumPy only, which limited its development because every operation must be vectorized for performance. We (the developers) also made some mistakes in interface design and learned from the physicists' feedback.
 
-   m_{K_S} = \sqrt{(E_{\pi^+} + E_{\pi^-})^2 - \left|\vec{p}_{\pi^+} + \vec{p}_{\pi^-}\right|^2}
-
-Since kaons have a well-defined mass, the :math:`m_{K_S}` values that correspond to real kaons form a peak in a histogram: see Figure :ref:`physics-example`. Not knowing which pions are due to kaon decays, every pair of pions in a collision event must be searched, in principle.
-
-.. figure:: figures/physics-example.pdf
+.. figure:: figures/awkward-0-popularity.pdf
```
hongsupshin: I like that you are talking about the origin and development in the beginning as well. Showing the popularity graph first delivers a stronger message.

papers/jim_pivarski/paper.rst
```diff
@@ -92,14 +84,39 @@ However, nested data structures are not unique to particle physics, so we presen
...
+             "STREET": string,
+             "TYPE": string,
+             "BIKEROUTE": string,
+             "F_STREET": string,
+             "T_STREET": option[string]},
+         "geometry": {
+             "type": string,
+             "coordinates":
+                 var * var * var * float64}}}
+
```
hongsupshin, line 116: Thanks for including this! In general, the new bike example description is very comprehensive and much easier to follow compared to the previous version!

papers/jim_pivarski/paper.rst
```diff
@@ -144,21 +185,21 @@ The same could be performed with the following, though the vectorized form is sh
                         np.sqrt(dx2 + dy2))
                 last = (km_east, km_north)
 
-            route_length.append(sum(segment_length))
-        total_length.append(sum(route_length))
+            polyline_length.append(sum(segment_length))
+        route_length.append(sum(polyline_length))
```
hongsupshin, line 189: Thanks for this example as well.

papers/jim_pivarski/paper.rst
```diff
@@ -214,137 +255,110 @@ A list of lists of lists would use these three buffers as the :code:`content` of
...
 
-This new array represents
+Without modifying the :code:`content`, this new array represents
 
 .. code-block:: python
 
     [[     2.2, 3.3], [   ], [     6.6], [     8.8, 9.9]]
 
-and we could reuse the original :code:`content` to construct it. Since :code:`content` is untouched, the slice can be a precompiled routine that treats the :code:`content` as an opaque pointer. The :code:`content` might contain other lists or records, like the example in Figure :ref:`example-hierarchy`. Similarly, :code:`a[:, :-1]` is computed by subtracting :code:`1` from the original :code:`stops`, and it is up to the :code:`-` operation to align the :code:`content` of its arguments before applying :code:`np.subtract`.
+because the first list starts at index :code:`1` and stops at :code:`3`, the second starts at :code:`4` and ends at :code:`4`, etc. The "removed" elements are still present in the :code:`content` array, but they are now unreachable, much like the stride tricks used for slicing in NumPy.
```
hongsupshin, line 274: This example is now much easier to follow. Thanks for the change.

papers/jim_pivarski/paper.rst
```diff
@@ -214,137 +255,110 @@ A list of lists of lists would use these three buffers as the :code:`content` of
...
+                        route_length[i] += distance
 
-    find_quarks(particles, ak.ArrayBuilder()).snapshot()
+                    # keep track of previous value
+                    first = False
+                    last_east = km_east
+                    last_north = km_north
 
-The ArrayBuilder is described in more detail in the next section.
+        return route_length
```
hongsupshin: Just to clarify, I have a quick question.

This example looks very similar to the previous code block at line 170. Since the function `compute_lengths` takes `bikeroutes` as input argument, my assumption is that the input data type is ak array - the previous example at line 170 looped over `bikeroutes_pyobj`, which was a JSON object. When ak array was first introduced, probably the most eye-catching part was an example like this: `latitude  = bikeroutes["features", "geometry", "coordinates", ..., 1]` where we can just use numpy-like syntax to call a value in a JSON object. So when I saw `feature.geometry.coordinates` at 325, I was a bit confused but I think you're referring to `__getattr__` (dot) operators you mentioned right before this code block. Is this correct?

papers/jim_pivarski/paper.rst
```diff
@@ -503,61 +526,73 @@ The kernels library contains many functions (428 with an :code:`"extern C"` inte
...
+would be required to maintain old analysis scripts.
+
+As an incentive for adopting Awkward 1.x in new projects, it has been heavily documented, with complete docstring and doxygen coverage (already exceeding Awkward 0.x).
 
 Summary
 -------
 
-By providing NumPy-like idioms on JSON-like data, Awkward Array fills a need required by the particle physics community. The inclusion of data structures in array types and operations was an enabling factor in this community's adoption of other scientific Python tools. However, the Awkward Array library itself is not domain-specific and is open to use in other domains.
+By providing NumPy-like idioms on JSON-like data, Awkward Array satisfies a need required by the particle physics community. The inclusion of data structures in array types and operations was an enabling factor in this community's adoption of other scientific Python tools. However, the Awkward Array library itself is not domain-specific and is open to use in other domains. We are very interested in applications and feedback from the wider data analysis community.
```
hongsupshin, line 550: Nice touch on more welcoming message!

hongsupshin (commented):
Thanks for making the changes quickly, Jim. I really like the new added details of examples and the reordering of the contents. I think the message is more compelling and more accessible to a wider audience. Most of my comments are just confirming your changes. I left one comment to just clarify whether I understood a code block correctly. Other than that, looks fine to me. I will leave a separate comment on your answer to the empty list.

papers/jim_pivarski/paper.rst
hongsupshin: Thanks for providing the detailed examples! That's definitely helpful to understand. I can see that simply wrapping ak arrays in a `pd.Series` can preserve information. I think whether it matters to drop empty lists or not would depend on specific use cases (so thanks for the explanation about the physics analyses). However, for a typical data science audience, the information of empty list, which can be triggered by missing data in many real world scenarios, can be quite important. Again, the indices are useful but as you addressed, it's not complete, especially if the array has lists of empty lists (e.g., `[[], [], []]`). I think this can be mitigated if a user has a complete information of all the entry indices, but again, this might not always the case. 

I think the last two paragraphs in your reply provided enough information of why the empty lists do not show in pandas transformation. You can consider adding this information either after your pandas example here or in your package documentation to provide users with more clarification.

papers/jim_pivarski/paper.rst
jpivarski: Right, this code block uses the fact that `__getitem__` with a single string argument is equivalent to `__getattr__` (both inside and outside of Numba). I forgot that I hadn't introduced that in the paper, but I'm using it here. Since this syntactic convenience is a minor point (distracting to introduce?), I should perhaps just convert this example to use `__getitem__`.

Or if I do find a place to introduce it (leaving the example alone), it could go with some text about limitations in Numba-complied functions. Just as Numba supports a subset of the Python language and NumPy library, some things that work outside of a Numba-compiled function don't work inside, such as multidimensional `__getitem__`, which is what motivated me to use the dot syntax here, because

```python
feature["geometry", "coordinates"]
```

(which would be allowed outride of Numba) would have to be

```python
feature["geometry"]["coordinates"]
```

inside, or

```python
feature.geometry.coordinates
```

anywhere, and thus the latter is generally easier. The pure Python code example also has modifications here to satisfy Numba, such as the introduction of `first` as a Boolean, rather than letting numbers be `None` the first time through a loop. Numba supports option-type variables (can be numbers or `None`), but apparently not the way I was using it.

The 250x speedup comes at a cost of having to finagle the Python code into an acceptable form, both in general Numba usage and in using Awkward with Numba. That probably deserves some explanation, since I have space. (I could explain that the difference between the two similar code blocks is that the first is the pure Python that I _would_ write and the second is what I had to write to get it to work. Specific limitations are temporary—I've seen Numba's scope grow a lot—but it's probably always going to be the case that _some_ finagling will be required.)

papers/jim_pivarski/paper.rst
jpivarski: I'll point out the missing lists explicitly. It might be a good way to say a little about how a Pandas analysis differs from an Awkward one—I think that a more typical way to do it in Pandas would be to have another DataFrame for the non-lists data. Working with multiple DataFrames with JOINable indexes would be more relational, and I think data analysis in Pandas is like that, with JOINs being somewhat common. I could perhaps write a progression:

SQL → Pandas → Awkward Array/NumPy → imperative

though this is a subjective claim. (The axis these types of analysis span is "locality." It's easy to do non-local things in SQL, Pandas has more emphasis on timeseries and moving windows, which are more local, Awkward and NumPy work by aligning each element of different arrays, which are more local still, and imperative programming finds it easiest to come things in the same nested for loop, more difficult to relate things outside of it.)

hongsupshin: Yes, I think a lot of readers will really appreciate that explanation about pandas vs. ak comparison. As you explained, I also think the use case Awkward Array is somewhat different from pandas, and being able to convert ak arrays into pandas dataframe is a nice-to-have feature but not the main point of the package. It's just that many data scientists simply go back and forth between np arrays and pandas dataframe, and so if they are ever using ak arrays in the same way, they might want to know more. If you are planning to add more text about this, just tag me when you are done. Thanks for the quick reply!

jpivarski: I added two paragraphs talking about the strengths of Pandas and Awkward for different kinds of problems. I didn't include my highly subjective continuum ("SQL → Pandas → Awkward/NumPy → imperative"), but I did include some of my thinking behind it: that Pandas's strength is that it has joins and can therefore make use of long-distance relationships, whereas Awkward can have more complex local structure.

hongsupshin (approved):
@jpivarski Thanks for implementing additional text to address the discussed issues. All look good to me!

papers/jim_pivarski/paper.rst
```diff
@@ -0,0 +1,600 @@
...
+.. [scipy] Pauli Virtanen et al. *SciPy 1.0: Fundamental Algorithms for Scientific Computing in Python*,
+       SciPy 1.0: Fundamental Algorithms for Scientific Computing in Python. Nature Methods, in press. DOI:10.1038/s41592-019-0686-2
+
+.. [hydra] R. K. Böck. *Initiation to Hydra*,
+        https://cds.cern.ch/record/864527 (1974), DOI:10.5170/CERN-1974-023.402.
+
+.. [phypy] Jim Pivarski. *Programming languages and particle physics*,
+        https://events.fnal.gov/colloquium/events/event/pivarski-colloq-2019 (2019).
+
+.. [pandas] Wes McKinney. *Data Structures for Statistical Computing in Python*,
```
deniederhut, line 581: Would you mind updating this entry to include the DOI? http://conference.scipy.org/proceedings/scipy2010/mckinney.html

jpivarski: Added in 7723619504e96997c89ff874e73f06b3e9bf82ab. Thanks for pointing this out!

PR #552, opened 2020-05-28
Paper: Submission 75: Quasi-orthonormal Encoding for Machine Learning Applications

summerswallow-whi:
process.

for more information.


himito:
Hi @summerswallow-whi,  I will be reviewing your paper and I would thank you for this interesting paper. 

I will put my comments to the code, however, I think it would be important that you give another pass to the writing of the paper, more precisely from the section "implementation" because sometimes it's very difficult to get a fluent reading due to grammatical errors or missing commas :(

Also, it would be interesting to know what are the disadvantages of your approach.

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+While most popular machine learning methods such as deep learning
+require numerical data as input, categorical data is very common
+practice. For example, a person's vitals could be a combination of both,
+they could include height, weight (numerical) and gender, race
+(categorical). The challenge is to convert the categorical data into a
+vector of some sort. 
+
+One-hot encoding which is discussed in the next section is very
+commonly used these days in machine learning but has the draw back
```
himito:
```suggestion
commonly used these days in machine learning but has the drawback
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+Coding methods can be categorized as *classic*, *contrast*,
+*Bayesian* and *word embeddings*. Classic, contrast and Bayseian
+encoding are given a good overview treatment by Hale's blog
+:cite:`Hale2018` with examples to be found as part of the ``scikit-learn`` category
+encoding package :cite:`scikit`. Both
+contrast encoding and Bayesian encoding use the statistics of the data
+to facilitate encoding. These two categories may be of use when more
+statistical analysis is required, however there has not been widespread
+adoption of these emcoding techniques for machine learning.
```
himito:
```suggestion
adoption of these encoding techniques for machine learning.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+The remainder of the section will describe some common classic
+categorical encodings
+
+Ordinal Encoding
+~~~~~~~~~~~~~~~~
+
+To begin our overview of fundamental encoding methods, we start with
+Ordinal (Label) Encoding. Ordinal encoding is the simplest and perhaps
+most naive approach encoding a categorical feature --- one simply
```
himito:
```suggestion
most naive approach encoding for a categorical feature --- one simply
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+assigns a number to each member of a category. This is often how data
+from surveys are encoded into spreadsheets for easy storage and
+calculation of basic statistics. An associated data dictionary is used
+to convert the values back and forth between a number and a category.
+Take for example the case of gender, male could be encoded as 1 and
+female as 2, with a data dictionary as follows:
+:code:``{'male': 1, 'female': 2}``
```
himito:
```suggestion
:code:`{'male': 1, 'female': 2}`
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+     Subaru &  3 &  (0,0,1,0,0) \\
+     Nissan &  4 &  (0,0,0,1,0) \\
+     Mitsubishi &  5 &  (0,0,0,0,1) \\
+     \bottomrule
+     \end{longtable*}
+
+     \caption{Examples of Ordinal and One-Hot Encodings \DUrole{label}{onehot}}
+   \end{table}
+
+The advantage is that one hot encoding doesn't induce an implicit
```
himito:
```suggestion
The advantage is that one hot encoding does not induce an implicit
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+as binary encoding applied to the hashed ordinal value. Hash encoding
+has several advantages. First, it is open ended so new categories can be
+added later. Second, the resultant dimensionality can be much lower than
+one-hot encoding. The chief disadvantage is that categories can collide
+if two categories accidentally map into the same hash value. This is a
+*hash collision* and must be fixed separately using a resolution
+mechanism. Bernardi's blog :cite:`hash` provides a good treatment of hash coding.
+
+Finally, baseN encoding is a generalization of binary encoding that uses
+a number base other than 2 (binary). Below is an example of the Japanese
```
himito: Add reference to the figure instead of `below` because the figure is rendered at the top of the page

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+     Mitsubishi & 5 & 12 & (1,2) & (1,-1) \\
+     \bottomrule
+     \end{longtable*}
+     \caption{Example of Ternary Codes}
+   \end{table}
+
+A disadvantage of all three of these techniques is that while it does
+reduce the dimension of the encoded feature, artificial geometric
+relationships may creep in between unrelated categories. For example,
+:code:``(0.7,0.7)`` may be confusion between Toyota and Honda or a weak Subaru
```
himito:
```suggestion
:code:`(0.7,0.7)` may be confusion between Toyota and Honda or a weak Subaru
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+Of course, with categorical encoding, the ability to decode an encoded vector back to a category can be very important. If the categorical variable is only an input to a machine learning system, retrieving a category may not be very important. For example, one may have a product rating model which delivers a rating based on a number of variables, some numeric like price, but others might be categorical like color, but since the output does not require category decoding, it is not important.
+
+In an application such as categorization or imputation :cite:`gondara`. Retrieving the category from a vector is crucial. In a training a modern classification model, a categorical output is often subject to an activation function which converts a vector into a probability of each category such as a *softmax* function. Essentially, the softmax is a continuous and differential version of a \`\`hard max'' function which would assign a :code:`1` to the vector representing the most likely category and a :code:`0` to all the other categories. The conversion to a probability distribution allows the use of a negative log likelihood loss function rather than the standard root mean squared error.
```
himito:
```suggestion
In an application such as categorization or imputation :cite:`gondara`, retrieving the category from a vector is crucial. In training a modern classification model, a categorical output is often subject to an activation function which converts a vector into a probability of each category such as a *softmax* function. Essentially, the softmax is a continuous and differential version of a \`\`hard max'' function which would assign a :code:`1` to the vector representing the most likely category and a :code:`0` to all the other categories. The conversion to a probability distribution allows the use of a negative log likelihood loss function rather than the standard root mean squared error.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+
+Typically, other classic encoding methods use thresholds to rectify a vector first into a binary or :math:`n`-ary value then decode the vector back to a label in accordance to the encoding. This makes them difficult to use as outputs of machine learning systems such as neural networks that rely on gradients due to lack of differentiability. Also, the decoding process is difficult to convert to a probability distribution, making negative log-likelihood or crossentropy loss functions more difficult to use.
+
+
+Theory
+------
+
+In this section, we'll briefly define and discuss quasiorthogonality, show how it relates to one-hot encoding and describe how this relationship can be used to develop a categorical encoding with lower cardinality.
```
himito:
```suggestion
In this section, we will briefly define and discuss quasiorthogonality, show how it relates to one-hot encoding and describe how this relationship can be used to develop a categorical encoding with lower cardinality.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+Quasiorthogonality
+~~~~~~~~~~~~~~~~~~
+
+In a suitably high dimensional space, two randomly selected vectors are very likely to be nearly orthogonal or quasiorthogonal. In such an :math:`n`-dimensional vector space, there are sets of :math:`K` vectors which are mutually quasiorthogonal where :math:`K\gg n.`. A more formal definition can be stated as follows.
```
himito:
```suggestion
In a suitably high dimensional space, two randomly selected vectors are very likely to be nearly orthogonal or quasiorthogonal. In such an :math:`n`-dimensional vector space, there are sets of :math:`K` vectors which are mutually quasiorthogonal where :math:`K\gg n`. A more formal definition can be stated as follows.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+Given an :math:`\epsilon` two vectors :math:`{\bf x}` and
```
himito:
```suggestion
Given an :math:`\epsilon`, two vectors :math:`{\bf x}` and
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+Encoding
+~~~~~~~~
+
+The principle behind QOE is simple. A quasiorthonormal basis :math:`\{{\bf q}_i\}` is
+substituted for the orthonormal basis :math:`\{{\bf u}_i\}` described above. So given a
+quasiorthonormal basis, we can define a QOE for a set :math:`L=\{l_i\}`
+by :math:`l_i \mapsto \mathbf{q}_i`.
+
+Decoding :math:`\mathbf{z}` under QOE would use a *qargmax* function analogous to the argmax function for one-hot encoding as shown in equation :ref:`eq:qargmax` which is nearly identical to equation :ref:`eq:argmax`.
```
himito:
```suggestion
Decoding :math:`\mathbf{z}` under QOE would use a *qargmax* function analogous to the argmax function for one-hot encoding as shown in equation :ref:`eq:qargmax`, which is nearly identical to equation :ref:`eq:argmax`.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+.. math::
+
+   \mathbf{Q}=  \left[\begin{matrix} 
+   \bigg| & \bigg| & &\bigg | \\ 
+   \mathbf{q}_1 & \mathbf{q}_2 & \cdots & \mathbf{q}_K \\
+   \bigg| & \bigg| & &\bigg | \end{matrix}\right].
+
+that transforms between the QOE space and the one hot encoding space. So
+given a argmax or softmax function, we can express the quasiorthonormal
```
himito:
```suggestion
given an argmax or softmax function, we can express the quasiorthonormal
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+and
+
+.. math::
+   :label: eq:convert
+
+   \mathrm{qsoftmax}(\mathbf{z}) = \mathrm{softmax}(\mathbf{Qz}).
+
+This facilitates the use of optimized functions such as ``softmax`` in libraries
+like ``tensorflow`` and using the above matrix enables quick
+implementation of QOE into these packages. Not only will using native functions accelerated performance it can exploit features such as auto differentiation built into the native functions. A useful property if one wishes to use the qsoftmax function as an activation function.
```
himito: Could you please rewrite the two last sentences. They are hard to understand :(

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+            qx = tf.matmul(tf.constant(basis), x,
+                           transpose_b=True)        
+            return tf.nn.softmax(tf.transpose(qx))
+        return func
+
+
+Pytorch
+~~~~~~~
+
+Presented below is a version of the ``qsoftmax`` function implemented using ``pytorch`` primitives. The use of the ``squeeze`` and ``unsqueeze`` operations convert between a 1-dimensional vector and a 2-dimension matrix having one column. This function is only designed to accept vector inputs. In some models, especially image related models, outputs of some layers maybe multidimensional arrays. If your use case requires a multidimensional imput to the ``qsoftmax`` function the code may need alteration.
```
himito:
```suggestion
Presented below is a version of the ``qsoftmax`` function implemented using ``pytorch`` primitives. The use of the ``squeeze`` and ``unsqueeze`` operations convert between a 1-dimensional vector and a 2-dimension matrix having one column. This function is only designed to accept vector inputs. In some models, especially image related models, outputs of some layers may be multidimensional arrays. If your use case requires a multidimensional input to the ``qsoftmax`` function the code may need alteration.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+    def qsoftmax(x, basis):
+       qx = torch.mm(torch.tensor(basis), 
+                     x.unsqueeze(0).t()).t().squeeze()
+       return torch.nn.functional.softmax(qx,dim=0)
+
+Construction of an Quasiorthonormal set
+---------------------------------------
+
+It is difficult find explicit constructions of quasiorthonormal sets in
```
himito:
```suggestion
It is difficult to find explicit constructions of quasiorthonormal sets in
```

summerswallow-whi:
> Hi @summerswallow-whi, I will be reviewing your paper and I would thank you for this interesting paper.
> 
> I will put my comments to the code, however, I think it would be important that you give another pass to the writing of the paper, more precisely from the section "implementation" because sometimes it's very difficult to get a fluent reading due to grammatical errors or missing commas :(
> 
> Also, it would be interesting to know what are the disadvantages of your approach.

I did a quick pass of the "implementation" section and I agree it could use some improvement in writing. I will revise that today.

As for the disadvantages, I'll be more explicit in the conclusion.

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+the literature. Several methods are mentioned by Kainen :cite:`kainan`, but
+these constructions are theoretical and hard
+to follow. There are a number of combinatorial problems related such as
```
himito:
```suggestion
to follow. There are a number of combinatorial problems related to such as
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+spherical codes :cite:`wiki:spheres` and Steiner Triple Systems :cite:`wiki:steiner`, which strive to find optimal solutions. These are extremely complicated mathematical constructions and not every optimal solution has been found.
+
+As a practical matter, optimal solutions are not necessary as long as the desired characteristics of the quasiorthonormal basis are obtained. As an example, while an optimal solution finds 28 quasiorthonormal vectors with dot products of 0.5 or under are possible in seven dimensions, you may only need 10 vectors. In other words, a suboptimal solution may yield fewer vectors than are possible for a given dimension, or a larger dimension may be required to obtain the desired number of vectors than is theoretically needed.
```
himito:
```suggestion
As a practical matter, optimal solutions are not necessary as long as the desired characteristics of the quasiorthonormal basis are obtained. As an example, while an optimal solution finds 28 quasiorthonormal vectors with dot products of 0.5 or under are possible in seven dimensions, you may only need 10 vectors. In other words, a suboptimal solution may yield fewer vectors that are possible for a given dimension, or a larger dimension may be required to obtain the desired number of vectors that is theoretically needed. 
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+To demonstrate how QOE can be used in machine learning, we provide a simple experiment/demonstration.
+This demonstration in addition to showing how to construct a classification system using QOE gives an sense of the effect of QOE on accuracy. As an initial experiment, we applied QOE to classification of the Modified National Institute of Standards and Technology (MNIST) handwriting dataset :cite:`mnist`, using the 60000 training examples with 10000 test
+examples. As there are 10 categories, we needed sets of quasiorthonormal
+bases with 10 elements. We took the spherical code for 24 points in
+4-dimensions, giving us 12 quasi-orthogonal vectors. The maximum
+pairwise dot product was 0.5 leading to an angle of 60\ :math:`^\circ`.
+We also took the spherical code for 56 points in 7-dimensions, giving 28
+quasi-orthogonal vectors. The maximum pairwise dot product was .33
+leading to an angle of a little over 70\ :math:`^\circ`
```
himito:
```suggestion
leading to an angle of a little over 70\ :math:`^\circ`.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+We used a hidden layer with 64 units with a ReLU activation function.
+Next there is a 20% dropout layer to mitigate overtraining, then an
+output layer whose width depends on the encoding used. We elected for this demonstration to use one of the simplest models hence there are no convolutional or pooling layers used as often seen in other sample MNIST handwriting classifers. The following example is implemented using ``tensorflow`` and ``keras``.
```
himito:
```suggestion
output layer whose width depends on the encoding used. We selected for this demonstration to use one of the simplest models hence there are no convolutional or pooling layers used as often seen in other sample MNIST handwriting classifiers. The following example is implemented using ``tensorflow`` and ``keras``.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+Extending to Spherical Encodings
+--------------------------------
+
+A Deeper Look at Softmax
+~~~~~~~~~~~~~~~~~~~~~~~~
+In principle, according to equation :ref:`eq:argmax`, the dot product against a basis be it orthonormal or quasiorthonormal recovers the category from a potentially noisy encoded value. If one takes a deeper dive into equations :ref:`eq:gsoftmax` and :ref:`eq:qsoftmax`, it is interesting to see what these functions are doing. Figure :ref:`fig:ortho` shows on the left, randomly selected values in a circle of radius 6. On the right shows the vectors after the softmax function is applied. Clearly with a few stragglers, most points either move very close to either of the basis vectors :math:`(0,1)` or :math:`(1,0)`. For this graphic the radius of 6 is chosen because upon examining the input to the softmax function in the example above, the average magnitude of each component is 5.5. In reality, 6 is probably a conservatively low value. Due to the exponential term in the softmax function, even large components would make the separation more pronounced.
```
himito: I don't understand the first sentence. Could you please rewrite it.

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+.. figure:: 3.pdf
+
+   Softmax on a quasiorthogonal basis :label:`fig:qortho`
+   
+It is clear from equation :ref:`eq:gsoftmax` that the exponential term in the numerator  dominates which is why the softmax is effective.  But consider if :math:`\mathbf{z}\cdot\mathbf{u}_i` is much less than zero such as :math:`\mathbf{z}=-5 \mathbf{u}_i`, then the exponential numerator for that term would severely attenuate the output even though the :math:`\mathbf{z}` lies along the same direction as :math:`\mathbf{u}_i`. 
+Since the :math:`\mathbf{z}\cdot\mathbf{u}_j=0` for all other *j*'s, it would still decode to :math:`i`, but because the :math:*i*-th term is so small, any noise could lead to decoding to a different value.
+
+So why bring all this up? Clearly, values encode to negative values along a basis vector would prove problematic. But it offers the prospect of further reducing dimensionality by encoding to no just the basis a vector but also its antipodal vector. To further our graphical example, in figure :ref:`fig:sphere`, we use :math:`(1,0)`, :math:`(0,1)` and their antipodal vectors :math:`(-1,0)` and :math:`(0,-1)` to encode values and apply a softmax using those vectors.
```
himito:
```suggestion
So why bring all this up? Clearly, values encoding to negative values along a basis vector would prove problematic. But it offers the prospect of further reducing dimensionality by encoding to no just the basis a vector but also its antipodal vector. To further our graphical example, in figure :ref:`fig:sphere`, we use :math:`(1,0)`, :math:`(0,1)` and their antipodal vectors :math:`(-1,0)` and :math:`(0,-1)` to encode values and apply a softmax using those vectors.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+     \caption{Results of MNIST Spherical Coding Experiment \DUrole{label}{tab:spherecode}}
+   \end{table}
+
+In this case, the 5-dimensional spherical codes performed close to the
+one-hot encoding by not as closely as the 7-dimension QO codes. The
+3-dimensional spherical codes performed on par with the 4-dimensional QO
+codes.
+
+While the extreme dimensionality reduction from 10 to 4 or 10 to 3 did
+not yield comparable performance to one-hot encoding. More modest
```
himito:
```suggestion
not yield comparable performance to one-hot encoding, more modest
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+     \endfirsthead
+     10 & 97.53\% (97.30\%) & 96.51\% (96.26\%) & 95.37\% (94.83\%) \\
+     20 & 97.68\% (98.02\%) & 96.82\% (97.11\%) & 95.74\% (95.83\%) \\
+     \bottomrule
+     \end{longtable*}
+     \caption{Results of MNIST Spherical Coding Experiment \DUrole{label}{tab:spherecode}}
+   \end{table}
+
+In this case, the 5-dimensional spherical codes performed close to the
+one-hot encoding by not as closely as the 7-dimension QO codes. The
```
himito:
```suggestion
one-hot encoding by not as closely as the 7-dimension QOE codes. The
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+3-dimensional spherical codes performed on par with the 4-dimensional QO
```
himito:
```suggestion
3-dimensional spherical codes performed on par with the 4-dimensional QOE
```

himito:
Thank you @summerswallow-whi for considering my remarks. Also, could you add captions to the subfigures in Figures 1, 2 and 3? By doing that, we could get an idea of each subfigure without having to go through the text.

papers/75_lu/75_lu.rst
summerswallow-whi: Will do. Good catch. Forgot that tables float.

papers/75_lu/75_lu.rst
summerswallow-whi: I rewrote the section this morning. Let me know if the revised phrasing helps or if it is still problematic. Will commit as soon as I get some of your other suggested changes in.

papers/75_lu/75_lu.rst
summerswallow-whi: Fixed

papers/75_lu/75_lu.rst
summerswallow-whi: Let me know if the last commit fixes this.

summerswallow-whi:
Thank you for your meticulous work in reviewing this. I wish I knew how to do the batch commit. But all your explicit changes were accepted. The latest update by me. Address your earlier comments about the implementation section, disadvantages in the conclusion as well as your comment about rewriting the first sentence of the "A Deeper Look at Softmax" subsection.

The only unaddressed issue you flagged is the caption on the graph pairs. I think I can do it but I have to return to the source. Since they involved random samples, the graphics will be slightly different but should not have any effect to the meaning of the paper. I'll get on that now.

summerswallow-whi:
> Thank you @summerswallow-whi for considering my remarks. Also, could you add captions to the subfigures in Figures 1, 2 and 3? By doing that, we could get an idea of each subfigure without having to go through the text.

I caption the subplots within the Figures. I also noticed the scaling was off on them. 

Let me know if there are any other issues you think should be addressed.

himito:
Thank you @summerswallow-whi for quickly applying my remarks. I'll read the paper again and come back to you with my comments. Btw, the link https://github.com/Westhealth/scipy2020/quasiorthonormal/ at the end of the conclusions does not work :(

githomin:
Hi @summerswallow-whi, I will be reviewing your paper as well. I found it very interesting. I agree that the paper could use an editing pass or two purely for language. I will try to leave more detailed suggestions, but they will not be comprehensive. A couple more substantive points: 1) You mention in the beginning that random vectors are w.h.p. quasi-orthonormal to each other. Why can't we create the quasi-orthonormal set by picking random vectors from the reduced dimension space that match the criterion until we have the number we need? 2) It would be much more interesting to see an example where the number of categories is large so that we can more plainly see the advantage of passing to quasi-orthonormality.

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+has several advantages. First, it is open ended so new categories can be
+added later. Second, the resultant dimensionality can be much lower than
+one-hot encoding. The chief disadvantage is that categories can collide
+if two categories accidentally map into the same hash value. This is a
+*hash collision* and must be fixed separately using a resolution
+mechanism. Bernardi's blog :cite:`hash` provides a good treatment of hash coding.
+
+Finally, baseN encoding is a generalization of binary encoding that uses
+a number base other than 2 (binary). Table :ref:`tab:ternary` is an example of the Japanese
+car makes using base 3,
```
himito:
```suggestion
car makes using base 3.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+.. math::
+   :label: eq:qsoftmax
+
+   \mathrm{qsoftmax}({\bf z})_i={e^{{\bf z}\cdot {\bf q}_i}\over \sum_{j=1}^K
+   e^{{\bf z}\cdot {\bf q}_j}}
+
+The only real difference in the formulation is that while still
+operating in :math:`{\mathbb R}^n` we are encoding :math:`K>n` labels.
+
+Returning to our example of Japanese car makes, Table :ref:`qoe` shows one-hot encoding and QOE of the five manufacturers. In the table, encodings are represented simply as vectors where :math:`\mathbf{u}_i` are unit vectors in :math:`\mathbb{R}^5` and
```
himito:
```suggestion
Returning to our example of Japanese car makes, table :ref:`qoe` shows one-hot encoding and QOE of the five manufacturers. In the table, encodings are represented simply as vectors where :math:`\mathbf{u}_i` are unit vectors in :math:`\mathbb{R}^5` and
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+To better exploit built-in functions of these packages, we define the following :math:`n\times K` *change of coordinates* matrix
+
+.. math::
+
+   \mathbf{Q}=  \left[\begin{matrix} 
+   \bigg| & \bigg| & &\bigg | \\ 
+   \mathbf{q}_1 & \mathbf{q}_2 & \cdots & \mathbf{q}_K \\
+   \bigg| & \bigg| & &\bigg | \end{matrix}\right].
+
+This transformation makes it easier to convert a set of parallel operations into matrix operations for which these aforementioned computational packages are well suited. Mathematically, the transformation maps the representation of a category encoded by QOE to a vector representing one hot encoding. Understanding this transformation makes it a simple matter to express and ``argmax`` or ``softmax`` function's quasiorthoromal variant by equations :ref:`eq:convert:argmax and :ref:`eq:convert`, respectively.
```
himito:
```suggestion
This transformation makes it easier to convert a set of parallel operations into matrix operations for which these aforementioned computational packages are well suited. Mathematically, the transformation maps the representation of a category encoded by QOE to a vector representing one hot encoding. Understanding this transformation makes it a simple matter to express and ``argmax`` or ``softmax`` function's quasiorthoromal variant by equations :ref:`eq:convert:argmax` and :ref:`eq:convert`, respectively.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+and
+
+.. math::
+   :label: eq:convert
+
+   \mathrm{qsoftmax}(\mathbf{z}) = \mathrm{softmax}(\mathbf{Qz}).
+
+The ``tensorflow`` and ``pytorch`` both supply optimized ``softmax`` functions as does ``scipy`` when using ``numpy``, making implementation of QOE not only easy but efficient. Not only will using native functions accelerated performance it can exploit features such as auto differentiation built into the native functions. A useful property if one wishes to use the qsoftmax function as an activation function.
+
+Since the matrix manipulation operations and input/output shape definitions differ from package to package, we provide a qsoftmax implementation in several popular packages. In order to facilitate the most general format possible, in our examples, we will express the quasiorthogonal basis as an list of list, but the input and the output is expressed in the appropriate native class (e.g. :code:`numpy.ndarray` in ``numpy``).
```
himito:
```suggestion
Since the matrix manipulation operations and input/output shape definitions differ from package to package, we provide a qsoftmax implementation in several popular packages. In order to facilitate the most general format possible, in our examples, we will express the quasiorthogonal basis as a list of list, but the input and the output is expressed in the appropriate native class (e.g. :code:`numpy.ndarray` in ``numpy``).
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+For ``numpy``, the implementation is straight-forward and follows equation :ref:`eq:convert` almost literally and is given below.
+
+.. code:: python
+
+    def qsoftmax(x, basis):
+        qx = np.matmul(np.asarray(basis),x)
+        return softmax(qx)
+
+Since ``qsoftmax`` given above requires the basis as a parameter as well as the input vector, it is a parameterized activation function. In many packages, only unparameterized functions can be used. The following function factory or metafunction can be used to return a function ``qsoftmax`` function for a given basis, rather than encoding the function above in a ``lambda`` expression.
```
himito:
```suggestion
Since ``qsoftmax`` given above requires the basis as a parameter as well as the input vector, it is a parameterized activation function. In many packages, only unparameterized functions can be used. The following function factory or metafunction can be used to return a  ``qsoftmax`` function for a given basis, rather than encoding the function above in a ``lambda`` expression.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+.. code:: python
+
+   def softmax(x):
+        ex=np.exp(x)
+        return ex/np.sum(ex)
+
+Tensorflow
+~~~~~~~~~~
+
+For ``tensorflow``, the following segment of code is an implementation of the ``qsoftmax``functions. By using native ``tensorflow`` functions, the resultant ``qsoftmax`` function will be automatically differentiated in a backwards neural network pass. It is also worth noting that quite often due to the way ``tensorflow`` performs batch processing,  the input to the activation function is not a vector but an array of vectors as a ``Tensor`` class.
```
himito:
```suggestion
For ``tensorflow``, the following segment of code is an implementation of the ``qsoftmax`` functions. By using native ``tensorflow`` functions, the resultant ``qsoftmax`` function will be automatically differentiated in a backwards neural network pass. It is also worth noting that quite often due to the way ``tensorflow`` performs batch processing,  the input to the activation function is not a vector but an array of vectors as a ``Tensor`` class.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+.. figure:: 3.pdf
+
+   Softmax on a quasiorthogonal basis :label:`fig:qortho`
+   
+It is clear from equation :ref:`eq:gsoftmax` that the exponential term in the numerator  dominates which is why the softmax is effective.  But consider if :math:`\mathbf{z}\cdot\mathbf{u}_i` is much less than zero such as :math:`\mathbf{z}=-5 \mathbf{u}_i`, then the exponential numerator for that term would severely attenuate the output even though the :math:`\mathbf{z}` lies along the same direction as :math:`\mathbf{u}_i`. 
+Since the :math:`\mathbf{z}\cdot\mathbf{u}_j=0` for all other *j*'s, it would still decode to :math:`i`, but because the :math:*i*-th term is so small, any noise could lead to decoding to a different value.
```
himito:
```suggestion
Since the :math:`\mathbf{z}\cdot\mathbf{u}_j=0` for all other *j*'s, it would still decode to :math:`i`, but because the :math:`i`-th term is so small, any noise could lead to decoding to a different value.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+dimensions. One should note that, though we went from 10 to 7
+dimensions, we did not fully exploit the space spanned by the
+quasiorthogonal vector set. Otherwise, we would likely have had the
+similar results if the categorical labels had a cardinality of 28 rather
+than 10.
+
+Conclusion
+----------
+
+These reduced dimensionality codes are not expected to improve accuracy when the training data is plentiful, but to save computation and representation by reducing the dimensionality of the coded category. As an example, in application such as autoencoders and specifically the imputation architectures presented by :cite:`gondara` and :cite:`lu` where the dimensionality not only dictates the number of outputs and inputs but also the number of hidden layers, a reduction in dimensionality has a profound impact on the size of the model used. Beyond that the reduced dimensionality codes such as QOE and spherical codes can address problems such as the curse of dimensionality and HDLSS where for small sample sizes it may improve accuracy.
```
himito:
```suggestion
These reduced dimensionality codes are not expected to improve accuracy when the training data is plentiful, but to save computation and representation by reducing the dimensionality of the coded category. As an example, in applications such as autoencoders and specifically the imputation architectures presented by :cite:`gondara` and :cite:`lu`, where the dimensionality not only dictates the number of outputs and inputs but also the number of hidden layers, a reduction in dimensionality has a profound impact on the size of the model used. Beyond that, the reduced dimensionality codes such as QOE and spherical codes can address problems such as the curse of dimensionality and HDLSS where for small sample sizes it may improve accuracy.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+Practically speaking, while algorithms to generate spherical codes and quasiorthonormal sets are few, :cite:`sphere` has a vast complication of spherical codes. At the extreme end, a spherical code with 196,560 vectors is available in 24 dimensions, enough to encode nearly 100,000 labels using QOE or 200,000 labels using spherical codes, *in just 24 dimensions!*
+
+In sum, the advantages of QOE and spherical codes is that they can reduce the dimensionality of the vector representation as compared to one-hot encoding, while not inducing artificial geometric relationships as ordinal or binary codes can. The disadvantage is that the accuracy of decoding an encoded vector in a noisy environment (such as classification output) is slightly less than one-hot encoding. This tradeoff ability makes QOE and spherical codes useful tools to be included in a data scientists tool box along side other established categorical coding techniques.
```
himito:
```suggestion
In sum, the advantages of QOE and spherical codes are that they can reduce the dimensionality of the vector representation as compared to one-hot encoding, while not inducing artificial geometric relationships as ordinal or binary codes can. The disadvantage is that the accuracy of decoding an encoded vector in a noisy environment (such as classification output) is slightly less than one-hot encoding. This tradeoff ability makes QOE and spherical codes useful tools to be included in a data scientists toolbox alongside other established categorical coding techniques.
```

himito (commented):
There are still quite a few spelling errors. Please review the entire document again.

summerswallow-whi:
> Hi @summerswallow-whi, I will be reviewing your paper as well. I found it very interesting. I agree that the paper could use an editing pass or two purely for language. I will try to leave more detailed suggestions, but they will not be comprehensive. A couple more substantive points: 1) You mention in the beginning that random vectors are w.h.p. quasi-orthonormal to each other. Why can't we create the quasi-orthonormal set by picking random vectors from the reduced dimension space that match the criterion until we have the number we need? 2) It would be much more interesting to see an example where the number of categories is large so that we can more plainly see the advantage of passing to quasi-orthonormality.

On your first point, if you think it requires an explanation, I can add a paragraph, but picking two vectors at random in a high dimensional space will be on average quasiorthogonal, but when you try a large number of vectors like 1000. Odds of all of them being mutually quasiorthogonal is not so great. Believe me that was the first approach I took.

On the second point, I guess my point to this paper is to make it more accessible to people playing around with this new (or rather revisited concept) everybody getting into ML has started out with the MNIST handwriting db. The code and examples can easily be run in a few minutes by any reader (plus the notebooks will be available at the link in the paper though I have not made it public yet)

summerswallow-whi:
> There are still quite a few spelling errors. Please review the entire document again.

I will take another pass within the next day or two. So if I'm a little slow in this regard. I apologize in advance. (I'm involved with 3 scipy papers + another paper for a conference in July)

githomin:
> On your first point, if you think it requires an explanation, I can add a paragraph, but picking two vectors at random in a high dimensional space will be on average quasiorthogonal, but when you try a large number of vectors like 1000. Odds of all of them being mutually quasiorthogonal is not so great. Believe me that was the first approach I took.

Yes, a paragraph would be great. The following snippet: 

```
def quasi_orthonormal(d, m, eps=0.25): 
    vectors = [] 
    for _ in range(m): 
        candidate = np.power(-1, np.random.randint(2, size=d)) / np.sqrt(d) 
        if all([np.inner(v, candidate) < eps for v in vectors]):         
            vectors.append(candidate) 
    return vectors 
```

gives me sets of quasi-orthonormal vectors comparable to those from the spherical codes in your example.

`[len(quasi_orthonormal(10, 100)) for _ in range(10)]` returns `[15, 12, 13, 13, 16, 13, 15, 13, 13, 14]`                                                                                                   

> On the second point, I guess my point to this paper is to make it more accessible to people playing around with this new (or rather revisited concept) everybody getting into ML has started out with the MNIST handwriting db. The code and examples can easily be run in a few minutes by any reader (plus the notebooks will be available at the link in the paper though I have not made it public yet)

Yes, that makes a lot of sense to me.

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+Extending to Spherical Encodings
+--------------------------------
+
+A Deeper Look at Softmax
+~~~~~~~~~~~~~~~~~~~~~~~~
+In principle, to recover a category from a potentially noisy encoded vector, the dot product of the encoded vector against each basis vector in accordance with equation :ref:`eq:argmax` whether the basis is orthonormal or quasiorthonormal. If one takes a deeper dive into equations :ref:`eq:gsoftmax` and :ref:`eq:qsoftmax`, it is interesting to see what these functions are doing. Figure :ref:`fig:ortho` shows on the left, randomly selected values in a circle of radius 6. On the right shows the vectors after the softmax function is applied. Clearly with a few stragglers, most points either move very close to either of the basis vectors :math:`(0,1)` or :math:`(1,0)`. For this graphic the radius of 6 is chosen because upon examining the input to the softmax function in the example above, the average magnitude of each component is 5.5. In reality, 6 is probably a conservatively low value. Due to the exponential term in the softmax function, even large components would make the separation more pronounced.
```
githomin:
```suggestion
In principle, to recover a category from a potentially noisy encoded vector, we take the dot product of the encoded vector and each basis vector (whether orthonormal or quasiorthonormal) in accordance with equation :ref:`eq:argmax`. If one takes a deeper dive into equations :ref:`eq:gsoftmax` and :ref:`eq:qsoftmax`, it is interesting to see what these functions are doing. Figure :ref:`fig:ortho` shows on the left, randomly selected values in a circle of radius 6. On the right are the vectors after the softmax function is applied. Clearly with a few stragglers, most points either move very close to either of the basis vectors :math:`(0,1)` or :math:`(1,0)`.
```

githomin: I don't fully understand the point of the last couple sentences. What does the radius of the circle signify?

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+
+.. figure:: 1.pdf
+
+   Softmax on an orthonormal basis :label:`fig:ortho`
+
+Similarly, figure :ref:`fig:qortho` shows on the the same type of distribution of randomly selected values and the right shows the effect after a quasiorthonormal softmax is applied with three basis vectors. Since the qsoftmax function would map the two dimensional input into a three dimensional space, for graphics sake, the three dimensional vectors are mapped back down to two dimensions using the quasiorthonormal basis. Again with the exception of a few stragglers, most points move very close to one of the three basis vectors.
```
githomin:
```suggestion
Similarly, figure :ref:`fig:qortho` shows the same type of distribution of randomly selected values and the right shows the effect after a quasiorthonormal softmax is applied with three basis vectors. Since the qsoftmax function maps the two dimensional input into a three-dimensional space, the three-dimensional vectors are mapped back down to two dimensions using the quasiorthonormal basis. Again with the exception of a few stragglers, most points move very close to one of the three basis vectors.
```

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,677 @@
...
+Once again we see the power of the exponential term and most points move very close to one of the four encoding vectors.
+
+Of course nothing comes for free, if a prediction gets confused
+between two antipodal unit vectors, the result could be that they cancel
+out and allow the noise to dictate the resulting category. By contrast,
+for one-hot encoding, the result would get decoded as one of the two
+possible values.
+
+With this risk in mind, we can further extend the idea to a
+quasiorthogonal basis by adding the antipodal vectors for each vector in
```
githomin: I think we can sum up these paragraphs into one. The issue isn't whether or not we have negative numbers in our basis. The issue is more that the number of basis vectors we use is higher than the dimension and thus there can be more ambiguity than with one-hot encodings where the two numbers are the same.

papers/75_lu/75_lu.rst
summerswallow-whi: Since (on monitoring the inputs to the softmax I found the output to be on average about 5.5. Part of this was justification as to why I chose 6 to randomly select from rather than 1 or 40. 
 I guess that wasn't well expressed

summerswallow-whi:
> Yes, a paragraph would be great. The following snippet:
> 
> ```
> def quasi_orthonormal(d, m, eps=0.25): 
>     vectors = [] 
>     for _ in range(m): 
>         candidate = np.power(-1, np.random.randint(2, size=d)) / np.sqrt(d) 
>         if all([np.inner(v, candidate) < eps for v in vectors]):         
>             vectors.append(candidate) 
>     return vectors 
> ```
> 
> gives me sets of quasi-orthonormal vectors comparable to those from the spherical codes in your example.
> 
> `[len(quasi_orthonormal(10, 100)) for _ in range(10)]` returns `[15, 12, 13, 13, 16, 13, 15, 13, 13, 14]`
> 

First of all, I'm glad to see you take an interest in this. Originally, I did use a brute force approach, but quickly found that it would not scale. I just added two paragraphs explaining why this is not a scalable solution. I did catch myself beginning to go into a rabbit hole of explanation. The probability argument given is probably the strongest reason, but there are computation complexity and geometric issues too. If I don't stop myself I could go on for pages.

summerswallow-whi:
>  I will be reviewing your paper as well. I found it very interesting, and I will be leaving detailed comments soon. One quick note. I don't think all the tables are rendering correctly, e.g., ref:binary and ref:onehot
@githomin  At least on my local build the tables look fine. I'll pull the PDF from procbuild to see if I see any problems there. 
Didn't find a problem with those tables, but I did find a couple of errors which I fixed.

papers/75_lu/75_lu.rst
summerswallow-whi: I'm all for being succinct as I know I can be long winded when I'm uncertain how to make a point. I guess the point being made here is the reason originally for considering QO vectors only is that the dot product is small. Though an oversimplification SC are QO vectors with their negatives. Turns out if  v is normal then  the dot product between v and -v would not be small but -1. The softmax however does separate these since the exponential terms squash the negatives. The third paragraph just cautions that like in the case of ordinal coding there if v and -v are unrelated categories you've now induced a relationship (one which softmax separates) but could have other statistical implications.

To be honest, when I saw the figures (which I was used to make sure the code was working), I saw the pattern and asked, "I wonder if this would work".

deniederhut:
Hey @summerswallow-whi  ! Thanks to some awesome behind the scenes work by @stargaser, we now have experimental support for publishing papers with ORCIDs. If you have one and would like to add it to your paper, it's as simple as adding an `:orcid:` tag under your name in the paper header. [You can see an example of how to do this in the `00_vanderwalt` example paper](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2020/papers/00_vanderwalt/00_vanderwalt.rst). If you don't have one, no worries! You can still publish your paper as is.

githomin:
> > Yes, a paragraph would be great. The following snippet:
> > ```
> > def quasi_orthonormal(d, m, eps=0.25): 
> >     vectors = [] 
> >     for _ in range(m): 
> >         candidate = np.power(-1, np.random.randint(2, size=d)) / np.sqrt(d) 
> >         if all([np.inner(v, candidate) < eps for v in vectors]):         
> >             vectors.append(candidate) 
> >     return vectors 
> > ```
> > 
> > 
> > gives me sets of quasi-orthonormal vectors comparable to those from the spherical codes in your example.
> > `[len(quasi_orthonormal(10, 100)) for _ in range(10)]` returns `[15, 12, 13, 13, 16, 13, 15, 13, 13, 14]`
> 
> First of all, I'm glad to see you take an interest in this. Originally, I did use a brute force approach, but quickly found that it would not scale. I just added two paragraphs explaining why this is not a scalable solution. I did catch myself beginning to go into a rabbit hole of explanation. The probability argument given is probably the strongest reason, but there are computation complexity and geometric issues too. If I don't stop myself I could go on for pages.

Your probably argument only applies if you take a set of size k and want them to be all mutually quasi-orthogonal. If instead you do something like my snippet (take a set of size m > k), you can just use the union bound, and the probability that you find a set of size k remains high.

By the way, I think this paper is great, and introducing quasi-orthonormal encodings to a wide audience is perfect for scipy. However, I found the sections on the spherical codings to be difficult to follow, and I think the paper could be simplified by using random vectors instead.

summerswallow-whi:
> Your probably argument only applies if you take a set of size k and want them to be all mutually quasi-orthogonal. If instead you do something like my snippet (take a set of size m > k), you can just use the union bound, and the probability that you find a set of size k remains high.

Now I understand what you are talking about and I tried that approach to. Here's the problem with the union approach. There is a combinatorial trap. I'm going to make the spherical code argument because it is easier to visualize. Suppose you want to get a good sphere code in 2 dimensions with a minimum maximum angle to be 80degrees (i.e., small dot product). I take your approach and start adding vectors and my first 3 happen to be perfectly distributed having a 120 degree angle between the three. However, no matter where I try to stick the fourth vector I can never do better than 60 degrees. But clearly there is a solution with the angle being 90 degrees.

> By the way, I think this paper is great, and introducing quasi-orthonormal encodings to a wide audience is perfect for scipy. However, I found the sections on the spherical codings to be difficult to follow, and I think the paper could be simplified by using random vectors instead.

Let me take another look at spherical encodings section. I think the three paragraphs in question had more to do with the way I originally conceptualize the idea, so the logic may be a bit muddled. In a nutshell, how I came to try spherical codes is I saw the graphs for `softmax` while testing it. Noticed nothing really happened in the negative quadrants and thought maybe the negative vectors could be used, tried it and results were on par with QOE.

You bring up an interest question. How well does random vectors on a sphere work? 
In the companion poster presentation, I have a notebook designated as "Just for Fun". I'll try this idea. It'll be posted at the link in the paper.

deniederhut:
Looks like the remote build is currently failing with `(ERROR/3) Unexpected indentation.` Is this still building locally for you?

summerswallow-whi:
> Looks like the remote build is currently failing with `(ERROR/3) Unexpected indentation.` Is this still building locally for you?

Nope. I had a co worker proofread. He edited directly. Must have done something. I'll fix it.

summerswallow-whi:
@deniederhut Thanks for bringing to my attention. Talk about a needle in a haystack. But I did find it.

papers/75_lu/75_lu.rst
summerswallow-whi: I rewrote the last couple of sentences.

papers/75_lu/75_lu.rst
summerswallow-whi: I rewrote this. Hopefully this is better

papers/75_lu/75_lu.rst
```diff
@@ -642,28 +643,44 @@ Similarly, figure :ref:`fig:qortho` shows the same type of distribution of rando
...
-reducing dimensionality by encoding to not just the basis a vector but also its
-antipodal vector. To further our graphical example, in figure
-:ref:`fig:sphere`, we use :math:`(1,0)`, :math:`(0,1)` and their antipodal
-vectors :math:`(-1,0)` and :math:`(0,-1)` to encode values and apply a softmax
-using those vectors.
+Because the expectation on one-hot encoding is that the value of a
+given vector component be either :math:`0` or :math:`1` and that
+negative values are not expected even in a noisy environment. This is
+evident in figure :ref:`fig:ortho`, where the results are all in the
+first quadrant (i.e. no negative values). This begs the question could
```
githomin: ~begs the question~ raises the question:

papers/75_lu/75_lu.rst
```diff
@@ -642,28 +643,44 @@ Similarly, figure :ref:`fig:qortho` shows the same type of distribution of rando
...
+the negative values be exploited with minimal detrimental effects.
```
githomin: ~effects.~ effects?

papers/75_lu/75_lu.rst
```diff
@@ -0,0 +1,759 @@
...
+
+Suppose, the probability of any two vectors being quasiorthogonal to a
+given :math:`\epsilon` is :math:`p`. Since there are :math:`{k\choose 2}` 
+pairs of vectors, the probability that you have a quasiorthogonal set
+is :math:`p^{k\choose 2}`. To put in concrete terms, if two random
+vectors have a 99% chance of being quasiorthogonal. Picking a set of
+20 is only 14% and 30 is around 1%. Other factors conspire to make
+this difficult including the increasing computational complexity and
+the geometric differences between a cube and sphere become more
+pronounced as :math:`k` and :math:`N` grow.
```
githomin, line 515: It'd be great if you could rewrite the above two paragraphs to suggest the possibility of taking a set of `m > k` vectors and using them which improves the probability considerably.

papers/75_lu/75_lu.rst
summerswallow-whi: fixed in last commit will resolve this

papers/75_lu/75_lu.rst
summerswallow-whi: Fixed in last commit

papers/75_lu/75_lu.rst
summerswallow-whi: I don't understand the suggestion. Are you talking about using an incremental approach?
It's one of I'm sure countless approaches one can try. If you look at section 4 and 5 of the Kainan/Kurkova paper, they suggest ways to construct such as with the hamming cube in your suggested code.

I had a chance to run your code. First of all, thanks for the suggestion and don't take this as a criticism because we all have to start out somewhere exploring this.

First to fairly compare to the example in the paper epsilon is .333 but your `d` should be 7. Second, you left out an absolute value around the inner product. With that running the snippet as follows (I set the iterations up to 10000 for completeness).

```
def quasi_orthonormal(d, m, eps=0.25): 
    vectors = [] 
    for _ in range(m): 
        candidate = np.power(-1, np.random.randint(2, size=d)) / np.sqrt(d) 
        if all([np.abs(np.inner(v, candidate)) < eps for v in vectors]):         
            vectors.append(candidate) 
    return vectors
```

with
```
[len(quasi_orthonormal(7, 10000, .333)) for _ in range(10)]
```
yielded [8, 8, 8, 8, 8, 8, 8, 8, 8, 8]

For 7-dimensions with an epsilon of .3333. You can find 28 vectors.

summerswallow-whi: @githomin are we good here are do you still want some changes?

deniederhut:
Hi @himito and @githomin ! Are you happy with the updates to the paper?

himito:
Hi @deniederhut, 
from my side all it's OK :man_dancing:

githomin:
Hi @deniederhut I think we've done what we can.

papers/75_lu/ourbib.bib
```diff
@@ -0,0 +1,150 @@
...
+
+
+@misc{scikit,
+      title={Category Encoders},
+      url={http://contrib.scikit-learn.org/category_encoders/},
+      author={Will McGinnis},
+      year={2016}
+}
+
+@misc{tensorflow,
```
deniederhut, line 60: I'm not sure this is the preferred citation for tensorflow - https://www.tensorflow.org/about/bib

papers/75_lu/ourbib.bib
```diff
@@ -0,0 +1,150 @@
...
+@misc{hash,
+      title={Don’t be tricked by the Hashing Trick}, 
+      url={https://booking.ai/dont-be-tricked-by-the-hashing-trick-192a6aae3087}, 
+      journal={Booking.com Data Science}, author={Bernardi, Lucas}, year={2018}, month={Jan}
+}
+                                                                                      
+@inproceedings{Gautam2013ANA,
+               title={A Novel Approach to the Spherical Codes Problem},
+               author={Simanta Gautam and Dmitry Vaintrob},
+               year={2013}
```
deniederhut: This reference is missing a conference / editors / etc

papers/75_lu/ourbib.bib
```diff
@@ -0,0 +1,150 @@
...
+      year={2018},
+      month={Sep}
+}
+@misc{kainan,
+      author = {Kainen, Paul},
+      year = {1992},
+      month = {01},
+      pages = {},
+      title = {Orthogonal dimension and tolerance},
+      url= {https://www.researchgate.net/publication/303520147}
```
deniederhut: Researchgate is an article aggregator, not a journal

deniederhut:
Not to be that guy, but 25% of the references in this paper are to wikipedia. I would consider omitting those; or, finding the original references; or, finding a review article that covers some of these topics.

papers/75_lu/ourbib.bib
summerswallow-whi: This wasn't a citation for tensorflow in the overview, it was a pointer to the Quickstart for Beginner's guide from which the examples given in the paper are literally taken from. However, if you think it would be helpful include an additional tensorflow citation I can add it. Just seems to me that since the code mimics that found precisely at the link. I'm certainly willing to do what you think best.

papers/75_lu/ourbib.bib
summerswallow-whi: I looked to see how it was cited by others. They treat it as a technical report so I did the same.

papers/75_lu/ourbib.bib
summerswallow-whi: This was a bit tricky to correct. The bibtex was taken from researchgate's cite. I looked found how others cited this so I corrected it to match.

summerswallow-whi:
> Not to be that guy, but 25% of the references in this paper are to wikipedia. I would consider omitting those; or, finding the original references; or, finding a review article that covers some of these topics.

@deniederhut 
In fairness, only 3 wiki references out of 16. One reference was erroneously labeled as wiki: To address the three remaining ones:
1. Word embeddings. There aren't any good aggregate articles that I could find, but several blog posts. I could substitute one if that is preferable. 
2. Steiner Systems, #1 the original reference is in German and from the 1800's most subsequent articles are very very mathematical and not really accessible. I feel the Wiki article is the best overview for a lay person
3. Curse of Dimensionality. The modern meaning has changed since Bellman coined the phrase so citing the original article would not be appropriate. The wiki article casts the widest overview over the subject, but if a blog article is more desirable I can substitute an appropriate one here.

Since machine learning is rapidly progressing, for overview material it appears that blog posts are the preferred method. 
In short, here are some possible actions
1. Word embeddings - Substitute a blog post: https://machinelearningmastery.com/what-are-word-embeddings/ is pretty good
2. Really the Wiki is the best place to learn about Steiner systems. 
3. "Curse of Dimensionality" I think the Wiki gives the best overview but this blog https://blog.datasciencedojo.com/curse-of-dimensionality-python/ isn't as comprehensive but would be OK if you really don't want a wiki citation.

I will prepare those two blog citations in the mean time and await you advice

papers/75_lu/ourbib.bib
summerswallow-whi: I added a citation to tensorflow, but I think it is important to keep the reference quickstart since that's where the code was taken

deniederhut:
> There aren't any good aggregate articles that I could find, but several blog posts

A quick google search turned up https://link.springer.com/chapter/10.1007/978-3-030-01171-0_12. Another option would be to cite the original word2vec and glove papers.

> the original reference is in German and from the 1800's most subsequent articles are very very mathematical and not really accessible

I would prefer to link a subsequent mathematical article (in English please 😄) even if it is not very accessible. Your readers have the option of looking this up in Wikipedia if the paper is too difficult.

> The modern meaning has changed since Bellman coined the phrase so citing the original article would not be appropriate

Hm. Maybe the answer here is that the curse of dimensionality is such a well-known phenomenon that it does not require citation. Perhaps instead of references you might add a footnote to a blog post aimed at a general audience.

summerswallow-whi:
gle search turned up https://link.springer.com/chapter/10.1007/978-3-030-01171-0_12. Another option would be to cite the original word2vec and glove papers.

I'll use that reference. Good overview. Since this isn't the point of the paper. That's fine.

> I would prefer to link a subsequent mathematical article (in English please 😄) even if it is not very accessible. Your readers have the option of looking this up in Wikipedia if the paper is too difficult.

The Design theory book is a bit mathematical, but it'll will serve this purpose

> Hm. Maybe the answer here is that the curse of dimensionality is such a well-known phenomenon that it does not require citation. Perhaps instead of references you might add a footnote to a blog post aimed at a general audience.
Ok. I added a footnote. 

@deniederhut Let me know if this fixes your concerns

deniederhut:
Thanks!

PR #553, opened 2020-05-29
Paper: SHADOW algorithm reference and testing framework

myxie:
process.

for more information.


deniederhut:
Looks like the paper build is currently failing with `Error: File 'fig/sample_dag-eps-converted-to.pdf' not found`.

myxie:
Hi @deniederhut, 

It looks like there are issues when converting from EPS format to PDF (I think epstopdf can be funny if there are not pre-defined figure directories in the document preamble). I have updated the figures to PNGs and it looks like it is building correctly now. 

Thank you for the heads up! 

Ryan.

deniederhut:
Yay!

AngelosPsy:
Dear @myxie,

I am one of the reviewers of this paper and I will send you my first comments  before the 18th.

Angelos

stsievert:
I've also been assigned as a reviewer to [this paper][1], and will also have some comments by the 18th.

[1]:http://procbuild.scipy.org/download/myxie-2020

davekits:
The SHADOW algorithm seems, acceptable workflow with new perceptive however, is any scheduling change in arrangement relates with WALL time in Linux machine?

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+addresses this issue by providing a workflow-oriented algorithm library
+and testing environment, in which the performance of single- and
+multi-objective workflow scheduling algorithms may be compared to
+implementations of de facto algorithms. To the best of our knowledge, there is
+no single-source repository of implementations of DAG or Workflow scheduling
+algorithms. The emphasis in SHADOW is on reproducibility   and accuracy in
+algorithm performance analysis, rather than a simulated demonstration of the
+application of a particular algorithm in certain environments. Additionally,
+with the popularity of Python in other domains that are also growing within
+the workflow community - such as Machine and Deep Learning - SHADOW provides
```
stsievert: Style nit: use an em-dash ("–"), not a dash ("-") for these kind of breaks. In plain latex, this is "--" instead of "-". I don't think that works in this conversion.

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+
+------------------------------------------------------------------------------------------------------------
+SHADOW: An algorithm reference and testing framework for scheduling data-intensive workflows
+------------------------------------------------------------------------------------------------------------
+
+.. class:: abstract
+
+  As the scale of science projects increase, so does the demand on computing
+  infrastructures. Data-intensive workflows and science pipelines that are being
+  developed for new, exa-scale projects (e.g. the SKA) necessitate the
```
stsievert: I would define SKA before referring to it by acronym.

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+    mConcatFit & 0.37 & 0.00 & 0.00 & 0.00 & 0.00 & 0.00 & 0.01 & 0.00 & 0.13 & 0.00 \\
+    & 1.00 & 0.00 & 0.00 & 0.00 & 0.00 & 0.00 & 0.03 & 0.00 & 0.25 & 0.00 \\
+    mBackground & 9.42 & 3.56 & 45.78 & 1.86 & 44.30 & 1.70 & 88.65 & 1.75 & 46.32 & 0.55 \\
+    mImgtbl & 0.12 & 0.06 & 0.06 & 0.02 & 0.01 & 0.00 & 0.35 & 0.02 & 0.15 & 0.00 \\
+    mAdd & 12.50 & 6.11 & 48.83 & 13.41 & 34.34 & 8.70 & 0.70 & 0.08 & 0.37 & 0.00 \\
+    mShrink & 2.75 & 1.93 & 17.15 & 0.30 & 0.02 & 0.00 & 0.18 & 0.00 & 0.09 & 0.00 \\
+    mJPEG & 0.00 & 0.00 & 0.06 & 0.00 & 0.00 & 0.00 & 0.00 & 0.00 & 0.19 & 0.00 \\
+    \end{tabular}
+    \caption{Updated relative cost values using the min-max feature scaling method described
+    in Equation :ref:`normalise`}
```
stsievert: Typo: in the compiled PDF, this displays as "... described in Equation ."

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+      self.solution_cost = 0
+
+
+Visualiser
+~~~~~~~~~~
+
+SHADOW provides wrappers to ``matplotlib`` that are structured around the
+``Workflow`` and ``Solution`` classes. The ``Visualiser`` uses the
+``Solution`` class to retrieve allocation data, and generates a plot based on
+that information. For example, Figure  is the result of visualising the
```
stsievert: Typo: The figure isn't referenced. I think you need a `:label:` definition in the `samplea_allocation.pdf` figure below.

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+These components are all contained within the main ``shadow`` directory;
+there are also additional codes that are located in ``utils``, which are covered in the **Additional Tools** section.
+
+.. figure:: fig/heft_with_calc.png
+  :scale: 80 %
+
+  An example workflow DAG adapted from; weights on the edges describe data
+  products from the respective parent node being sent to the child. In SHADOW,
+  task computation cost is represented by  the total number of Floating Point
+  Operations required to run the task (see Table ). This is intended to
```
stsievert: Typo: the table isn't referenced. "(see Table )".

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+directory, as well as the implemented sample algorithms in the ``algorithms``
+module. The ``algorithms`` module is concerned with the implementations of
+algorithms, with the intention of providing both a recipe for implementing
+algorithms using SHADOW components, and benchmark implementations for
+performance analysis and testing. The visualiser is a useful way to add
+graphical components to a benchmarking recipe, or can be invoked using the
+command line interface to quickly run one of the in-built algorithms.
+
+These components are all contained within the main ``shadow`` directory;
+there are also additional codes that are located in ``utils``, which are covered in the **Additional Tools** section.
```
stsievert, line 208: Do you intend for the first two paragraphs of the "Models" subsection to be above the subsection? It looks like the first two paragraphs give an overview of the following subsections. (by first two paragraphs I mean "The `models` module is likely the main entry point ... which are covered in the Additional Tools section.")

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+LIGO) had their task runtime, memory and I/O rates profiled, from which they
+created a WorkflowGenerator tool [#]_. This tool uses the distribution sizes
+for runtime etc., without requiring any information on the hardware on which
+the workflows are being scheduled. This means that the characterisation is
+only accurate for that particular hardware, if those values are to be used
+across the board; testing on heterogeneous systems, for example, is not
+possible unless the values are to be changed.
+
+This is dealt with in varied ways across the literature. For example,
+:cite:`rodriguez2018` use the distributions rom :cite:`bharathi2008` paper,
```
stsievert: Typo: 

``` diff
- use the distributions rom 
+ use the distributions from 
```

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+  Operations required to run the task (see Table ). This is intended to
+  alleviate the difficulty of converting the runtime between  different test
+  environment configurations. :label:`heftcalc`
+
+
+The ``models`` module provides the ``Workflow`` class, the foundational data
+structure of shadow. Currently, a ``Workflow`` object is initialised using a
+JSON configuration file that represents the underlying DAG structure of the
+workflow, along with storing different attributes for task-nodes and edges
+:ref:`heftcalc`.
```
stsievert: In the compiled PDF, this shows as  "along with storing different attributes for task-nodes and edges 2." Perhaps this diff is intended:

``` diff
- along with storing different attributes for task-nodes and edges :ref:`heftcalc`
+ along with storing different attributes for task-nodes and edges shown in Figure :ref:`heftcalc`
```

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+     & \multicolumn{2}{c}{Runtime}&  \multicolumn{2}{c}{
+    I/O Read} & \multicolumn{2}{c} {I/O Write} & \multicolumn{2}{c}{Peak Memory} &
+    \multicolumn{2}{c}{CPU Util}\\
+    job &  Mean (s) & Std. Dev. & Mean (MB) & Std. Dev. & Mean (MB)& Std. Dev. &
+    Mean (MB)& Std. Dev. & Mean (\%) & Std. Dev \\
+    \hline
+    mProject  PP & 9.47 & 0.49 & 11.22 & 0.38 & 44.30 & 1.70 & 64.66 & 1.75 & 476.20 & 0.16 \\
+    mDiffFit & 10.61 & 9.00 & 266.27 & 8.52 & 10.29 & 7.40 & 92.61 & 10.77 & 456.48 & 2.57 \\
+    mConcatFit & 0.37 & 0.00 & 0.00 & 0.00 & 0.00 & 0.00 & 0.01 & 0.00 & 0.13 & 0.00 \\
+    & 1.00 & 0.00 & 0.00 & 0.00 & 0.00 & 0.00 & 0.03 & 0.00 & 0.25 & 0.00 \\
```
stsievert: Typo: This row is missing a "job" element. I think this diff will resolve the problem:

``` diff
- & 1.00 & 0.00 & 0.00
+ mBgModel & 1.00 & 0.00 & 0.00
```

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+  An example workflow DAG adapted from; weights on the edges describe data
+  products from the respective parent node being sent to the child. In SHADOW,
+  task computation cost is represented by  the total number of Floating Point
+  Operations required to run the task (see Table ). This is intended to
+  alleviate the difficulty of converting the runtime between  different test
+  environment configurations. :label:`heftcalc`
+
+
+The ``models`` module provides the ``Workflow`` class, the foundational data
+structure of shadow. Currently, a ``Workflow`` object is initialised using a
```
stsievert, line 226: Did you mean to capitalize "shadow"?

``` diff
- shadow
+ SHADOW
```

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+
+   from shadow.models.environment import Environment
+   env = Environment('sys.json')
+   HEFTWorkflow.add_environment(env)
+
+
+
+The ``Machine`` class is also defined in environment.py - this is a
+helper-class that makes developer access to specifications (provided compute,
+memory, bandwidth etc.) convenient and intuitive. The Workflow class
```
stsievert: Why do you include the line detailing the the `Machine` class? I don't see any reference of the `Machine` class until (possibly) in the `NSGASolution` example.

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+    """
+    Implementation of the original 1999 HEFT algorithm.
+
+    :params wf: The workflow object to schedule
+    :returns: The makespan of the resulting schedule
+    """
+    upward_rank(workflow)
+    workflow.sort_tasks('rank')
+    makespan = insertion_policy(workflow)
+    return workflow.solution
```
stsievert, line 480: What's the purpose of calling `upward_rank`? Is it an impure function that modifies `workflow`?

In the paper, you say the `heft` should return "the solution object attached to the workflow;" however, that doesn't match the docstring of this function. If this algorithm returns the a solution object, what's the purpose of defining `makespan`?

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+graphical components to a benchmarking recipe, or can be invoked using the
+command line interface to quickly run one of the in-built algorithms.
+
+These components are all contained within the main ``shadow`` directory;
+there are also additional codes that are located in ``utils``, which are covered in the **Additional Tools** section.
+
+.. figure:: fig/heft_with_calc.png
+  :scale: 80 %
+
+  An example workflow DAG adapted from; weights on the edges describe data
```
stsievert: Typo: "An example workflow DAG adapted from; weight on the edges..."

Is this DAG a "de facto standard"?

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,787 @@
...
+
+.. figure:: fig/heft_example.png
+
+   This is a replication of the costs provided in the original HEFT paper
+   :cite:`topcuoglu2002`. The table shows a different runtime for each
+   task-machine pairing. :label:`heftnocalc`
+
+This approach is less flexible for
+scheduling workflows, but is a common approach used in the scheduling
+algorithm literature; an example of this is shown in Figure
```
stsievert: I'd appreciate some citations to support the claim "This approach ... is a common approach used in the scheduling algorithm literature."

stsievert (commented):
Thanks for the submission @myxie! This submission looks to be a pretty technical description of some software to solve the "science workflow scheduling" problem, which manages scheduling large scale enterprise/science problems on a distributed system. Given the authors apparent experience at the Square Kilometer Array (SKA), I suspect this software is used in the processing of data SKA collects.

Some questions I have or had:

* What exactly is "science workflow scheduling"? This paper didn't mention it, and I had to look at the references to answer my question. I'd appreciate a paragraph explaining science workflow scheduling after "Science workflow scheduling is a field with varied contributions in algorithm development and optimization, which address a number of different sub-problems within the field."
* What's your use case of science workflow scheduling? I had a hard time mapping the use of the software to a real world problem, and the constraints that might be seen. I suspect the motivating use case is with the SKA. How large is the data? How many machines do you have? What's a qualitative description of the machines?
* Why is the DAG in figures 2 and 3 a "de facto standard"? I presume this a "test against [a] de facto standard" as mentioned in the abstract. What problem is the DAG aimed at solving? What produced the (hypothetical?) data? How large is the data? I tried to look at the reference, but it's behind a paywall.

Some smaller and more detailed comments on specific lines are below:

msarahan:
Hi @myxie, I'm another reviewer. This is pretty far outside my area of expertise, but let me try to ask some intelligent questions.

Given the wide variety of hardware capability, how easy is it to add additional attributes, such as CPU instruction sets or GPU capabilities, as parameters in your model? Are these even necessary to include in your model, or are they more or less abstracted away in any benchmarking you may run?

Since you vaguely mention monetary cost as something that can be a metric, it might be fun to show a demo of optimizing workflows using different cloud compute resources, for example, such that you can tweak which nodes you use whether you want top speed or lowest price, or somewhere in between, and how you would schedule them. This is by no means critical or important, so don't do lots (any?) extra work for it. I only thought it might help demonstrate the general usefulness of your tool.

myxie:
> Thanks for the submission @myxie! This submission looks to be a pretty technical description of some software to solve the "science workflow scheduling" problem, which manages scheduling large scale enterprise/science problems on a distributed system. Given the authors apparent experience at the Square Kilometer Array (SKA), I suspect this software is used in the processing of data SKA collects.

Thank you for taking the time to review the submission @stsievert! I really appreciate your time and comments, and looking forward to taking the on board. 

> Some questions I have or had:
> 
>     * What exactly is "science workflow scheduling"? This paper didn't mention it, and I had to look at the references to answer my question. I'd appreciate a paragraph explaining science workflow scheduling after "Science workflow scheduling is a field with varied contributions in algorithm development and optimization, which address a number of different sub-problems within the field."

This makes sense - I have some paragraphs dedicated to 'Workflow scheduling' in the following section, but I have not adequately addressed the relationship between Workflow scheduling and Science workflow scheduling (and task scheduling!). 

In terms of the positioning of this explanation, I see two options: 

1. Have a second paragraph after the 'Workflow scheduling' one that describes how science-workflow scheduling is related to workflow/task scheduling, or
2. As you have suggested, define it when it is mentioned in the introductory paragraphs. 

>     * What's your use case of science workflow scheduling? I had a hard time mapping the use of the software to a real world problem, and the constraints that might be seen. I suspect the motivating use case is with the SKA. How large is the data? How many machines do you have? What's a qualitative description of the machines?

SHADOW is intending to solve more of a 'meta-problem' with respect to task and workflow scheduling; namely, the lack of algorithm implementations, and a framework to test the performance of new algorithms against existing ones. You are right that it was born out of scheduling problems we are investigating for the SKA, where we are looking at scheduling approaches to deal with the data. However, in order to demonstrate that the existing approaches don't work as well as an SKA-specific approach, we need existing algorithms to benchmark against. This was discussed in the last 'paragraph' of the introduction, but I think this passage may need more prominence to ensure the point of the framework gets across (around line 90 in the git diff). 

In addition, there are others in the literature who have expressed a desire for a software framework that wraps up common bench-marking metrics and provides an environment where algorithms may be tested on the 'same playing field' (Maurya and Tripathi 2018). So the intention is that SHADOW is targeted towards task scheduling and workflow scheduling algorithm development in-general, as opposed to the SKA specifically. Ideally, SHADOW is environment/problem agnostic, provided the algorithm you are implementing is a DAG-based scheduling problem. 

On reflection, this is confusion is likely added to by the name SHADOW itself - this was a working title based and I am not attached to it in any way, so if you think it would be prudent to adjust this to more accurately reflect the 'algorithm benchmarking/reference library' nature of the project, I will happily do so. 

>     * Why is the DAG in figures 2 and 3 a "de facto standard"? I presume this a "test against [a] de facto standard" as mentioned in the abstract. What problem is the DAG aimed at solving? What produced the (hypothetical?) data? How large is the data? I tried to look at the reference, but it's behind a paywall.

This is poor separation of concerns on my behalf; the DAG in figures 2 and 3 is taken from the Topcuoglu et al. 2002 paper (the original paper for the HEFT algorithm), which has then been used in a number of other publications as a useful diagram to introduce the scheduling algorithm. It is not used as a benchmarking tool, just a visual guide. 

The de-facto standard workflows that I refer to in the context of SHADOWGen and the DAX translator are those discussed in Bharathi et al. (2008) and Juve et al. (2013), which have become the common workflows on which to profile new algorithm's performance.  

I will be able to further clarify this with (I think) minimal effort - if you have anything specific advice on how you would improve the clarity here, I would appreciate it!  
> 
> Some smaller and more detailed comments on specific lines are below:
I'll address these in the comment :)

papers/ryan_bunney/paper.rst
myxie: > What's the purpose of calling `upward_rank`? Is it an impure function that modifies `workflow`?

Yes - it is one of two parts of the 'HEFT' algorithm, which ranks all nodes in the task-DAG before using the `insertion_policy`. `upward_rank` modifies the `rank` variable of each task in the workflow. 

> In the paper, you say the `heft` should return "the solution object attached to the workflow;" however, that doesn't match the docstring of this function. If this algorithm returns the a solution object, what's the purpose of defining `makespan`?

This reflects a refactor, which changed the return value of each heuristic call when I added the concept of a `Solution` object. Unfortunately, I did not change the docstrings to match. I will push a commit later to update this, and edit the code-snippet accordingly. Thank you for picking up on this!

papers/ryan_bunney/paper.rst
myxie: Not quite - I have answered this in more detail in reply to your larger comment.

papers/ryan_bunney/paper.rst
myxie: Yep, absolutely. Thank you for picking up on this.

papers/ryan_bunney/paper.rst
myxie: I certainly did...

papers/ryan_bunney/paper.rst
myxie: I'll double check, but I think it is "---" in RST. I will update it locally and see what it looks like before committing again.

papers/ryan_bunney/paper.rst
myxie: Yes, I would too. Oops. (Thank you!)

myxie:
> Hi @myxie, I'm another reviewer. This is pretty far outside my area of expertise, but let me try to ask some intelligent questions.

Hi @msarahan, thank you for taking the time to review this! 

> Given the wide variety of hardware capability, how easy is it to add additional attributes, such as CPU instruction sets or GPU capabilities, as parameters in your model? Are these even necessary to include in your model, or are they more or less abstracted away in any benchmarking you may run?

For the purpose of task scheduling, CPU instruction sets are probably a bit too low-level for the type of analysis we would typically be running (e.g. how long does one set of tasks take, can we schedule this before a deadline, what is the least-expensive resources provision vs time it takes to run tasks etc.). There is definitely the possibility of using CPU/GPU specifications (e.g. FLOPs/Cycle, Cycles/Second), and then having a small tool that calculates the number of FLOPs/second this gives us - but this seems slightly different to what you are suggesting/asking. So, to answer your question, I would agree that they are abstracted to a higher-level than the CPU instruction set

There is definitely the ability to add hardware and flag it as a GPU, and then have application Tasks that are GPU-specific. At the moment, this would require some work directly with the `NetworkX` graph, but moving forward there will be explicit support for GPU-specific tasks. 

> Since you vaguely mention monetary cost as something that can be a metric, it might be fun to show a demo of optimizing workflows using different cloud compute resources, for example, such that you can tweak which nodes you use whether you want top speed or lowest price, or somewhere in between, and how you would schedule them. This is by no means critical or important, so don't do lots (any?) extra work for it. I only thought it might help demonstrate the general usefulness of your tool.

This is definitely something I will be preparing for the poster submission I'll be making to the conference! Because SHADOW is more of a framework that provides the Workflow model and example algorithm implementations, it will be demonstrating less how particular algorithms may work, and more a comparison between how different algorithms, when doing the mulit-objective optimization, may make the trade-offs. I may be able to scrape together an example figure by the time the edits are due...

AngelosPsy:
Hello @myxie,

This is an interesting and impressive work. I read and reviewed the paper before reading the comments by the other reviewers so as to be independent. I see that there are plenty of other comments already but I would have liked to also add the following (if there is an overlap with other comments, I mention it but apologies if I missed something).

1)	My main concern is as it now, the paper is hard to follow for a general audience, as the one that will attend the conference. If it is possible, I would strongly advice that a *running* example is included that can be used for describing the SHADOW framework. I think this will help the reader to follow the text. For this example, maybe start with what the goal would be and refer to it at each subsection. I am afraid that currently, without a clear goal from the get go, the included examples read a bit too abstract and hard to follow.
2)	Try to limit abbreviations throughout the text, especially in the introduction.
3)	Whenever an abbreviation is included, this should be defined on time. For example, the DAG abbreviation is defined on the workflow scheduling section but it is previously referred in the introduction.
4)	Try to define specific terms that are used before (I think this has been also mentioned by the other reviewers). For example, what is specifically a “science workflow scheduling”, or what was “oriented scheduling algorithm framework”.
5)	Also, try to be concrete as possible in all sentenced. To illustrate in the sentence “Integrating SHADOW into a larger simulation…” it should be mention “simulation” of what? Or “this is a helper-class that makes developer access to specifications” specifications of what?
6)	Table 3 misses the bottom line, right?

papers/ryan_bunney/paper.rst
stsievert: Good catch with the typo. `---` is the right command in LaTeX.

> the em dash is best limited to two appearances per sentence. Otherwise, confusion rather than clarity is likely to result.
>
> —https://www.thepunctuationguide.com/em-dash.html

deniederhut:
Hey @myxie ! Thanks to some awesome behind the scenes work by @stargaser, we now have experimental support for publishing papers with ORCIDs. If you have one and would like to add it to your paper, it's as simple as adding an `:orcid:` tag under your name in the paper header. [You can see an example of how to do this in the `00_vanderwalt` example paper](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2020/papers/00_vanderwalt/00_vanderwalt.rst). If you don't have one, no worries! You can still publish your paper as is.

myxie:
> Hello @myxie,
> 
> This is an interesting and impressive work. I read and reviewed the paper before reading the comments by the other reviewers so as to be independent. I see that there are plenty of other comments already but I would have liked to also add the following (if there is an overlap with other comments, I mention it but apologies if I missed something).

Hi @AngelosPsy - thank you very much for your reply! I apologise for not responding sooner. 

>     1. My main concern is as it now, the paper is hard to follow for a general audience, as the one that will attend the conference. If it is possible, I would strongly advice that a _running_ example is included that can be used for describing the SHADOW framework. I think this will help the reader to follow the text. For this example, maybe start with what the goal would be and refer to it at each subsection. I am afraid that currently, without a clear goal from the get go, the included examples read a bit too abstract and hard to follow.

This is a great idea, and I have added it in to make things (hopefully) flow more organically. 

>     2. Try to limit abbreviations throughout the text, especially in the introduction.

>     3. Whenever an abbreviation is included, this should be defined on time. For example, the DAG abbreviation is defined on the workflow scheduling section but it is previously referred in the introduction.

>     4. Try to define specific terms that are used before (I think this has been also mentioned by the other reviewers). For example, what is specifically a “science workflow scheduling”, or what was “oriented scheduling algorithm framework”.

Thank you for these reminders - they have been rectified in the updated text! 

>     5. Also, try to be concrete as possible in all sentenced. To illustrate in the sentence “Integrating SHADOW into a larger simulation…” it should be mention “simulation” of what? Or “this is a helper-class that makes developer access to specifications” specifications of what?

Fixed! 

>     6. Table 3 misses the bottom line, right?
Yes it does - I have fixed this too. 


I will be pushing a new update with (hopefully) all of the feedback covered in the next 24-48 hours. Thank you all for the great feedback and help.

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,782 @@
...
+sequence of tasks will have precedence constraints that limit when a task may
+start. A DAG task-graph is represented formally as a graph :math:`G = (V,E)`, where
+:math:`V` is a set of :math:`v` vertices and :math:`E` is a set of :math:`e`
+edges :cite:`kwok1999` --- an example is featured in Figure :ref:`refdag`, which will be build upon as the paper progresses. Vertices and Edges represent communication and
+computation costs respectively.  The objective of the DAG-scheduling problem
+is to minimise the execution length of the final schedule; this is referred to
+as the *makespan*.
+
+.. figure:: fig/heft_with_calc.png
+  :scale: 200 %
```
stsievert: This figure is wider than the page width.

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,782 @@
...
+~~~~~~~~~~~~~~~~~~~
+
+The workflow scheduling problem is a known NP-Hard
+problem :cite:`coffman1972,kwok1999,kousalya2017`. It
+is commonly represented in the literature as a Directed Acyclic Graph
+(DAG) :cite:`casavant1988,chaudhary1993,ullman1975,kwok1999`; a
+sequence of tasks will have precedence constraints that limit when a task may
+start. A DAG task-graph is represented formally as a graph :math:`G = (V,E)`, where
+:math:`V` is a set of :math:`v` vertices and :math:`E` is a set of :math:`e`
+edges :cite:`kwok1999` --- an example is featured in Figure :ref:`refdag`, which will be build upon as the paper progresses. Vertices and Edges represent communication and
```
stsievert: "---" doesn't render properly with this conversion. "---" is "—" in latex.

stsievert: (this is elsewhere in the paper; I'm pointing out example)

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,782 @@
...
+them. There are also a number of implementations that are present on
+open-source repositories such as GitHub, but these are not always official
+releases from papers, and it is diffic ult to keep track of multiple
+implementations to ensure quality and consistency.  SHADOW is intending to solve the meta of a 'meta-problem' with respect to task and workflow scheduling, by wrapping implementations of these algorithms in a popular language (Python) and providing a testing environment alongside them. Kwok and Ahmed
+:cite:`kwok1999` provide a comprehensive overview of the metrics and
+foundations of what is required when benchmarking DAG-scheduling algorithms,
+Maurya et al. :cite:`maurya2018` extend this work and describe key features of
+a potential framework for scheduling algorithms; SHADOW takes inspiration
+from, and extends, both approaches.
+.. _ssec:danda:
```
stsievert: This renders in the paper, which I presume isn't intended.

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,782 @@
...
+SHADOW: An algorithm reference and testing framework for scheduling data-intensive workflows
+------------------------------------------------------------------------------------------------------------
+
+.. class:: abstract
+
+  As the scale of science projects increase, so does the demand on computing
+  infrastructures. Data-intensive science pipelines that are being
+  developed for new, exa-scale projects (e.g. the Square Kilometre Array (SKA)) necessitate the
+  development of both new applications, as well as techniques for the
+  processing of entire application pipelines. Sheduling algorithms are
```
stsievert: Typo:

``` diff
- Sheduling
+ Scheduling
```

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,782 @@
...
+any way as they would a ``NetworkX`` object.
+
+SHADOW is not intended to accurately simulate the execution of a
+workflow in an real-world environment; for example, working with delays in
+processing, or node failure in a cluster. Strategies to mitigate these
+are often implemented secondary to the scheduling algorithms ---
+especially in the case of static scheduling --- and would not be a fair
+approach to benchmarking the relative performance between each
+application. Instead, it provides algorithms that may be used -
+statically or dynamically --- in a larger
```
stsievert: > may be used - statically or dynamically --- in a larger

Typo. It should be "may be used — statically or dynamically — in a larger"

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,782 @@
...
+excellent resources for determining the effectiveness of the implementations
+at the application level; however, they do not  provide a standardised
+repository of existing algorithms, or a template workflow model that can be
+used to ensure consistency across performance testing. Current implementations
+of workflow scheduling algorithms may be found in a number of different
+environments; for example, HEFT and dynamic-HEFT implementations exist in
+WorkflowSim [#]_ --- but one must traverse large repositories in order to reach
+them. There are also a number of implementations that are present on
+open-source repositories such as GitHub, but these are not always official
+releases from papers, and it is diffic ult to keep track of multiple
```
stsievert: Typo: 

``` diff
- diffic ult
+ difficult
```

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,782 @@
...
+
+------------------------------------------------------------------------------------------------------------
+SHADOW: An algorithm reference and testing framework for scheduling data-intensive workflows
+------------------------------------------------------------------------------------------------------------
+
+.. class:: abstract
+
+  As the scale of science projects increase, so does the demand on computing
+  infrastructures. Data-intensive science pipelines that are being
+  developed for new, exa-scale projects (e.g. the Square Kilometre Array (SKA)) necessitate the
```
stsievert: Why is SKA abbreviated? It's never referred to again.

stsievert (commented):
Most of my review in https://github.com/scipy-conference/scipy_proceedings/pull/553#pullrequestreview-428484678 still stands. I'm still not seeing a clear reason to use SHADOW. Sure, I know why it's better than it's competitors but I don't even know if I have a "science workflow scheduling" problem. I do see a clear definition of "science workflow scheduling," but don't see an practical example of where it might occur. If I did, I would know when to use SHADOW and when it would be relevant. I'm imagining a detailed example that specifies the data input rate, the processing considerations and any budget considerations.

> For example, science workflow scheduling is used heavily with the Square Kilometer Array (SKA). The SKA collects astronomy images from X telescopes. Each image is XMB, and images are taken every X seconds. In total, the SKA produces X GB of data *per second*.
>
> This processing poses additional constraints. The processing requires producing estimates of X, which requires X1, X2, and X3.
>
> Efficient processing of this data is required, as the computational power at SKA is large but finite. SKA utilizes a X supercomputer, which has to handle X simultaneous different analyses.

In addition, I spotted a couple typos and formatting issues. I'm sure there's some more typos; I'd run this through spell checker.

myxie:
> Most of my review in [#553 (review)](https://github.com/scipy-conference/scipy_proceedings/pull/553#pullrequestreview-428484678) still stands. I'm still not seeing a clear reason to use SHADOW. Sure, I know why it's better than it's competitors but I don't even know if I have a "science workflow scheduling" problem. I do see a clear definition of "science workflow scheduling," but don't see an practical example of where it might occur. If I did, I would know when to use SHADOW and when it would be relevant. 

Thank you again @stsievert for reading through the updates. I have (hopefully) added some additional paragraphs to demonstrate why one might use SHADOW.

It is clear that my reference to the SKA is a confusing one, as SHADOW is tangential to the actual processing of SKA data (given that the SKA does not currently exist yet). Rather (taken from my comment to @msarahan above) "SHADOW is intending to solve more of a 'meta-problem' with respect to task and workflow scheduling; namely, the lack of algorithm implementations, and a framework to test the performance of new algorithms against existing ones. You are right that it was born out of scheduling problems we are investigating for the SKA, where we are looking at scheduling approaches to deal with the data. However, in order to demonstrate that the existing approaches don't work as well as an SKA-specific approach, we need existing algorithms implementations to benchmark against."

I have clarified the language in both the abstract and the introduction to make this clearer, and removed all mention of the SKA - which, as you point out, is mentioned only once. I hope that this makes the intended purpose of the framework clearer.  

> **Abstract** [our] intention is for the framework to become a reference implementation of scheduling algorithms, and provide algorithm designers an opportunity to develop and test their own algorithms with the framework. 

>**Introduction** The intended audience of SHADOW is those developing and testing novel workflow scheduling algorithms, as well as those interested in exploring existing approaches within an accessible framework. ...The emphasis in
SHADOW is on reproducibility and accuracy in algorithm performance analysis...

Additionally, I have placed the following (small) discussion of how science workflows may be used in  'production':

>  An example of science workflow usage is Montage[#], which takes sky images and re-projects, background corrects and add astronomical images into custom mosaics of the sky :cite:`bharathi2008,juve2013`. This workflow is deployed on a workflow management system (for example, Pegasus :cite:`deelman2015`), which coordinates the deployment and execution of the workflow. It is this workflow management system that passes the workflow to a workflow scheduling algorithm, which will pre-allocate the individual application tasks to nodes on the execution environment (e.g. a local grid or a cloud environment) in preparation for the workflow's execution.  

> I'm imagining a detailed example that specifies the data input rate, the processing considerations and any budget considerations.

I would say that a discussion of a particular workflow problem (e.g. Astronomy, Genetics, Seismology etc.) data, resource, or budget requirements are relevant to a particular algorithm or the problem environment - for example, algorithms for scheduling on the grid often are concerned with fairness or workflow runtime, whereas the same problem may require a different algorithm for the Cloud due to a costs and provisioning differences. This part is not relevant to SHADOW per se, as SHADOW is itself not an algorithm or technique - but the development of that algorithm would be possible using the SHADOW framework (and encouraged!). 

> In addition, I spotted a couple typos and formatting issues. I'm sure there's some more typos; I'd run this through spell checker.

I have addressed these in the most recent update (and run them through a spell checker!).

deniederhut:

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,799 @@
...
+provide algorithm designers an opportunity to develop and test their own
+algorithms with  the framework. SHADOW code is hosted on GitHub at
+https://github.com/myxie/shadow; documentation for the project is available in
+the repository, as well as at https://shadowscheduling.readthedocs.org.
+
+
+Introduction
+------------
+
+To obtain useful results from the raw data produced by science experiments, a series of scripts or applications is often required to produce tangible results. These application pipelines are referred to as Science Workflows :cite:`alkhanak2016`, which are typically a Directed-Acyclic Graph (DAG) representation of the dependency relationships between application tasks in a pipeline. The processing of Science Workflows is therefore an example of the DAG-Task scheduling problem, which is in the class of NP-Hard problems :cite:`kwok1999`. Science workflow scheduling
```
stsievert: I feel these first 4 sentences move very quickly. There are 3 claims in these sentences, and I don't have enough context to see why each claim is true. I'd expect that's the true for conference attendees too.

I think moving the example to this paragraph would help provide more context for the reader. Here's one suggestion:

> To obtain useful results from the raw data produced by science experiments, a series of scripts or applications is often required to produce tangible results. These application pipelines are referred to as Science Workflows :cite:`alkhanak2016`, which are typically a Directed-Acyclic Graph (DAG) representation of the dependency relationships between application tasks in a pipeline.  An example of science workflow usage is Montage[#] ... (e.g. a local grid or a cloud environment) in preparation for the workflow's execution.
>
> The processing of Science Workflows is therefore an example of the DAG-Task scheduling problem, which is in the class of NP-Hard problems :cite:`kwok1999`. Science workflow scheduling ...

stsievert: Suggestion: it would be nice to have some numbers in the example too. Roughly, how many images are there? Roughly, how large is each image? Roughly, how many images are taken per day?

stsievert (commented):
> Additionally, I have placed the following (small) discussion of how science workflows may be used in 'production':

Thanks! That's the context I was looking for. Now I have a use case in mind while I'm reading the paper. I think providing this context earlier would help make the paper more clear (details below).

> I would say that a discussion of a particular workflow problem (e.g. Astronomy, Genetics, Seismology etc.) data, resource, or budget requirements are relevant to a particular algorithm or the problem environment.

I agree. Now there are citations to provide more detail on science workflow scheduling.

AngelosPsy:
Thank you @myxie for addressing most of my comments. I am afraid, though, that there are still some issues with the manuscript (ms). I am actually repeating some of my previous comments. I still think that the manuscript could be a bit more to the point and some parts could actually be deleted in order to ensure that the examples are explained better. Here is a summary of my comments:

1)	There are a few parts still that are not defined concretely enough. To illustrate, in the abstract it is mentioned “and optimised for different science use cases, environments, and objectives”. How are science use cases defined? Science is really broad. For example, I work in social sciences but I doubt that what the cases described also later in the manuscript apply necessarily to all brances of social sciences. Similar arguments go later in the introduction where there is a reference to “NP-Hard Problems”, which is a common term in complexity theory but not sure that all readers in Scipy will understand immediately what these problems refer to. So, please, again scan the ms and be as concrete as possible.
2)	On page 2, I was a bit confused with the ‘existing approaches’ section and not sure whether it is needed. From what I understand, there is a bit of an overlap between this section and parts of the introduction (see second column on p. 1). I was wondering if this section is needed for understanding the paper at all.
3)	“SHADOW is intending to solve the meta of a ’meta-problem’ with respect to task and workflow scheduling,…” Consider rephrasing. At the moment is reads a bit too poetic.
4)	“Additionally, by using the NetworkX.DiGraph as the storage object for the workflow structure, users may extend the SHADOW Workflow object in any way as they would a NetworkX object.” I wonder if this paragraph adds much.
5)	Is the “Existing approaches” on p. 5 needed? Will deleting this section influence negatively the comprehension of the ms?
6)	I appreciate that the authors use more examples in this version of the paper. However, I was wondering, why there is not a running example throughout the paper? For example, I hoped that the Montage example could be used throughout the paper because it is concrete enough.

These are my comments but I am open if you want to oppose them.

papers/ryan_bunney/paper.rst
myxie: This makes a lot of sense, thank you. I have moved it to the beginning to add some immediate context. 

I have added something of context from a publication that has run some performance characterisation on  number of different workflows - it is less specific on the images/size of the image, but hopefully it gives an example of the scope of the workflow itself! 

>  A Montage pipeline may consist of more than 10,000 jobs, perform more than 200GB of I/O (read and write), and take 5 hours to run :cite:`juve2013`.

stsievert: :+1: That gives much more context, especially with the specific numbers!

Style nit: the whole first page is one run-on paragraph. I might break into several paragraphs to separate the different topics into different paragraphs (problem definition, problem solutions and solution availability).

papers/ryan_bunney/paper.rst
```diff
@@ -0,0 +1,797 @@
...
+in code that uses it, such as in simulation frameworks like
+WorkflowSim
+:cite:`topcuoglu2002,chen2012a`; others are not implemented in any public way at all :cite:`yu2006,abrishami2010`. These
+are also typically used as benchmarking or stepping stones for new algorithms;
+for example, the Heterogeneous Earliest Finish Time (HEFT) heuristic continues to be used as the foundation for scheduling
+heuristics :cite:`durillo2012,caniou2018`, meta-heuristics, and even
+mathematical optimisation procedures :cite:`bridi2016`, despite being 20 years
+old. The lack of a consistent testing environment and implementation of
+algorithms makes it hard to reproduce and verify the results of published
+material, especially when a common workflow model cannot be verified
```
stsievert: Typo:

``` diff
- cannot be verified
+ cannot be verified.
```

myxie:
> Thank you @myxie for addressing most of my comments. I am afraid, though, that there are still some issues with the manuscript (ms). I am actually repeating some of my previous comments. I still think that the manuscript could be a bit more to the point and some parts could actually be deleted in order to ensure that the examples are explained better. Here is a summary of my comments:

Hi @AngelosPsy - on reading your new comments, I agree I did not adequately address them, and I'm sorry you have had to repeat yourself. Hopefully this new round of updates has clarified things a little more. 
> 
>     1. There are a few parts still that are not defined concretely enough. To illustrate, in the abstract it is mentioned “and optimised for different science use cases, environments, and objectives”. How are science use cases defined? Science is really broad. For example, I work in social sciences but I doubt that what the cases described also later in the manuscript apply necessarily to all brances of social sciences. Similar arguments go later in the introduction where there is a reference to “NP-Hard Problems”, which is a common term in complexity theory but not sure that all readers in Scipy will understand immediately what these problems refer to. So, please, again scan the ms and be as concrete as possible.

For this scenario, I have tried to reduce the us of jargon (i.e., removed references to NP completeness etc.). Additionally, I have cleaned up the opening to ensure there is more concreteness by removing the vague notion of 'science and environments (etc.). 

>     2. On page 2, I was a bit confused with the ‘existing approaches’ section and not sure whether it is needed. From what I understand, there is a bit of an overlap between this section and parts of the introduction (see second column on p. 1). I was wondering if this section is needed for understanding the paper at all.

I agree the positioning here was not ideal. The intention was to let the reader know that SHADOW is not alone in providing a software framework to do work on workflow scheduling algorithms, and I wanted to give credit to the previous work that has been done in this area. I think it also adds another point of reference as to the differences between SHADOW and what already exists, and therefore why SHADOW is a useful tool. 

I have moved this to the end (prior to the conclusion) and rephrased the section as 'Alternatives to SHADOW'. 

>     3. “SHADOW is intending to solve the meta of a ’meta-problem’ with respect to task and workflow scheduling,…” Consider rephrasing. At the moment is reads a bit too poetic.

Yes it was - and not concrete! This has been changed. 
> 
>     4. “Additionally, by using the NetworkX.DiGraph as the storage object for the workflow structure, users may extend the SHADOW Workflow object in any way as they would a NetworkX object.” I wonder if this paragraph adds much.

This was in the wrong position and so I have moved it. I wanted to communicate that because we are using NetworkX, which is a popular Python framework for working with Graph data structures, you can interact with our code like a NetworkX object (i.e. it is not completely *new*). I have changed the context and made it clearer that it is referring to users of NetworkX. 

>     5. Is the “Existing approaches” on p. 5 needed? Will deleting this section influence negatively the comprehension of the ms?

This was intended to be a validation for why SHADOWGen has a particular cost generation, and why it standardises the approach to cost generation for workflow generation and testing. I have pulled this section into the 'Cost Generation in SHADOWGen so the context is more obvious, and the flow is better'. I am of course happy to remove the whole section if you feel it is not of interest to the SciPy community. 

>     6. I appreciate that the authors use more examples in this version of the paper. However, I was wondering, why there is not a running example throughout the paper? For example, I hoped that the Montage example could be used throughout the paper because it is concrete enough.

I think I misunderstood your initial comment regarding the running example. What I have done now is use the Workflow in Figure 1/2 and continued to refer to that example as I go through the different modules: 

>  Using the example workflow shown in Figures :ref:`refdag` and :ref:`heftcalc`, we can demonstrate how to initialise a ``Workflow`` in SHADOW, and what options exist for extending or adapting the object. 

and

> For example, looking at Figure :ref:`heftcalc` we see the dependency between tasks 0 and 1, and the weight $18$ on the edge. This is reflected in the above component of the JSON file.

>  In our example, we have three machines on which we are attempting to schedule the workflow from Figure :ref:`heftcalc`. The different performance of each machine is described in Table :ref:`hefttable`, with the JSON equivalent below:

and so on. If this is not what you are looking for, then I'm not quite sure what you mean. I do agree that a running example is helpful, and have found that the paper reads more consistently now even with the smaller running example. I hope that this is enough for you! 

> These are my comments but I am open if you want to oppose them.

These have been very helpful comments, and I really appreciate your time in continuing to read the paper and engaging with my updates.

papers/ryan_bunney/paper.rst
myxie: Done!

AngelosPsy:
Hello @myxie,

Thank you for your quick replies. Apologies also for not being concrete enough about what I meant with 'running example'. 
I had something in mind in lines of: We had the X workflow, that we wanted to achieve Y, and for that we had tasks A, B, and C, with of course every capital letter referring to concrete paradigms of a process, goal, and tasks that can be easily understood from a broad audience. This is what I had in mind as I think that talking just for 'tasks 0 and 1' maybe is not concrete enough. But if you think this is not possible or it won't help the paper anymore it is OK.

myxie:
> Hello @myxie,
> 
> Thank you for your quick replies. Apologies also for not being concrete enough about what I meant with 'running example'.
> I had something in mind in lines of: We had the X workflow, that we wanted to achieve Y, and for that we had tasks A, B, and C, with of course every capital letter referring to concrete paradigms of a process, goal, and tasks that can be easily understood from a broad audience. This is what I had in mind as I think that talking just for 'tasks 0 and 1' maybe is not concrete enough. But if you think this is not possible or it won't help the paper anymore it is OK.

Hi @AngelosPsy, 

Your example makes sense if we were focusing on a science application  (i.e. we are working on a particular Workflow and concerned about its optimisation). Workflow scheduling is *usually* Workflow agnostic, and so takes a more abstract approach. This is why I have used a generic example, because the use case for SHADOW is going to involve developing an algorithm on which multiple different workflows will be tested and evaluated. I do see where you are coming from, and perhaps it may be a useful case study for a future publication, but based on what I am trying to achieve in this paper I think it is not necessary. 

Thank you again for your detailed comments and commitment to this review - it has been a great opportunity to think critically about the work!

deniederhut:
Hi @AngelosPsy and @stsievert ! Are you happy with the updates to this paper?

AngelosPsy:
Hello @deniederhut,

Yes, I am fine with the last version of the paper and I do not have further comments.

stsievert:
I am happy with the updates, and don't have any more comments. I am certainly glad the paper provides context for the naive reader.

PR #555, opened 2020-05-29
Paper: Having your cake and eating it: Exploiting Python for programmer productivity and performance on micro-core architectures using ePython

mesham:
Paper: Having your cake and eating it: Exploiting Python for programmer productivity and performance on micro-core architectures using ePython


aweinstein:
Hi: I was assigned to review this paper. I gave it a quick read and overall it looks very good. I will have a detailed revision before the deadline.

mesham:
Thanks aweinstein, please let us know any comments, concerns etc and we are happy to address them and discuss any points if need be.

aweinstein:
The paper presents ePython, a Python implementation designed to run on micro-core architectures. The paper is clear and well written, and it gives a description of ePython that allows the reader to understand its basic characteristics and performance. 

Some comments follow:

My main observation is that, based on the syntax, I see that ePython is a Python 2 implementation. Given that Python 2 is no longer supported, the authors should: (a) Make explicit which version of Python they are implementing, and (b)  explain why they are not implementing Python 3 (I understand that there might be good reasons for not implementing Python 3, but please state them in the paper).

Regarding figure 2 and its description, I have two suggestions:
- Make clearer which part of the diagram lives in the host computer, and which ones in the micro-core. For instance, you can have one box for each component.
- Explain how the physical layer of the communication between the host and the micro-core works.  Later in the paper, there are statements about the memory-mapped interface and RS232, but I think this should be discussed when presenting the architecture.

In the introduction, you mention that "MircoPython is hundreds of KBs". Can you bound this number or give some examples? That would be interesting to get an idea of the difference for ePython in this regard (if finding this number is to much work, please ignore this comment, it is not a big deal).

Another minor point is that code snippets use an `import *` style. Since this is not recommended, I suggest a footnote (or similar) to justify its use, or even better, use an explicit import.

I have some other suggestions related to typos, that I will list as individual comments, to make them easier to address. Please notice that I am not a native English speaker, so please ignore any comment related to the English writing that you don't find appropriate.

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
...d PicoRV32 are, given current activity in the community, the more interesting targets. The MicroBlaze is developed by Xilinx, a large multi-national corporation who also develop the underlying FPGAs and there is significant commitment by Xilinx to the technology. On the other-hand, the PicoRV32 is an implementation of the RISC-V Instruction Set Architecture (ISA). RISC-V is an open standard ISA and, first introduced in 2010, one of the major reasons for its popularity has been the fact that it is provided under open source licenses that do not require fees. This means that anyone is free to download the specification and develop their own implementation of the ISA, which indeed the PicoRV32 project have done. Furthermore, because all these CPUs share the same ISA, then the software eco-system such as compilers, debuggers, profilers etc is the same. Therefore the developers of a new RISC-V based CPU can often take the existing RISC-V software eco-system and run these with little or no modifications on their chip. With a large community, who are mixture of commercial and academic contributors, RISC-V is currently a very topical and active area of research and commercial exploitation.
```
aweinstein: Check writing around "profilers etc is the same".

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
+
+.. code-block:: python
+  :linenos:
+
+  from parallel import *
+  from random import randint
+
+  a=reduce(randint(0,100), "max")
+  print "The highest random number is "+str(a)
```
aweinstein: Add space around the `+` symbol.

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
+
+This approach was initially developed with the objective of running rather simple examples on the micro-cores directly and exposing programmers to the fundamental ideas behind parallelism in a convenient programming language. However, going beyond the work of [epython]_, we realised that there was potential for ePython to support real-world applications on micro-cores, but to do so a more powerful approach to programmer interaction was required. This is because not all parts of an application are necessarily suited for offloading to micro-cores, so an approach where specific functions can be selected for offload conveniently was required to extend the technology, which is the focus of this paper in the next section.
```
aweinstein: Replace "which is the focus of this paper in the next section." by "which is the focus of next section."

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
+
+.. figure:: epython_memory_hierarchy.png
+   :align: center
+   :scale: 60%   
+
+   Illustration of memory hierarchy for the Epiphany
```
aweinstein: Add a period at the end of the caption.

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
...BC). The Parallella combines a host dual core ARM A9 CPU, with 1 GB of DRAM and the 16 core Epiphany-III. The theoretical off-chip bandwidth of the Epiphany is 600 MB/s, however in practice the maximum obtainable is around 150 MB/s. For our two soft-cores we use the same base-board, a Pynq-II SBC, and a Xilinx Zynx-7020 reconfigurable FPGA. The Zynq-7020 chip is especially interesting, as in a single physical package not only is there the reconfigurable fabric which we can use to represent our micro-cores of interest, but furthermore a dual core ARM A9 CPU which runs Linux. Therefore in a single chip we have the combination of a dual-core host CPU on the one-hand, and logic configured as multiple micro-core CPUs on the other. The board also contains 512 MB RAM with an off-chip bandwidth of 131.25 MB/s. This specific FPGA contains 53,200 programmable Look-Up Tables (LUTs), and around 627 KBs of block RAM (BRAM). In-fact it is this BRAM, effectively the amount of local memory per core, which is the limiting factor here and we can fit a maximum of eight 64 KB MicroBlaze or PicoRV32 CPUs and supporting infrastructure onto the Zynq, which is the configuration used throughout this paper.
```
aweinstein: Remove dash in "In-fact".

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
+
+Memory model
+------------
+Whilst at first glace the approach we present seems rather straight-forward, when one starts to use these abstractions for programming non-trivial kernels they soon realise that there are specific questions around data consistency, especially when data is held on the host. This is not helped by the fact that Python does not specify a standard memory model, with individual implementations being free to adopt whichever memory model they wish. In contrast to many other Python implementations, ePython adopts a rather weak memory model, which the programmer should be aware of.
```
aweinstein: "Whilst at first glace the approach we present seems rather straight-forward, when one starts to use these abstractions for programming non-trivial kernels they soon realise"

Check this sentence: 
-"glace" -> "glance"
- Should there be a dash in "straight-forward"?
- "They soon": It is not clear to me to whom you refer by "they"

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
...o the micro-core and then caching it. In terms of caching policy, we adopt a write-through approach, where the locally held copy (effectively in our cache) will then be used for all the reads, and writes are performed on both the local copy of data and also written back to the variable's location elsewhere in the hierarchy. Locally held cache copies of data are evicted automatically by the ePython runtime as required, such that the memory space can then be reused for subsequent data. Access to any data, whether it be a scalar or array element, held in memory locations outside the core will always first check whether there is a copy held locally, and if not perform the explicit data movement required. At the time of writing, by default the explicit data movement into our cache will wait until the data is required and then move it, but the disadvantage of this approach is that it stalls execution until the memory operation complexities. As such the programmer can, via decorating their code, instruct the data movement to be done ahead of time via non-blocking pre-fetching, thus not stalling the micro-cores on data access, and in the future this will likely become the default approach.
```
aweinstein: "but the disadvantage of this approach is that it stalls execution until the memory operation complexities." 

This sentence looks awkward to me.

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
...res (100Mhz). However, clearly from the results a six times difference in clock frequency is not the only reason for the performance gap, and other architectural differences play a role too. If we normalise for clock frequency, floating point operations on the PicoRV32 are still approximately 9 times slower than on the Epiphany, and this is because the Epiphany contains a hardware FPU which is superscalar, providing the capability of processing up to two floating point operations concurrently. By contrast, the PicoRV32 does not contain an FPU and as such all floating point arithmetic must be performed in software. Again normalising for clock frequency, array accesses are around 9.5 times slower on the PicoRV32 than on the Epiphany, and this is because on the Epiphany and MicroBlaze the cost of a memory load in cycle per instruction (CPI) is 1 cycle, whereas on the PicoRV32 it is 5 cycles. The Epiphany provides a variable length pipeline of up to eight stages and the MicroBlaze a five stage pipeline, by contrast the PicoRV32 is not pipelined and this results in an average CPI of 4 instructions, with the next instruction not being able to begin until the proceeding one has completed.
```
aweinstein: "This is potentially not surprising given the fact that because it is a physical chip" 

The "because" looks awkward to me.

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
+
+Table 2 illustrates the runtime in seconds across different technologies when natively compiled. It can be seen that this is very significantly faster, over 500 times, than using the ePython interpreter on the Epiphany. For comparison we have developed a C version of the benchmark specifically for the Epiphany and this represents the alternative of writing a bespoke implementation for the architecture, with all the complexities associated with this, and compiling and running directly. We also ran a version of this benchmark on an AMD64 CPU (as both the ePython interpreter and native code generation support x86), which are ubiquitous in HPC and consumer grade computing.
+
+.. table:: Runtime of natively compiled Python code via ePython, against bespoke C code, on both the Epiphany and AMD64 x86 CPU
```
aweinstein: Add a period at the end of the caption.

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
+
+Table 2 illustrates the runtime in seconds across different technologies when natively compiled. It can be seen that this is very significantly faster, over 500 times, than using the ePython interpreter on the Epiphany. For comparison we have developed a C version of the benchmark specifically for the Epiphany and this represents the alternative of writing a bespoke implementation for the architecture, with all the complexities associated with this, and compiling and running directly. We also ran a version of this benchmark on an AMD64 CPU (as both the ePython interpreter and native code generation support x86), which are ubiquitous in HPC and consumer grade computing.
```
aweinstein: "very significantly faster" looks awkward to me.

aweinstein: "with all the complexities associated with this, and compiling and running directly."

This looks awkward to me.

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,273 @@
...
+
+The reader can clearly see that the performance obtained by ePython is very architecture specific, which is not surprising given the diversity of the different types of micro-cores and associated level of complexity. Whilst we expected there to be a rather significant performance overhead associated with the ePython interpreter, the magnitude of this when compared to native code compilation surprised us. This is in part due to the fact that performance was really not an objective when we first developed the interpreter and, without a doubt, certain decisions were made which retrospectively limit performance. Whilst it would be possible to address some of these, instead one can see that the performance overhead of ePython can in large be ameliorated by using native code compilation to run the Python code directly on the micro-cores, without the need for an interpreter to be present.
```
aweinstein: "there to be a rather"

Sound awkward to me.

papers/maurice_jamieson/maurice_jamieson.rst
aweinstein: And also around the `=`.

deniederhut:
Hey @mesham ! Thanks to some awesome behind the scenes work by @stargaser, we now have experimental support for publishing papers with ORCIDs. If you have one and would like to add it to your paper, it's as simple as adding an `:orcid:` tag under your name in the paper header. [You can see an example of how to do this in the `00_vanderwalt` example paper](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2020/papers/00_vanderwalt/00_vanderwalt.rst). If you don't have one, no worries! You can still publish your paper as is.

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Thanks, we have reworded that and the sentence after it to make it clearer

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Thanks, we have done this and also ensured the subsequent code examples are consistent with this too

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Thanks, this has been done

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Thanks, done this here and for other captions which also missed the period from the end

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Done, there were a couple of other instances of this too which we have also fixed

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Thanks, we have changed glace to glance (very good spot!) I've just googled it, and I don't think straightforward needs a dash in the middle, so I have omitted it if that's OK (I have a tendency to pop in dashes, so actively try and resist this urge!) We have changed "they soon realise" to "it soon becomes apparent" - hopefully this makes it clearer we mean the programmer when they use ePython

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Ah, good spot - instead of "complexities" we meant "completes" at the end- I have replaced this and hopefully it makes the sentence much clearer!

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Thanks, after re-reading this I agree with you. We have rephrased the sentence as "This is potentially not surprising given the fact that it is a physical chip, and as such can run....." - I think this reads better

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Thanks, that's done now

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Yes I agree, and not very scientific either :) I have simply removed the word "very", so instead it is "significantly faster" - I think that makes is scan better and the reader can then decide the magnitude of the significance for themselves.

mesham: I agree, and from re-reading it it isn't particularly clear either. We have rephrased this as "Developing such code in C is a significant undertaking, as the programmer must deal with numerous architecture specific complexities and low level concerns. Whilst it is this programming complexity that we believe Python has significant potential to overcome for micro-cores, we nevertheless felt it was interesting to include a C version as a comparison in a performance study such as this."

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Thanks, yes I agree. We have rephrased that sentence to be "Whilst we expected a performance overhead associated with the ePython interpreter, the magnitude of this when compared to native code compilation surprised us." This makes it simpler I think

mesham:
Thanks very much for your review @aweinstein , I appreciate the time you have taken to look at this and give your comments. We have updated the paper and I will describe here the changes made. Firstly thanks for all the comments highlighting the typos you found, these were really helpful and we have actioned all of them (and commented on each individual one the changes made there.)

In terms of the specific points you mentioned in the review:

- Python 2 implementation and not Python 3: You are absolutely correct here, and this is partly due the fact that we started this work a few years ago, but also due to Python 3 requiring more support (e.g. more memory footprint) on the interpreter. I have made explicit on page 3 that we currently target 2.7, and mention in the future work discussion of the conclusions about Python 3. It is our plan to move to supporting Python 3, but we think that the native code compilation is key here and we are waiting for this to reach maturity. It is because this avoids the memory footprint overhead, and as such it will unlock the potential of doing this in the next version.

- Your suggestions around figure 2. We have added a "host" and "micro-core" braces  to the top of this figure, which hopefully makes it much more obvious about what lives where. For your second comment we have added a discussion at the top of this section about how the different parts connect together, and added some more detail around the connection between the host and the micro-cores (hopefully this is enough, it can get quite complex in reality to ensure these links provide the data consistently.)

- MicroPython size - this is a very good point and I thought best to reference the size of MicroPython which is stated on their website, so we updated that section with this. It isn't quite as simple as this, as often microcontrollers have ROM associated too which can be flashed to hold the MicroPython code itself. This isn't something typically available with micro-cores, so we don't have this luxury, but we have added in a couple of sentences around this to set it in context and explain these fundamental differences a little more.

- Move to explicit import - ah, good spot - ePython supports both so I have changed these as you suggest in the code snippets.

Hopefully this meets the points you raise, thank you once again for taking the time to read through our paper and for your comments. If there is anything else, or anything you think we should address in more detail then let us know.

@deniederhut - these revisions have pushed us over to the next page (e.g. onto a 9th page.) Hopefully that's OK, we could cut some stuff out to reduce it, but it seems a shame to do so.

deniederhut:
Hm. We don't count references in the page limit, so you aren't that far onto the 9th page. We may be able keep this paper under the limit with some editing to make it more concise. E.g. the following passage from the introduction:

> In many ways we are seeing the embedded and HPC  worlds  converging,  the  embedded  world  which  has  always focussed on energy efficiency now interested in parallelism, and the HPC community having to consider energy efficiency in order to facilitate realistic future exa-scale machines. These micro-core architectures,  providing  significant  parallelism  and  performance for  low  power  are  therefore  of  great  interest  to  both  communities.

Could be rewritten as something like:

> The low cost of micro-core architectures enable massive parallelism -- a new paradigm for embedded systems. This has attracted the HPC community, which does not have a history of working with low power or low memory devices.

mesham:
Thanks @deniederhut - I will take a look as you suggest and try to snip these bits down

aweinstein:
@mesham Thank you for correcting the paper.

@deniederhut I think this paper is good to go.

deniederhut:
Hi @mesham ! Have you had a chance to work on reducing the paper length?

mesham:
@deniederhut thanks for the prod - will do so over the weekend and aim to have it done by Monday if that's OK?

deniederhut:
Yup!

mesham:
@deniederhut Thanks - we have committed a revised version which makes some of the sentences and paragraphs a bit more concise as you suggested. As such, we have managed to get the paper back within the page limit :)

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,277 @@
...
+.. [picorv32] C. Wolf. *PicoRV32 - A Size-Optimized RISC-V CPU*,
+          On Github, https://github.com/cliffordwolf/picorv32/
+
+.. [pezy-sc] T. Ishii. *Introduction to PEZY-SC*
+          http://accc.riken.jp/wp-content/uploads/2015/09/ishii.pdf
+
+.. [kalray] B.D de Dinechin. *Kalray MPPA: Massively parallel processor array: Revisiting DSP acceleration with the Kalray MPPA Manycore processor*
+          Hot Chips 27 Symposium (HCS), 2015 IEEE, pages 1--27
+
+.. [ajayi2017celerity] T. Ajayi. *Celerity: An Open-Source RISC-V Tiered Accelerator Fabric*
```
deniederhut: This reference looks incomplete

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,277 @@
...
+
+.. [epiphany] A. Olofsson. *Kickstarting high-performance energy-efficient manycore architectures with epiphan*
```
deniederhut: "epiphany"

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,277 @@
...
+          Signals, Systems and Computers, 2014 48th Asilomar Conference on.
+
+.. [parallella] Adapteva. *Parallella-1.x Reference Manual*
```
deniederhut, line 253: Does this have a year published? Also the link is returning 404 for me

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,277 @@
...
+          http://www.parallella.org/docs/parallella\_manual.pdf
+
+.. [microblaze] Xilinx. *MicroBlaze Processor Reference Guide*
```
deniederhut, line 256: Does this have a year for when it was published?

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,277 @@
...
+          https://www.xilinx.com/support/documentation/sw_manuals/xilinx2018_2/ug984-vivado-microblaze-ref.pdf
+
+.. [epiphany-specifications] Adapteva. *Epiphany Architecture Reference*
+          http://www.adapteva.com/docs/epiphany_arch_ref.pdf
+
+.. [opencl] J.E. Stone. *OpenCL: A parallel programming standard for heterogeneous computing systems*
```
deniederhut: This reference is missing a lot of information, like the full list of authors -> https://ieeexplore.ieee.org/document/5457293

deniederhut:
@mesham a lot of these references look incomplete. Were you planning on correcting them?

mesham:
@deniederhut  ah thanks for spotting this, we have fixed them by adding extra information to them all - hopefully that's sufficient, if not let me know

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Yup, added extra information here

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Good spot :) fixed

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Yes, and I have also fixed the link (I had escaped the underscore, but realised this isn't needed)

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Yes, added

papers/maurice_jamieson/maurice_jamieson.rst
mesham: Ah yes, sorry - added

papers/maurice_jamieson/maurice_jamieson.rst
```diff
@@ -0,0 +1,278 @@
...
...ode directly on the micro-cores, without the need for an interpreter to be present. Therefore pur present focus is in maturing the native code generation as we think this has demonstrated some worthwhile early results. In addition to exploring opportunities for further performance improvements, currently the architecture specific runtime library is not included in this dynamic loading, so the minimum code size is around 15KB (runtime and bootloader together). If we were to extend the dynamic loading approach to the runtime too, then the minimum size will be around 1.5KB plus the size of the largest function. This will open up the possibility of running over a number of additional micro-core architectures which contain tiny amounts of memory per core (only around 2KB or 3KB). Furthermore, our dynamic loading approach to native code compilation can be extended to fetch parts of third-party libraries, such as Numpy or Sklearn. This will require some thought, as we will need to split apart the ELF into its constituent components, but it would be of significant benefit to the micro-core software ecosystem if such a rich set of existing numerical frameworks could be supported by ePython.
```
deniederhut: I think we want 'Therefore pur present' to be 'Therefore our present'

mesham: Ah yes, thanks @deniederhut  - fixed and latest version committed :)

PR #566, opened 2020-05-30
Paper: Learning from evolving data streams

jacobmontiel:
This is a pull request for SciPy 2020. Submission 115 by Jacob Montiel.


mmccarty:
Hello @jacobmontiel, I will be reviewing your paper. I'm going to set aside some time to get you some feedback early this week. Until then, I'm wondering if you have run across any of the following:
  - https://github.com/python-streamz/streamz
  - https://ml.dask.org/incremental.html

Their authors may also be good reviewers for this paper or may know of others: @cj-wright @TomAugspurger @stsievert @jrbourbeau

jacobmontiel:
Sounds good to me @mmccarty. Yes, I am familiar with those projects and it would be great to have some comments. Martin Durant (@martindurant) is familiar with this project as well.

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+.. class:: abstract
+
+   Ubiquitous data poses challenges on current machine learning systems to store, handle and analyze data at scale.
+   Traditionally, this task is tackled by dividing the data into (large) batches. Models are trained on a data batch
+   and then used to obtain predictions.  As new data becomes available, new models are created which may contain
+   previous data or not. This training-testing cycle is repeated continuously. Stream learning is an active field where
+   the goal is to learn from infinite data streams. This gives rise to additional challenges to those found in the
+   traditional batch setting: First, data is not stored (it is infinite), thus models are exposed only once to single
+   samples of the data, and once processed those samples are not seen again. Models shall be ready to provide
+   predictions at any time. Resources such are memory and time are limited, consequently, they shall be carefully
```
mmccarty: ~~Resources such are~~ Compute resources such as

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+   managed. The data can change (evolve) over time and models shall be able to adapt accordingly.
```
mmccarty: ~~change (evolve)~~ drift
"drift" is a common term we use in my line of work or just say "evolve"

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+   This is a key difference with respect to batch learning, where data is assumed static and models will fail in the
+   presence of change. Model degradation is a side-effect of batch learning in many real-world applications requiring
+   additional efforts to address it. This papers provides a brief overview of the core concepts of machine learning for
```
mmccarty, line 25: > Model degradation is a side-effect of batch learning in many real-world applications requiring
   additional efforts to address it.

I suggest you also note that there is effort to monitor and detect model degradation.

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+------------
+
+The minimum pipeline in machine learning is composed of: (1) data collection and processing, (2) model training, and
+(3) model deployment. Traditionally, data is collected and processed in batches. Although this approach is
+state-of-the-art in multiple applications, it is not suitable in the context of evolving data streams as it implies
+important compromises. In batch learning, data is assumed to be accessible (usually stored on a physical device), the
+limitations on this approach have been pushed as storage and processing power became cheaper. Similarly, advances in
+data management and storage have enabled the handling of large volumes of data. However, there exist instances where
+storing large volumes of data is not practical. Additionally, keeping an accurate model in dynamic systems implies
+running the batch learning pipeline multiple times, which represents a challenge in fast-changing environments.
```
mmccarty: The assumptions made in batch pipelines are 

1. Data is accessible
1. Data is large
1. The pipeline will be run multiple times (I assume for retraining)

Perhaps there is a way of saying that more succinctly?

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+
+.. class:: keywords
+
+   machine learning, data streams, concept drift, scikit, open-source
+
+Introduction
+------------
+
+The minimum pipeline in machine learning is composed of: (1) data collection and processing, (2) model training, and
+(3) model deployment. Traditionally, data is collected and processed in batches. Although this approach is
```
mmccarty: Perhaps add a 4th step, "model validation"? However, I'm not sure how that impacts the narrative throughout the paper.

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+   Batch learning systems are characterized by the investment in resources like memory and training time as the volume
+   of data increases. Once a *reasonable investment threshold* is reached, data becomes unusable turning into a missed
+   opportunity. On the other hand, efficient management of resources makes stream learning an interesting alternative
+   for big data applications. :label:`fig:investment`
+
+As shown in the previous examples, dynamic environments pose an additional set of challenges to batch learning systems.
+Model degradation is a predominant problem in multiple real-world applications. As enough data has been generated and
+collected, proactive users might decide to train their models to make sure that they agree with the current data. This
+is complicated for two reasons: First, batch models (in general) are not able to use new data into account, so the
+machine learning pipeline must be run multiples times as data is collected over time. Second, the decision for such an
```
mmccarty, line 94: Because the model needs to be retrained?

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+action is not trivial and involves multiple aspects. For example, should a new model be trained only on new data? This
+depends on the amount of variation in the data. Small variations might not be enough to justify retraining and
+re-deploying a model. This is why a reactive approach is predominantly employed in the industry. Model degradation is
+monitored and corrective measures are enforced if a user-defined threshold is exceeded (accuracy, type I, and type II
+errors, etc.). Fig. :ref:`fig: investment` depicts another important aspect to consider, the tradeoff between the
+investment in resources such as memory and time (and associated cost) and the pay-off in predictive performance. In
+stream learning, resource-wise efficiency is fundamental, predictive models not only must be accurate but also must be
+able to handle theoretically infinite data streams. Models must fit in memory no matter the amount of data seen
+(constant memory). On the other hand, training time is expected to grow sub-linearly with respect to the volume of data
```
mmccarty: > Models must fit in memory no matter the amount of data seen
(constant memory).

Why does the model size stay constant for streaming?

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+for classification. The objective is to predict the target :math:`\hat{y}` for an unknown sample :math:`\vec{x}`. For
+illustrative purposes, this paper focuses on the classification task.
+
+In stream learning, models are trained incrementally, one sample at a time, as new samples :math:`(\vec{x}_t,y_t)`
+become available. Since streams are theoretically infinite, the training phase is non-stop and predictive models are
+continuously updating their internal state in agreement with incoming data. This is fundamentally different from the
+batch learning approach, where models have access to all (available) data during training. As previously mentioned, in
+the stream learning paradigm, predictive models must be resource-wise efficient. For this purpose, a set of requirements :cite:`Bifet2011DataStreamMining` must be fulfilled by streaming methods:
+
+- **Process one example at a time, and inspect it only once.** The assumption is that there is not enough time nor
```
mmccarty: ~~example~~ sample

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+  space to store multiple samples, failing to meet this requirement implies the risk of missing incoming data.
+- **Use a limited amount of memory.** Data streams are assumed infinite, thus storing data for further processing is
+  impractical.
+- **Work in a limited amount of time.** In other words, avoid bottlenecks generated by time-consuming tasks which in
```
mmccarty, line 127: What's the minimum amount of time? Or how is it determined?

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+  the long run could make the algorithm fail.
+- **Be ready to predict at any point.** Stream models are continuously *updated* and must be able to provide predictions
+  at any point in time.
+
+Concept drift
++++++++++++++
+
+A challenging element of dynamic environments is the chances that the underlying relationship between features
+:math:`X` and target(s) :math:`\vec{y}` can evolve (change) over time. This phenomenon is known as **Concept Drift**.
```
mmccarty, line 136: Ah! I see now why you didn't use "drift" above :)
However, above its "change (evolve)" and here its "evolve (change)". Please be consistent.

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+As shown in the previous examples, dynamic environments pose an additional set of challenges to batch learning systems.
+Model degradation is a predominant problem in multiple real-world applications. As enough data has been generated and
+collected, proactive users might decide to train their models to make sure that they agree with the current data. This
+is complicated for two reasons: First, batch models (in general) are not able to use new data into account, so the
+machine learning pipeline must be run multiples times as data is collected over time. Second, the decision for such an
+action is not trivial and involves multiple aspects. For example, should a new model be trained only on new data? This
+depends on the amount of variation in the data. Small variations might not be enough to justify retraining and
+re-deploying a model. This is why a reactive approach is predominantly employed in the industry. Model degradation is
+monitored and corrective measures are enforced if a user-defined threshold is exceeded (accuracy, type I, and type II
+errors, etc.). Fig. :ref:`fig: investment` depicts another important aspect to consider, the tradeoff between the
```
mmccarty: figure reference needs to be fixed

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+:ref:`fig:driftpatterns`, are usually considered: 
+
+- **Abrupt.** When a new concept is immediately introduced. The transition between concepts is minimal. In this case,
+  adaptation time is vital since the old concept becomes is no longer valid.
+- **Incremental.** It can be interpreted as the transition from an old concept into a new concept where intermediate
+  concepts appear during the transition.
+- **Gradual.** When old and new concepts concur within the transition period. It can be challenging since both concepts
+  are somewhat valid during the transition.
+- **Recurring.** If an old concept is seen again as the stream progresses. For example, when the data corresponds to a
+  seasonal phenomenon such as the circadian rhythm.
```
mmccarty: What about **outlier** which is in the figure 2?

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+
+Although the incremental nature of stream methods provides some robustness to concept drift, specialized methods have
```
mmccarty: The term "incremental nature" threw me off here since it was just discussed as part of **concept drift** above. Does this apply to only streams with *incremental* drift? Or all streams because of **concept drift** in general?

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+
+   \ell(y,\hat{y}) = \begin{cases} 0, & y = \hat{y} \\
+                                  1, & y \neq \hat{y}
+                     \end{cases}
+
+Due to the incremental nature of stream leaning methods, special considerations are used to evaluate their performance.
+Two prevalent methods in the literature are *hold-out* and *prequential* evaluation. The hold-out evaluation is a
+popular method in both batch and stream learning where testing is performed on an independent set of samples. On the
+other hand, prequential evaluation :cite:`dawid1984prequential`, is specific to the stream setting. In prequential
+evaluation, tests are performed on new data samples *before* they are used to train (update) the model. The benefit of
```
mmccarty: ~~train (update)~~ retrain

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+this approach is that all samples are used for both test and training.
+
+This is just a brief overview of machine learning for streaming data. However, it is important to mention that the
+field of machine learning for streaming data covers other tasks such as regression, clustering, anomaly detection, to
+name a few. we direct the reader to :cite:`Gomes2017` for an extensive and deeper description of this field, the
```
mmccarty: ~~we~~ We

mmccarty (commented):
Feedback on **Introduction** section.

@jacobmontiel - Looks great so far! I added a few recommendations to help with clarity and asked questions that came to mind. I wanted to get this to you sooner rather than later in case you wanted to discuss. Looking forward to reading further. More to come!

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+  that attackers constantly change their behavior to beat these systems.
+- Supply chain. Several sectors use automatic systems in their supply chain to cope with the demand for products
+  efficiently. However, the COVID-19 pandemic brought to attention the fragility of these systems to sudden changes,
+  e.g., in less than 1 week, products related to the pandemic such as face masks filled the top 10 searched terms in
+  Amazon [#]_. Many automatic systems failed to cope with change resulting in the disruption in the supply chain.
+- Climate change. Environmental data is a quintessential example of the five *v*'s of big data: volume, velocity,
+  variety, veracity, and value. Understanding this data has many implications in our daily lives, e.g., food production
+  can be severally impacted by climate change, disruption of the water cycle has resulted in a rise of heavy rains with
+  the associated risk of floodings. IoT sensors are now making environmental data available at a faster rate and machine
+  learning systems must adapt to this new norm.
```
stsievert: Nit: maybe have one or two in-depth examples with hard numbers rather than 5 brief overviews? I'm thinking of something like this:

> Financial markets generate huge volumes of data daily. Depending on the state of such markets and multiple ex- ternal factors data can become obsolete quickly rendering it useless for creating accurate models. Predictive models must be able to adapt fast to be useful in this dynamic environment.
>
> For example, company XXX records stocks at a 0.5ms resolution, at least for the 1 million stocks that are most relevant. That means that each day, 10TB of data is recorded. The models that process this data are at least 100MB in size and require GPUs. Two challenges immediately arise:
>
> How can models use the 10TB of data to update their internal state? Remember, the model's goal is to do well tomorrow and has all historical data available. How can it utilize historical information and today's information? Of course, the market performance might drift over time or around holidays.

papers/jacob_montiel/jacob_montiel.rst
stsievert: Why does model memory grow with the number of data? Most models I've seen have a fixed number of weights/parameters (linear models, neural nets).

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+depends on the amount of variation in the data. Small variations might not be enough to justify retraining and
+re-deploying a model. This is why a reactive approach is predominantly employed in the industry. Model degradation is
+monitored and corrective measures are enforced if a user-defined threshold is exceeded (accuracy, type I, and type II
+errors, etc.). Fig. :ref:`fig: investment` depicts another important aspect to consider, the tradeoff between the
+investment in resources such as memory and time (and associated cost) and the pay-off in predictive performance. In
+stream learning, resource-wise efficiency is fundamental, predictive models not only must be accurate but also must be
+able to handle theoretically infinite data streams. Models must fit in memory no matter the amount of data seen
+(constant memory). On the other hand, training time is expected to grow sub-linearly with respect to the volume of data
+processed. New samples must be processed as soon as they become available so it is vital to process them as fast as
+possible to be ready for the next sample in the stream.
```
stsievert, line 105: I found this paragraph to be confusing. I got the gist, but I'd make some edits:

``` diff
- On the other hand, training time is
+ Additionally, training time is
```

> training time is expected to grow sub-linearly with respect to the volume of data

Can you expand upon sub-linearly, or provide a reference? I'm familiar with the definition of "sub-linear," and I think I know what you're talking about, but it'd be nice to have more clarity.

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+the stream learning paradigm, predictive models must be resource-wise efficient. For this purpose, a set of requirements :cite:`Bifet2011DataStreamMining` must be fulfilled by streaming methods:
+
+- **Process one example at a time, and inspect it only once.** The assumption is that there is not enough time nor
+  space to store multiple samples, failing to meet this requirement implies the risk of missing incoming data.
+- **Use a limited amount of memory.** Data streams are assumed infinite, thus storing data for further processing is
+  impractical.
+- **Work in a limited amount of time.** In other words, avoid bottlenecks generated by time-consuming tasks which in
+  the long run could make the algorithm fail.
+- **Be ready to predict at any point.** Stream models are continuously *updated* and must be able to provide predictions
+  at any point in time.
```
stsievert, line 130: These are a set of requirements that dictate how the streaming algorithm must behave. What objective function does streaming ML minimize?

It looks like streaming ML tries to detect changes in distribution (that's what the first example shows). Is that accurate?

papers/jacob_montiel/bibliography.bib
```diff
@@ -0,0 +1,166 @@
...
+@ARTICLE{skmultiflow,
+  AUTHOR = {Montiel, Jacob and Read, Jesse and Bifet, Albert and Abdessalem, Talel},
+  URL = {https://github.com/scikit-multiflow/scikit-multiflow},
+  YEAR = {2018},
+  JOURNAL = {Journal of Machine Learning Research},
+  PAGES = {1--5},
+  TITLE = {{Scikit-Multiflow: A Multi-output Streaming Framework}},
+  VOLUME = {19},
+  NUMBER = {72},
+  DOI = "10.5555/3291125.3309634",
```
stsievert: This DOI doesn't resolve for me. This is the page I see when I click on the PDF link: https://dx.doi.org/10.5555/3291125.3309634

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+been proposed to detect drift. Multiple methods have been proposed in the literature, the authors in
+:cite:`Gama2014Survey` provide a thorough survey of this topic. In general, the goal of drift detection methods is to
+accurately detect changes in the data distribution while showing robustness to noise and being resources-wise efficient.
+Drift-aware methods use drift detection mechanisms to react faster and efficiently to changes. For example, the
+*Hoeffding Tree* algorithm :cite:`Domingos2000HT`, a kind of decision tree for streams, does not handle concept drift
+explicitly. An drift-aware version is the *Hoeffding Adaptive Tree* :cite:`Bifet2009HAT`, which uses
+*ADaptive WINdowing* (*ADWIN*) :cite:`Bifet2007ADWIN` to detect drifts. If a drift is detected at a given branch, an
+alternate branch is created and eventually replaces the original branch if it shows better performance on new data.
+
+*ADWIN*, a popular drift detection method with mathematical guarantees, keeps a variable-length window of recent items;
```
stsievert, line 179: > a popular drift detection method with mathematical guarantees

Guarantees under what assumptions?

papers/jacob_montiel/jacob_montiel.rst
```diff
@@ -0,0 +1,533 @@
...
+   SVM - Kappa        : 0.8984
+   SVM - Training time (s)  : 4.70
+   SVM - Testing time  (s)  : 1.73
+   SVM - Total time    (s)  : 6.43
+   SVM - Size (kB)          : 3.4531
+
+In Fig. :ref:`fig:prequential`, we observe the evolution of both estimators as they are trained on data from the stream.
+Although ``NaiveBayes`` has better performance at the beginning of the stream, ``SGDClassifier`` eventually outperforms
+it. In the plot we show performance measured by a given metric (accuracy, kappa, etc.) in two ways: *Mean* corresponds
+to the performance over the entire stream, resulting in a smooth line. *Current* indicates the performance over a
```
stsievert: What's kappa? I think if it's mentioned in two figures it should be defined.

stsievert: I'd also make this change:

``` diff
- *Mean* corresponds to the performance over the entire stream,
+ *Mean* corresponds to the average performance on all data seen previously,
```

stsievert (commented):
Thanks for the submission @jacobmontiel! I'm not an official reviewer but @mmccarty's ping (thanks!) got me interested. This submission is generally well written, and explains the general idea rather well.

However, I'm still not sure what "machine learning for streaming data" means exactly. I get the concept, but I'm having a hard time mapping the concept to objectives/equations. I've asked a couple questions below to help clarify.

mmccarty:
Thank you @martindurant and @stsievert for your feedback! Instructions for reviewers can be found [here](https://github.com/scipy-conference/scipy_proceedings#instructions-for-reviewers) if you would like to learn more about the process.

deniederhut:
> I'm not an official reviewer

You are now. Thanks for joining in Scott!

martindurant:
I am also not officially tagged as a reviewer, but I have some comments that are meant purely as suggestions and hopefully will make the article better.

Firstly, lett me say that skmultiflow is impressive, and the algorithms it presents to users should be very useful when moving beyond the standard batch-learning paradigm. Offering them in the standard sklearn structure is a very wise choice and the paper demonstrates how easy it can be to set up runs and measure accuracy and drift across an input.

My primary concern, stemming from my bias due to working on streamz, is the omission of any mention of real-time analysis. While the introduction references many cases where speed and minimum latency are of the essence, the code examples stream data from static, pre-prepared datasets. One vague mention is made of streaming (in the sense of real-time) data from some protobuf input, but how would this be arranged? Of course, I would suggest strteamz as a way to demonstrate actual dynamic input, coming from, say, a network socket, user interaction or something else.

Some specific comments, mostly from the overview. I will quote text and comment rather than use github code review, I hope that is acceptable.

- Machine learning for streaming data
  - "when the data corresponds to a seasonal phenomena such as circadian rhythm"; either name wether, which is seasonal, or say 'periodic' phenomena
  - "the authors in"; remove
  - "For example, the Hoeffing...explicitly"; remove or shorten and merge with following sentence, since you had said you were about to introduce a drift-aware method
  - "such that it holds that"; this isn't understandable. perhaps "guaranteeing that there has been no change in the data distribution within the window"
  - "the hold out evaluation"; I would remove "the" from all occurrences and there should be a reference at the first mention

- "scikit multiflow"; because the title is in fixed-width, it isn't obviously a section head, i.e., that the overview is now over. Since the next words in the body are the same, perhaps a heading like "describing the software" might be used. Should the package name always be fixed-width?
  - "under the principles of" -> "as"
  - "designed to seemingly interact"; should be "seamlessly"?
  - "compatible with Jupyter"; is there anything concerning Jupyter in the library? Anything python is "compatible"... I would remove this.
  - "democratization of machine learning"; this is a bit much!
  - "different levels of expertise"; this is repeated in the following sentence, can remove.
  - "amount of implementation required" - code required?

- Concusion
  - "we also describe"; surely this is the main point! Just remove "also".

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: This is definitely a typo, thanks for pointing it out. I am fixing it as suggested.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Agree, I tried to provide a generic word for a broader audience, but it might be better to use the field's conventions.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Can you elaborate on this? You mean something like:

> Model degradation is a side-effect of batch learning in many real-world applications requiring additional efforts to monitor, detect, and (if required) enforce mitigation strategies to keep systems functional.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: I agree that there are more steps involved but was worried that it would be distracting for the reader. That is why I opted for a "minimum pipeline" description.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Partly is the cost of re-running the pipeline. But it is also problematic to define *how* to proceed with the re-training. For example if the model was trained with 10M samples and is now useless (because of degradation), should we wait until a similar batch (in volume) is available? Should we merge old and new data? Is this approach sustainable is data is continuously drifting?

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: @mmccarty that is the upper bound. Stream models are designed to either limit the amount of memory or implemented self-maintenance mechanisms. For example, stream ensemble methods usually replace their least efficient members.

@stsievert Yes, there are some methods that are relatively memory efficient. However, a trend in batch learning has been to sacrifice memory for predictive performance. For example, an AlexNet model for ImageNet can use up to 240MB ([reference](https://arxiv.org/pdf/1510.00149.pdf) ) of memory. Now the challenging part, let's consider the setup of data streams where data keeps growing (which is actually the case of ImageNet and its different versions). Then you either train a new (larger) model or create multiple instances (each one taking memory) to account for all the data.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Nice catch.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: That depends on the application. The rule of thumb is in the order of milliseconds per update. The common assumption in the field is that models will be exposed to very fast data streams.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: What about using "drift" in the abstract, and not changing this part?

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: My bad. I am adding the corresponding discussion.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: I see your point. Stream learning is sometimes referred to as "incremental". What about:

> Although the continuous learning nature of stream methods provides some robustness to concept drift

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: "Train" is the term used in the field. Models are "trained" one sample at a time. 

Retrain is not really used as it has the connotation (from batch learning) of creating new models.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Thanks for the suggestion. I will get some numbers for that and will update accordingly.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: A picture is better in this case. This is from a forthcoming [paper](https://arxiv.org/abs/2005.07353):

image

AXGB in the plots are adaptations of XGBoost for data streams.

---

I am adding your suggestion about the wording.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Both batch and stream learning have the same goal of minimizing error. Thus, objective functions are the same general form.

However, the context is different. In stream learning, data is available one sample at a time and it is theoretically infinite. Detecting changes in distribution is only one of the challenges, but it is not its main goal.

papers/jacob_montiel/bibliography.bib
jacobmontiel: That is strange, it is listed here: https://dl.acm.org/doi/abs/10.5555/3291125.3309634

I will get back to you as soon as I figure out what is happening.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Data is a (potentially infinite) sequence of real i.i.d. values.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Thanks, this sounds better. I am changing it.

jacobmontiel:
Thanks for the feedback @martindurant !

> My primary concern, stemming from my bias due to working on streamz, is the omission of any mention of real-time analysis. While the introduction references many cases where speed and minimum latency are of the essence, the code examples stream data from static, pre-prepared datasets. One vague mention is made of streaming (in the sense of real-time) data from some protobuf input, but how would this be arranged? Of course, I would suggest strteamz as a way to demonstrate actual dynamic input, coming from, say, a network socket, user interaction or something else.

I completely agree with this, let me draft something to address this point. I am aware of a project that used scikit-multiflow as an ML library in an edge architecture. I was also experimenting with `streamz` and the idea of the example of a dynamic input sounds very interesting. 

---
> "when the data corresponds to a seasonal phenomena such as circadian rhythm"; either name wether, which is seasonal, or say 'periodic' phenomena

changed
> "the authors in"; remove

removed
> "For example, the Hoeffing...explicitly"; remove or shorten and merge with following sentence, since you had said you were about to introduce a drift-aware method

removed
> "such that it holds that"; this isn't understandable. perhaps "guaranteeing that there has been no change in the data distribution within the window"

changed
> "the hold out evaluation"; I would remove "the" from all occurrences and there should be a reference at the first mention

removed "the" and added reference
> "under the principles of" -> "as"

changed
> "designed to seemingly interact"; should be "seamlessly"?

fixed typo
> "compatible with Jupyter"; is there anything concerning Jupyter in the library? Anything python is "compatible"... I would remove this.

Removed. I agree that anything Python is compatible with Jupyter (although some packages are not designed with that use-case in mind).
> "democratization of machine learning"; this is a bit much!

We have to dream big :-) I reworded it to "democratization of stream learning" to be specific. Let me know what you think.
> "different levels of expertise"; this is repeated in the following sentence, can remove.

removed the following sentence. I prefer this one.
> "amount of implementation required" - code required?

changed to "code required from the user"
> "we also describe"; surely this is the main point! Just remove "also".

:-) removed

deniederhut:
Hey @jacobmontiel ! Thanks to some awesome behind the scenes work by @stargaser, we now have experimental support for publishing papers with ORCIDs. If you have one and would like to add it to your paper, it's as simple as adding an `:orcid:` tag under your name in the paper header. [You can see an example of how to do this in the `00_vanderwalt` example paper](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2020/papers/00_vanderwalt/00_vanderwalt.rst). If you don't have one, no worries! You can still publish your paper as is.

papers/jacob_montiel/jacob_montiel.rst
mmccarty: You got it! Thanks!

papers/jacob_montiel/jacob_montiel.rst
mmccarty: Ah, that makes sense. Thanks!

papers/jacob_montiel/jacob_montiel.rst
mmccarty: Might want to add that note.

stsievert: > Data is a (potentially infinite) sequence of real i.i.d. values.

Is the assumption of an infinite stream of real iid values even relevant? The data in Figure 4 are definitely not identically distributed. Do you mean iid within the window?

papers/jacob_montiel/jacob_montiel.rst
stsievert: > the same goal of minimizing error

The ad hoc goal of supervised machine learning is to perform well on the test set. I don't think that's the stated goal. Do you mean "minimize error on the training set"? Let's say I had one of these "streaming data" models in the real world. How would I test it? For example, would a valid test be deploying a streaming data model to make stock buying/selling decisions?

If that's the case, how relevant is [regret]? Regret tries to minimize the sum of error over all previously seen examples. The financial model example is the stereotypical example because regret is (by definition) the return in dollars.

[regret]:https://en.wikipedia.org/wiki/Regret_(decision_theory)

jacobmontiel:
@martindurant I went looking batck into the discussion we had in the past. What about something like this for an example on data ingestion using `streamz`:

```python
from streamz import Stream
from skmultiflow.trees import HoeffdingTreeClassifier

@Stream.register_api()
class extended(Stream):
    def __init__(self, upstream, model, **kwargs):
        self.model = model
        super().__init__(upstream, **kwargs)

    def update(self, x, who=None):
        # Tuple x represents each sample
        # x[0] are features and
        # x[1] are labels
        y_pred = self.model.predict(x[0])
        # incrementally learn from the sample
        self.model.partial_fit(x[0], x[1])
        # output indicating if the model
        # correctly classified the sample
        self._emit(y_pred == x[1])

s_in = Stream.from_kafka(**config)
ht = HoeffdingTreeClassifier()

s_learn = s.map(read).extended(model=ht)
out = s_learn.sink_to_list()

s_in.start()
```

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: > What's kappa? I think if it's mentioned in two figures it should be defined.

Cohen's kappa coefficient, a.k.a Kappa statistics ([Wikipedia entry](https://en.wikipedia.org/wiki/Cohen%27s_kappa))

jacobmontiel: I added the description of the metrics in the text.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Yes, they are i.i.d. in the window. The goal is to identify that a new distribution is observed as fast as possible while being robust to noise.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: You can have a holdout test set as in the batch setting and use that as an unbiased estimator of performance. On the other hand, you can use the train data (before using it for learning) to test a model. This is known as the prequential evaluation and has the advantage of using all the data. In both cases, you are interested in minimizing errors in the test data. Additionally, in the stream context, since data continuously arrives and can change, most learning methods focus on the most recent test data. In a financial model, you are interested in a model that maximizes profit in the present but is also robust to changes in the data and can adapt accordingly (without human intervention).

martindurant:
> What about something like this for an example on data ingestion using streamz

Sure, something like that would work nicely - not that it needs to go into this paper, of course. You would probably want two node types though, one for training with new labeled data and one for prediction only. I guess in your version, the output list is, effectively, a measure of accuracy with time.

deniederhut:
Hi @mmccarty and @stsievert ! Are you happy with the updates made to the paper?

papers/jacob_montiel/jacob_montiel.rst
mmccarty: Yes, that would work.

mmccarty:
Thanks @deniederhut and @jacobmontiel I'm happy with the changes. Looks great! Its been a pleasure working with everyone on this review. Special thanks to @martindurant and @stsievert for the assistance!

jacobmontiel:
Thanks @mmccarty @martindurant @stsievert 

There are just a couple of details that I would like to add. I was busy with another paper deadline, but I will get those last changes asap. Just bringing this up so the PR is not closed before I push those minor changes.

deniederhut:

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: I have added hard numbers to the existing examples. I kept all the examples as a trade-off to show that there are multiple scenarios where we can find streaming data.

papers/jacob_montiel/bibliography.bib
jacobmontiel: There seems to be a problem with that DOI. I have removed it from the reference and added the URL of the paper which is open.

jacobmontiel:
> You would probably want two node types though, one for training with new labeled data and one for prediction only.

@martindurant I added a new section to discuss real-time applications. I kept the example for simplicity but added a note regarding the above possibility as long as we keep the *test-then-train* order.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Sorry, I lost track of this item. I agree and have re-worded this section based on your suggestion. I am quite happy with the result.

papers/jacob_montiel/jacob_montiel.rst
jacobmontiel: Changed :-)

jacobmontiel:
@martindurant I added the *Streamz* repository in a footnote as I couldn't find another way to cite it. Is this ok with you?

martindurant:
Certainly!

On July 2, 2020 2:48:18 AM EDT, Jacob <notifications@github.com> wrote:
>@martindurant I added the *Streamz* repository in a footnote as I
>couldn't find another way to cite it. Is this ok with you?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/scipy-conference/scipy_proceedings/pull/566#issuecomment-652818000

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

papers/jacob_montiel/bibliography.bib
```diff
@@ -0,0 +1,196 @@
...
+
+@ARTICLE{Numpy,
+  author={S. {van der Walt} and S. C. {Colbert} and G. {Varoquaux}},
+  journal={Computing in Science   Engineering}, 
+  title={The NumPy Array: A Structure for Efficient Numerical Computation}, 
+  year={2011},
+  volume={13},
+  number={2},
+  pages={22-30},
+}
```
stargaser, line 138: please add `doi = "10.1109/MCSE.2011.37" per https://www.scipy.org/citing.html#numpy

stargaser (commented):
Hello, I'm one of the proceedings co-chairs, doing a formatting and reference check. Overall looks good: I only noted one NumPy reference that didn't have a doi, commented in-line.

papers/jacob_montiel/bibliography.bib
jacobmontiel: Thanks for pointing this out @stargaser. I have added the DOI.

PR #569, opened 2020-05-30
Paper: Compyle: Python once, HPC anywhere!

prabhuramachandran:
This is a draft of the paper and should be complete in a few more days.


serge-sans-paille:
Hi, I'm @serge-sans-paille, I'll be reviewing your paper, trying to provide valuable feedbacks!

serge-sans-paille:
Some early thoughts after a first scan of the paper: it seems important to mention that compyle introduces extra operators (`Map`, `Reduce`, `Scan`) and that it optimized a DSL based on the combination of these operators with elements wise operations. Maybe discuss the advantage and limitation of the approach.

This approach is *very* similar to this academic work:  https://ctnzr.io/papers/PPoPP-2011.pdf

prabhuramachandran:
@serge-sans-paille -- thank you!  We are still working on the paper and should wrap it up by this weekend.  Thanks for the reference!  Very useful indeed.

prabhuramachandran:
@serge-sans-paille -- Thanks for your patience! The paper should now be ready for review.  We've incorporated your suggested reference as well.

govarguz:
Dear @prabhuramachandran, dear @adityapb, 
compyle aims to port a python prototype code to a multi-core and GPU environment based on main parallel-mediated algorithms. As it is the case of `map`, `reduce` and `scan`. The focus of compyle is to effectively port a python proof-of-concept code to HPC-environments, which is a relevant functionality nowadays that the use of e.g. GPUs is more accessible than in the past. The example you have chosen is also relevant, although a 3D MD Lennard-Jones code would provide more insight and more realistic performance numbers and a comparison for the potential users-base. You could in principle compare it to some of the MD packages with a python mediated interface like, [E++](https://github.com/espressopp/espressopp) or [HooMD](https://github.com/glotzerlab/hoomd-blue).
In this direction, I would suggest to add two plots:
 
1. Strong scaling ( N particles remain fixed, number of processors increase form 8,16...till for example 128) comparison with only CPUs (here I am not sure if communications between nodes is also built, please remark this in the Ms.). 

2. Weak scaling (N number of particles vary and the number of processors remain fix ) comparison of both codes, as well, Compyle PoC and E++ (or HooMD).

The available plots are already quite illustrative comparing cython serial, CUDA and OpneCL. However, in my opinion adding points 1 and 2, would provide an enhanced discussion on how this programming paradigm compares to "a certain package" with a python interface.

deniederhut:
Hi @serge-sans-paille ! Have you had a chance to look through the paper since the most recent changes?

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,896 @@
...
+from them requires the use of parallel algorithms. Unless the initial
+development is done with this in mind, one cannot easily convert a serial code
+into a parallel one.
+
+There are many powerful tools available in the Python ecosystem today that
+facilitate high-performance computing. PyPy_ is a Python implementation in
+Python that features a JIT that allows one to execute pure Python code at
+close to C-speeds. Numba_ uses the LLVM_ compiler infrastructure to generate
+machine code that can rival native C code. Numba also supports execution on
+GPUs. There are also libraries like Pythran_ that transpile a subset of Python
```
serge-sans-paille: Pythran is a compiler, not a library :-)

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,896 @@
...
+to C++ and support multi-core execution using OpenMP. Packages like cppimport_
+and pybind11_ make it a breeze to integrate Python with C++ code. In addition,
+there are powerful interfaces to GPUs via packages like PyOpenCL_ or PyCUDA_.
+Of these, Numba has matured a great deal and is both easy to use and
```
serge-sans-paille: Cython should be mentioned in this paragraph too. Especially as it provides some loop parallelism directives.

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,896 @@
...
+versatile.
+
+Given this context, one may wonder why Compyle exists at all. While compyle
+grew out of a project that pre-dates numba, the real reason that compyle
```
serge-sans-paille: nit: consistent capitalization of compyle and numba.

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,896 @@
...
+look at a simple molecular dynamics simulation where :math:`N` particles
+interact with each other via a Lennard-Jones potential. This problem is
+discussed at length in :cite:`schroeder2015`.
+
+In order to implement this, the typical workflow for a Python programmer would
+be to prototype the molecular dynamics simulation code in pure Python and
+obtain a proof of concept. One would then optimize this code so as to run
+larger problems in a smaller amount of time. Very often this would mean
+changing some data structures, writing vectorized code using NumPy arrays, and
+then resorting to tools like numba to extract even more performance. Numba is
```
serge-sans-paille: Maybe mention that going to numba / cython means devectorizing to make the looping explicit?

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,896 @@
...
+useful computational codes involve non-trivial algorithms, which are not
+always easy to parallelize.
+
+What compyle attempts to do is to allow one to write code **once** in a highly
+restrictive subset of pure Python and have this run in parallel on both CPUs
+and GPUs. This is a significant difference from all the tools that we have
+mentioned above.
+
+The difficulty in doing this is that it does require a change in approach and
+also a loss of the typical conveniences with high-level Python.
```
serge-sans-paille: Maybe argument that low-level Python still has some pros versus a lower-level language here?

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,896 @@
...
+* Specific support to run code on a particular backend. This is for code that
+  will only work on one backend by definition. This is necessary in order to
+  best use different hardware and also use differences in the particular
+  backend implementations. For example, the notion of local (or shared) memory
+  only has meaning on a GPGPU. In this category we provide support to compile
+  and execute Cython code, and also create and execute a GPU kernel. These
+  features are not discussed in this article.
+
+In general the subset of Python that compyle supports are:
+
```
serge-sans-paille, line 233: based on these restrictions, it appears compyle is more of a DSL for computation kernels embed into a restricted subset of Python. This is very different from what the title advertised, maybe change the title to reflect that concept?

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,896 @@
...
+    Time taken for simulation using :math:`O(N)` (Linear) and :math:`O(N^2)`
+    (Simple) approach. :label:`time-nnps-vs-simple`
+
+.. figure:: speedup_comp_impl.png
+
+    Speed up using :math:`O(N)` over :math:`O(N^2)` approach.
+    :label:`nnps-simple`
+
+Figure :ref:`speedup-nnps` shows the speedup relative to serial code running
+using Cython for the OpenCL and CUDA backends using the NNPS algorithm and
```
serge-sans-paille: I have difficulties parsing this sentence :-)

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,896 @@
...
+
+While compyle is really powerful and convenient, it does use a rather verbose
+and low-level syntax. In practice we have found that this is not a major
+problem. The more serious issue is the fact that we cannot directly use
+external libraries in a platform neutral way. For example, there are ways to
+use an external OpenCL or CUDA library but this will not be usable on a CPU.
+Obviously one cannot use normal Python code and use basic Python data
+structures. Furthermore, one cannot use well established libraries like scipy
+from within the parallel constructs. These are limitations that are beyond the
+scope of compyle at this point.
```
serge-sans-paille: Explain why: there's no way compyle - generated code would run on the interpreter, etc

serge-sans-paille (changes requested):

deniederhut:
Hey @prabhuramachandran  ! Thanks to some awesome behind the scenes work by @stargaser, we now have experimental support for publishing papers with ORCIDs. If you have one and would like to add it to your paper, it's as simple as adding an `:orcid:` tag under your name in the paper header. [You can see an example of how to do this in the `00_vanderwalt` example paper](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2020/papers/00_vanderwalt/00_vanderwalt.rst). If you don't have one, no worries! You can still publish your paper as is.

prabhuramachandran:
Hi @deniederhut -- thanks for this!  We will add our ORCIDs as we address the reviewer comments this weekend.

prabhuramachandran:
Dear @govarguz,

Thank you for going over our submission and for your comments!  Our responses are inline below.

> Dear @prabhuramachandran, dear @adityapb,
> compyle aims to port a python prototype code to a multi-core and GPU environment based on main parallel-mediated algorithms. As it is the case of `map`, `reduce` and `scan`. The focus of compyle is to effectively port a python proof-of-concept code to HPC-environments, which is a relevant functionality nowadays that the use of e.g. GPUs is more accessible than in the past. The example you have chosen is also relevant, although a 3D MD Lennard-Jones code would provide more insight and more realistic performance numbers and a comparison for the potential users-base. You could in principle compare it to some of the MD packages with a python mediated interface like, [E++](https://github.com/espressopp/espressopp) or [HooMD](https://github.com/glotzerlab/hoomd-blue).

Thank you for the suggestion.  Our own work is not in the area of molecular dynamics.  Our only goal was to take a non-trivial problem and demonstrate how we are able to write from scratch a pure Python solution that immediately runs on CPUs and GPUs.  We also picked a 2D problem just to keep things simple and in line with the simple paper we cited. We do not have much experience with the two packages you have mentioned.  We will try to see if we are quickly able to get a working example.  One issue is that many of these packages are fairly complex so getting an apples to apples comparison is probably going to take a bit of time.  

> In this direction, I would suggest to add two plots:
> 
> 1. Strong scaling ( N particles remain fixed, number of processors increase form 8,16...till for example 128) comparison with only CPUs (here I am not sure if communications between nodes is also built, please remark this in the Ms.).

Our package by itself does not handle distributed computing at all.  We do have a separate package for this but that is not ready yet.  Given this, I am not sure how we can demonstrate scale up with 128 processors.

> 2. Weak scaling (N number of particles vary and the number of processors remain fix ) comparison of both codes, as well, Compyle PoC and E++ (or HooMD).
> 

We can try to do this.

> The available plots are already quite illustrative comparing cython serial, CUDA and OpneCL. However, in my opinion adding points 1 and 2, would provide an enhanced discussion on how this programming paradigm compares to "a certain package" with a python interface.

We agree that this would be a useful comparison and will try to at least address a simple comparison even if we are not able to do everything suggested.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: Sorry, fixed.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: Done, thanks.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: I've tried to fix this over the entire manuscript.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: I have added a remark to note this.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: I have tried to add a little text to elaborate your comment in the text.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: I do agree I am not quite sure I should call it a DSL since really it is strictly a Python subset and all the code is valid Python.  I agree that the title is bombastic and have changed it to "Compyle: a Pythonic DSL for parallel computing".  I hope that works.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: Hopefully I've fixed that.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: I have attempted to but am not sure if what I've done is good enough.

prabhuramachandran:
@serge-sans-paille -- thanks for your comments!  I have tried to address all of them, I have left comments and not resolved the comments (I'm not sure if that is what I am supposed to do).  Please let me know if it looks better now.  Thanks.

deniederhut:
>  I have left comments and not resolved the comments (I'm not sure if that is what I am supposed to do)

🤷‍♀️ different reviewers have different preferences, so this is not something we have prescribed a rule for.

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,910 @@
...
+mentioned above.
+
+The difficulty in doing this is that it does require a change in approach and
+also a loss of the typical conveniences with high-level Python. While Compyle
+does not allow arbitrary Python code, since the code is still written in
+Python and not another language, it makes it much easier for users to write
+and manage the code.
+
+Compyle provides important parallel programming algorithms that one typically
+requires to in order to write parallel programs. These are the element-wise
```
grlee77: seem like a typo is making this sentence hard to read
maybe change "requires to in order to write" to "requires when writing" ?

papers/compyle_pr_ab/compyle.rst
```diff
@@ -0,0 +1,910 @@
...
+operations (or maps), reductions, and parallel prefix scans. These primitives
+are written such that the same program can be executed on both multi-core CPUs
+and GPUs with minimal or no changes to the code.
+
+This is currently not possible with any of the other tools. In addition,
+Compyle has the following features:
+
+- Generates either Cython or ANSI C code depending on the backend and this
```
grlee77, line 135: I think this Cython, or CUDA/OpenCL C source code vs. a more opaque intermediate representation is attractive when things don't work as expected and one may want to manually check the generated code. Also, most scientific programmers are much more likely to be familiar with C than LLVM IR or assembly code.

I didn't see anywhere in the manuscript or online docs where it is stated how the user can view the transpiled C or Cython code? I tried some manual tinkering with the Elementwise examples provided and found that one can access a `.tp.get_code()` attribute. I would make a brief mention in the paper that there is a method by which the user can retrieve the transpiled code and document the public API for this in the online docs.

papers/compyle_pr_ab/compyle.rst
serge-sans-paille: Great! Being a strict subset of Python make it an *eDSL*, an *Embedded DSL*

prabhuramachandran: Well, to be honest, I do not like the word DSL as it seems like jargon.  The word is itself imprecise.  So I find that the word doesn't really improve understanding or make anything clearer.  Is NumPy a DSL for scientific computing?  Does knowing that it is a DSL help anyone who may want to use it?  I am not so sure.   Do you think this works: "Compyle: a Python package for parallel computing"?  I agree that the original title is not great but am not entirely happy with throwing in the word DSL or eDSL.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: Thanks for pointing this out.  This is something we have wanted to improve but it has fallen by the wayside.  The complete Cython files are written inside `~/.compyle/source` but not the GPU code.  We are wondering if we should have an environment variable to help do this so it prints it out?  Could you let us know of your thoughts on this?  A method is surely something we should expose cleanly and document but anything else you would find useful?

grlee77: CuPy has [some environment variables](https://docs-cupy.chainer.org/en/stable/reference/environment.html#environment-variables) related to this type of thing which I have found useful:
- `CUPY_DUMP_CUDA_SOURCE_ON_ERROR` causes the source to be printed to the terminal if there is an error during compilation.
- `CUPY_CACHE_SAVE_CUDA_SOURCE` causes the `.cu` source to be cached in addition to the `.cubin` files that are always cached.

grlee77:
Also, can you double-check figure 3? I find it surprising that the CUDA and OpenCL results are aligned so exactly. When I ran `examples/molecular_dynamics/peformance_comparison.py` locally, I got:

linear_speedup_opencl_cuda_cython

i.e. comparable, but not identical performance

adityapb:
> Also, can you double-check figure 3? I find it surprising that the CUDA and OpenCL results are aligned so exactly. When I ran `examples/molecular_dynamics/peformance_comparison.py` locally, I got:
> 
> linear_speedup_opencl_cuda_cython
> 
> i.e. comparable, but not identical performance

Hi @grlee77, the plot you have generated is using the `O(N)` algorithm. Figure 3 shows the speedup using the simple `O(N^2)` implementation. For getting the result corresponding to figure 3, you can run the following:

```
python performance_comparison.py --nnps simple
```

I have double checked the plot generated using the above command and I am getting the same result.

The `O(N)` comparison is showed in figure 4 which is quite similar to what you have generated. I think I'll perhaps add more details of how each of the results can be reproduced in the README of the molecular_dynamics example.

prabhuramachandran:
Dear @govarguz,

> 1. Strong scaling ( N particles remain fixed, number of processors increase form 8,16...till for example 128) comparison with only CPUs (here I am not sure if communications between nodes is also built, please remark this in the Ms.).
> 2. Weak scaling (N number of particles vary and the number of processors remain fix ) comparison of both codes, as well, Compyle PoC and E++ (or HooMD).

Thanks for your suggestions, as I had mentioned earlier, doing 1 is a bit difficult currently. @adityapb has implemented a 3D version of the code and has compared the results of HooMD and our code and added the plots for this.  Our code is about 2x faster than HooMD on a GPU.  Since we weren't sure if we had made a mistake he has also verified that the results are indeed similar for a smaller problem as well.  All of these have been added to the manuscript.  We hope this addresses the main concern of a reasonable and fair comparison.  The paper has exceeded 8 pages right now so if this is a problem we can try to compress it a bit by merging some of the plots together (the KE and PE for example).

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: I've changed it to "Compyle: a Python package for parallel computing" and mention eDSL in the abstract.  I hope this is OK.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: Thanks for the suggestion!  That is very useful.  We will address some of these in our next release.  Hopefully we will be able to cut a new release before SciPy starts.

prabhuramachandran: Also you could do `.tp.source` to get the source but it isn't as clean as I would like.  We will address this as well in the next release (and update the docs as well).

prabhuramachandran:
@deniederhut -- I believe we've addressed the major issues and suggestions raised thus far.

govarguz:
Dear @prabhuramachandran, @adityapb,

good to hear about the reached speed-up! I am going through the new _Ms._ version in this weekend and submit my latest reviews by tomorrow. 
Write to you soon!
Horacio.

prabhuramachandran:
@govarguz -- Thank you!  BTW, I think a head-to-head comparison is never fair, so we are just happy that we are in the same ballpark (being faster is nice).  Our intention is not to claim anything special -- just that it is possible to get good performance.  We hope the text does not suggest that we are taking a potshot at HooMD.

govarguz:
Agreed!

papers/compyle_pr_ab/compyle.rst
```diff
@@ -13,27 +13,30 @@
...
    Compyle allows users to execute a restricted subset of Python on a variety
-   of HPC platforms. It currently supports multi-core execution using Cython,
-   and OpenCL and CUDA for GPU devices. Users write low-level code in Python
-   that is automatically transpiled to high-performance Cython or C. Compyle
-   also provides a few very general purpose and useful parallel algorithms
-   that allow users to write code once and have them run on a variety of HPC
-   platforms.
+   of HPC platforms. It is an embedded domain-specific language (eDSL) for
+   parallel computing. It currently supports multi-core execution using
+   Cython, and OpenCL and CUDA for GPU devices. Users write low-level code in
```
govarguz: here the term "low-level" may cause confusion, why note single core or "not parallelized"

papers/compyle_pr_ab/compyle.rst
```diff
@@ -13,27 +13,30 @@
...
+   Compyle also provides a few very general purpose and useful parallel
+   algorithms that allow users to write code once and have them run on a
+   variety of HPC platforms.
 
    In this article, we show how to implement a simple two-dimensional
    molecular dynamics simulation package in pure Python using Compyle. The
    result is a fully parallel program that is relatively easy to implement and
    solves a non-trivial problem. The code transparently executes on multi-core
-   CPUs and GPGPUs allowing simulations with millions of particles.
+   CPUs and GPGPUs allowing simulations with millions of particles. A 3D
```
govarguz: A 3D MD code , instead of just "3D"  perhaps MD can be defined a bit upper when molecular dynamics appears first in the text.

papers/compyle_pr_ab/compyle.rst
```diff
@@ -828,14 +833,13 @@ performance. This can be done easily in Compyle using
...
-Figure :ref:`time-hoomd` shows the results of this comparison.
-We found our implementation to be about 2x faster than HooMD.
-Figures :ref:`hoomd-pe` and :ref:`hoomd-ke` show the kinetic and potential
-energy for 1,500 particles simulated for 10,000 timesteps in periodic
-boundary conditions.
+We have also implemented a 3D version of the simulation with both periodic and
+non-periodic boundary conditions. We compared our implementation with HooMD_
+for a 3D periodic simulation on an NVIDIA Tesla P100 GPU. Figure
+:ref:`time-hoomd` shows the results of this comparison. We found our
+implementation to be about 2x faster than HooMD. Figures :ref:`hoomd-pe` and
```
govarguz: I would move both plots, KE and PE to an appendix section within the article.

govarguz (approved):

papers/compyle_pr_ab/compyle.rst
grlee77: Yes, although it is good to know that the results are similar, I don't think the energy levels are important to the main point of the paper.

grlee77:
> Hi @grlee77, the plot you have generated is using the O(N) algorithm. Figure 3 shows the speedup using the simple O(N^2) > implementation. For getting the result corresponding to figure 3, you can run the following:
> 
> python performance_comparison.py --nnps simple
> 
> I have double checked the plot generated using the above command and I am getting the same result.

Okay, I ran that one and I got curves that were more similar, but still not exactly overlapping. I suppose it is probably just some difference in the hardware used. I suppose if they are so closely overlapping you could change the "x" marker on the blue line to an "o" or some other symbol that would be more visible in the overlapped plot, but it is not a big deal either way.

grlee77:
One other related library that may be worth mentioning is [Reikna](http://reikna.publicfields.net/en/latest/). I have never looked at it in detail, but my impression is that it also allows GPGPU programming via templated code that gets translated to either PyOpenCL or PyCUDA kernels. In that respect it has some similarity to the current project, but it does not have any equivalent of Compyle's Cython code generation capability. 

Overall my impression of the paper is favorable. I think the writing is good and the library seems to fill a niche in being able to transpile common code to Cython, OpenCL or CUDA. I am not aware of another tool for that. There are limitations to what can be done, but I think the authors did a good job of pointing those out and making clear what the tool can and cannot do.

I did try running the provided examples and the code worked as advertised.

deniederhut:
Sweet! Do I understand correctly that there are just a few suggested changes left?

prabhuramachandran:
@deniederhut -- yes, there are a few suggestions that we will get to in a day or two, does that work?

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: Since these are tangential to the paper anyway, can we skip these two plots and just mention that the values are close in the text?

deniederhut:
The open review period ends in two days. We'll probably want to make those changes quickly so your reviewers have time to take a final look at the paper.

papers/compyle_pr_ab/compyle.rst
grlee77: That sound fine to me, especially given that the example script to recreate the result is now in the repository.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: This is now fixed.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: I've made this more precise to "Users write code in a restricted subset of Python that is automatically ...".  I believe this should clarify it.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: Done, thanks.

papers/compyle_pr_ab/compyle.rst
prabhuramachandran: Yes, the energy plot code was not there earlier and we just merged a PR that does this.

prabhuramachandran:
Thanks all, I have pushed the last set of changes which address the remaining comments and also cites Reikna.  We also added a link to the colab notebook if readers wish to try the examples with a GPU. We also added a small acknowledgments section at the end which I felt was necessary.  Thank you all for your patience and feedback!
@deniederhut -- I think at this point the manuscript is mostly done.

govarguz:
The Ms. has interesting points and it is applicable. However, I think there might be a tiny addition to make, while keeping the example of MD in mind. _**Compyle, can be a potential tool for the MD-method development community. In particular for the proof-of-concept methods that require parallelization**_. Hence, I strongly suggest to add such a line in the _future work_ or _conclusions_ section which shall be citing a couple of article in this regard. Such as: https://doi.org/10.1016/j.cpc.2018.12.017 and  https://doi.org/10.1039/C4SM02344G. And I agree the Ms. is mostly done!

prabhuramachandran:
@govarguz -- I am not sure I would be comfortable doing that as I am not an expert on MD. The point of the MD example was just to pick a non-trivial example that required a scan.  We could have used a more fluid-centric example which is up my alley but ends up making the code more complex.  The intent was not to suggest anything specific be done in the area of MD-method development (maybe MD folks have enough tools and don't need another one, I do not know).  I do know that there are many many good MD codes that are available (I'm told everyone has their favorite).  So it would be inappropriate for us to suggest anything to the community without being a part of it. In this particular case, I would essentially be paraphrasing your words and your experience, which I am not sure we should do. We are definitely very happy if Compyle proves useful for MD-method development but what we are getting at is that Compyle is potentially useful to many other folks too.   I am not sure if I am communicating this clearly here.

govarguz:
@prabhuramachandran , sure, I see your point. And exchanging final ideas is the aim of this review.
 For sure compyle can be useful for many folks. However, my suggestion is precisely in those lines to name a few of those communities. As for example, the MD-method-development (in general as you picked the MD example!) by citing some of the perspective methods developments (in a proof-of-concept fashion). In addition, to my eyes scientific pythonic tools may reach a lot of communities in open-source.
I don't want you to rephrase my text, but cite a community that can potentially profit of Compyle, give it a little visibility, through a couple of citations.

prabhuramachandran:
OK, that makes sense.  In any case, it helps to know that this may be useful in the MD domain (am happy to take your word for it).  I've added the reference and made a few edits to have everything fit in 8 pages.  I hope that works.

papers/compyle_pr_ab/references.bib
```diff
@@ -0,0 +1,82 @@
...
+                  Kaushik, Rohan and Patil, Mrinalgouda and Tavker, Deep and
+                  Menon, Dileep and Kurapati, Vikas and Sebastian, Amal S and
+                  Dutt, Arkopal and Agarwal, Arpit },
+  journal={arXiv preprint arXiv:1909.04504},
+  year={2019},
+  url={https://arxiv.org/abs/1909.04504}
+}
+
+@InProceedings{pysph16,
+  author    = {Prabhu Ramachandran},
```
deniederhut, line 32: @prabhuramachandran could I ask you to add the DOI for this reference? http://conference.scipy.org/proceedings/scipy2016/prabhu_ramachandran_pysph.html

prabhuramachandran: Thanks for catching that!  I had an older reference entry.  This is now fixed.

cbcunc (approved):
Passes final proceedings committee walk-through.

prabhuramachandran:
I am just pushing a change for some grammatical errors I found.

PR #570, opened 2020-05-30
Paper: pandera: Statistical Data Validation of Pandas Dataframes

cosmicBboy:
PDF build link: http://procbuild.scipy.org/download/cosmicBboy-nbantilan-pandera

I'm still working on adding a figure or two, should be complete in the next few days.


cosmicBboy:
This should be ready for review!

chendaniely:
Hi @cosmicBboy, nice to see you here!

To the other reviewers: I know Niels from my master's program,
and I recognize `pandera` from the pyOpenSci submissions as well
(I didn't review his pyOpenSci submission).
If this means I need to retract my review, just let me know.

This project has been independently accepted by pyOpenSci,
so it has already gone through an independent review of code and documentation.
For a talk submission, I think this will be a great addition to the scipy talk tracks around reproducibility and data management.

The author does a good job surveying similar tools in the python ecosystem, and how `pandera` can complement the pandas API with a data validation API.
The code examples in the paper are easy to follow (although I did not attempt to run them on my computer).

The biggest change to the paper would be removing gender and sex from your examples, which is to say there is nothing technically wrong with the paper.
In particular the line "`income` negatively correlated with dummy variable `is_female`."
The examples can be switched to using 'E'xperiment and 'P'lacebo to be consistent with an example later on in the paper.

For figure 2, I would split the "error" compartment so it shows things that can be returned in the "SchemaError": invalid data, and failure cases.

The only question that I had with the package is checking the data type of columns,
do you have to pass in `pa.type`, can we also use numpy types or built-ins?
If you can use `int` or `float` built-ins, I would change the example on page 4 when the schema is first created to use other types so the user can see other types can be used to validate type.

cosmicBboy:
Thanks for the review @chendaniely! Before making the edits, I wanted to get some clarification on this feedback:

> The biggest change to the paper would be removing gender and sex from your examples, which is to say there is nothing technically wrong with the paper.

What is the rationale for such a change? Is the concern that this is an example using a sensitive/protected class?

> In particular the line "income negatively correlated with dummy variable is_female."

If this is the case I think it would be more within the context of that section of the paper to come up with alternative examples that doesn't use such variables, as opposed to referencing a later example for the sake of consistency. For instance, the statement can us a variable like `job_category` and the statement in question can be modified to: "income is negatively correlated with the dummy variable `job_category = food services`, which is a variable derived from the `job_category` column".

> For figure 2, I would split the "error" compartment so it shows things that can be returned in the "SchemaError": invalid data, and failure cases.

👍

> The only question that I had with the package is checking the data type of columns,
do you have to pass in pa.type, can we also use numpy types or built-ins?

python built-in types are currently not supported, but numpy type string aliases are e.g. `float64` or `datetime64[ns]`, I can modify a few of the column definitions to use those with a comment.

chendaniely:
re: referncing gender in examples -- yes it's to avoid using a binary gender example and probably better if you can construct an example that does not use it.
It's only something I picked up on because of feedback from teaching/example materials I've used in the past.

chendaniely:
Also to the conference organizers and other reviewers: this talk topic was also submitted to SciPy last year as a tutorial submission (it was not accepted). I do believe that the talk format is a much better fit for the conference.

deniederhut:
I'm glad we found a good place for it at the conference 😄

deniederhut:
Hey @cosmicBboy ! Thanks to some awesome behind the scenes work by @stargaser, we now have experimental support for publishing papers with ORCIDs. If you have one and would like to add it to your paper, it's as simple as adding an `:orcid:` tag under your name in the paper header. [You can see an example of how to do this in the `00_vanderwalt` example paper](https://raw.githubusercontent.com/scipy-conference/scipy_proceedings/2020/papers/00_vanderwalt/00_vanderwalt.rst). If you don't have one, no worries! You can still publish your paper as is.

cosmicBboy:
thanks @deniederhut! I do have an orcid ID, I'll add it along with my edit to address @chendaniely's review over the weekend

aheilbut:
The author describes pandera, a data validation library for pandas.
The paper is clear and well written, and introduces the pandera features and interfaces very well.  Schemas and statistical data validation are underappreciated and underutilized, and I can see a library such as pandera being very useful and an important contribution to the pandas ecosystem. 

I have a couple of relatively minor suggestions.
1. It would be helpful to include a bit more explanation of the motivation for pandera in the context of the related tools that are mentioned (great_expectations, schema, and pandas_schema).  However, it isn't clear what exactly motivated the author to develop pandera (ie. were there features missing or deficiencies in the other implementations?)   

2. If there is an example of a "real-world" application of pandera that could be briefly described, that would also make the paper stronger.  As the author discusses, one of the major trade-offs of using a tool such as this is the additional development time required to define schemas.  A case study in which pandera was demonstrated to be useful in a larger project would help to justify its importance.

cosmicBboy:
Thanks for the review @aheilbut! I can re-work the section on the motivation behind developing `pandera`: to be honest, I basically wanted to combine concepts/designs from the packages mentioned in **related tools** (and some other packages that are no longer maintained) and build something light-weight with easy customizability in mind.

> If there is an example of a "real-world" application of pandera that could be briefly described, that would also make the paper stronger.

I use `pandera` extensively in my work at Talkspace  to validate data going through out ETL pipelines, ML modeling infrastructure, and also jupyter notebooks for ad hoc analyses, unfortunately I'm not really at liberty to go into detail on any particular project as a case study. Do you think I'd be able to achieve a similar effect if I outlined some high-level descriptions of use cases like the ones I mention above? In my talk I'll also go through an analysis using a publicly available dataset.

deniederhut:
Speaking only for myself, I always like to see motivating examples and case studies. It's totally understandable that you can't talk about your application, and your suggestion about doing this on an open source dataset (Yelp reviews or something?) is a good one.

aheilbut:
I think it'd help even if you described at a pretty abstract level how it was used.. ie. you have an pipeline that involves 20 different tables, each one has a schema, and maybe one or two more specific examples of why that is useful because it helped you catch a particular type of error that could occur.  

On the issue of motivations and the other packages, I think you want to make sure there is a clear argument in the paper that pandera is superior and necessary for some reason.  ie. be really explicit that you needed to combine features A & B and couldn't do that otherwise, or package X implements feature Y wrong, or whatever it was that must have annoyed you enough to build this yourself rather than just using those other packages

cosmicBboy:
Just updated the **related tools** section with more justification for `pandera` and added a **Use Case Vignette** section as per @aheilbut's review comments

deniederhut:
Awesome! @aheilbut are you happy with the recent changes?

deniederhut:

aheilbut:
Yes, I think that additional justification and discussion about the other packages is very helpful.  Looks good!

papers/niels_bantilan/refs.bib
```diff
@@ -0,0 +1,192 @@
...
+}
+
+@book{van2018statistical,
+  title={Statistical data cleaning with applications in R},
+  author={Van der Loo, Mark and De Jonge, Edwin},
+  year={2018},
+  publisher={Wiley Online Library}
+}
+
+@misc{reback2020pandas,
```
deniederhut: Could we use the Pandas paper citation here? https://conference.scipy.org/proceedings/scipy2010/mckinney.html

papers/niels_bantilan/refs.bib
```diff
@@ -0,0 +1,192 @@
...
+@inproceedings{rocklin2015dask,
+  title={Dask: Parallel computation with blocked algorithms and task scheduling},
+  author={Rocklin, Matthew},
+  booktitle={Proceedings of the 14th python in science conference},
+  number={130-136},
+  year={2015},
+  organization={Citeseer}
+}
+
+@article{ma2019pyjanitor,
```
deniederhut: Looks like this citation is missing a lot of info -- http://conference.scipy.org/proceedings/scipy2019/eric_ma.html

deniederhut:
Thanks!

PR #575, opened 2020-06-19
add code spellchecks

ocefpaf:
@deniederhut this will add a GitHub Action that will act like a CI to check for misspelled words with `codespell`. I also added 3 other pre-commit checks that may be useful here and help with review/diffs: `trailing-whitespace`, `end-of-file-fixer`, and `check-added-large-files`.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -1,0 +1,20 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [master]
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+    - name: set PY
+      run: echo "::set-env name=PY::$(python --version --version | sha256sum | cut -d' ' -f1)"
+    - uses: actions/cache@v1
+      with:
+        path: ~/.cache/pre-commit
+        key: pre-commit|${{ env.PY }}|${{ hashFiles('.pre-commit-config.yaml') }}
+    - uses: pre-commit/action@v1.0.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+repos:
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v2.5.0
+  hooks:
+    - id: trailing-whitespace
+    - id: end-of-file-fixer
+    - id: check-added-large-files
+
+- repo: https://github.com/codespell-project/codespell
+  rev: v1.16.0
+  hooks:
+    - id: codespell
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Attribution (CC By) license.
 - Papers are openly available at http://conference.scipy.org/proceedings/, with
   no pay walls barring consumption or author processing charges.
 
-- From 2015 onward, papers have DOIs (making them easily citable) and are also
+- From 2010 onward, papers have DOIs (making them easily citable) and are also
   openly available from those DOIs.
 
 The community is involved in the entire process for creating the proceedings,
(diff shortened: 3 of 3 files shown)


deniederhut:
Thanks for the PR! I haven't used GitHub actions, so I'll need a minute to look into the setup for these. Three immediate thoughts:

- We don't use the master branch (I will remove that presently) -- does the action need to specify that the branch is 2020?
- If so, does that mean we need to update the yaml for the action every year? Or is it easy to have this check PRs on any branch?
- Could I ask you to rebase this onto dev? We've found it easier to manage the history of code changes by keeping a trunk (dev) that is free of any paper commits, and then branching individual conference years (like 2020) from there.

ocefpaf:
> * We don't use the master branch (I will remove that presently) -- does the action need to specify that the branch is 2020?

It will run as a CI service "failing" when there are misspelled words. (Which BTW can lead to false positives so we should use it just as a guide.)

It has to be on the branch where people sending PRs are using as base. I guess that it is 2020, right?

> * If so, does that mean we need to update the yaml for the action every year? Or is it easy to have this check PRs on any branch?

Nope. As long as people sending PR will be basing of the branch that has these files.

> * Could I ask you to rebase this onto dev? We've found it easier to manage the history of code changes by keeping a trunk (dev) that is free of any paper commits, and then branching individual conference years (like 2020) from there.

Sure.

deniederhut:
Whoops! Sorry, not sure how this PR got closed

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,20 @@
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [master]
```
mepa, line 6: Is the intention to have these checks run each time an author pushes to their fork's branch? If so then could either
- remove this line since there is no `master` branch and so the spellcheck, etc. run on any branch, or 
- specify exact and/or [pattern-matched](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet) branches.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,20 @@
...
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+    - name: set PY
+      run: echo "::set-env name=PY::$(python --version --version | sha256sum | cut -d' ' -f1)"
```
mepa, line 15: Looks like there's been a change to GitHub Actions since submitting this PR, @ocefpaf.
See [here](https://github.com/mepa/scipy_proceedings/runs/2367138526?check_suite_focus=true)
```
 Error: Unable to process command '::set-env name=PY::63275e7f1b70b809f46598627d8499d8aa7b7f7186de6de089338d1da80504f1' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

We should upgrade to using Environment Files, as suggested.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
...
+  rev: v2.5.0
+  hooks:
+    - id: trailing-whitespace
+    - id: end-of-file-fixer
+    - id: check-added-large-files
+
+- repo: https://github.com/codespell-project/codespell
+  rev: v1.16.0
+  hooks:
+    - id: codespell
```
mepa, line 12: [This](https://github.com/mepa/scipy_proceedings/runs/2367276232?check_suite_focus=true) is the current output of the CI job log when it runs.

So I think we would need some way to have these run not on the entire repo of files. Perhaps via the [config file parameters](https://pre-commit.com/#pre-commit-configyaml---top-level)?

mepa (commented):
Thanks @ocefpaf! I think the automated spell checks along with other pre-commit checks could save time for reviewers. With the changes mentioned and further testing and potential changes to ensure the flow is also convenient for authors, this will be a great addition.

PR #582, opened 2020-07-07
Paper: Edits to HOOMD-blue Paper

b-butler:
A coauthor noticed a potential bug in the usage of another package in one of the code examples. I also take the opportunity to fix some typos that were spotted late. I had to do some minor rewording to keep the paper within the 8 page limit, but no content has changed.


deniederhut (approved):
Cool, this looks fine to me

PR #593, opened 2020-07-13
Slides submission

christine-e-smit:
This is my slides submission, so I'm assuming that the paper submission stuff about building is not relevant.


christine-e-smit:
It looks as though I should have picked a different directory name.

PR #601, opened 2020-07-17
Add Mjolnir talk PDF from SciPy 2020

CAM-Gerlach:
I can also include the LibreOffice ODP as well, though with the enormous size of this repo (> 500 MB) that might not be advisable. 

Has there been consideration of splitting it into repos by year with a separate one for common tools/build system, to make clones, pulls and pushes not take a large amount of time, especially for those without access to high-bandwidth internet at the moment? Thanks!


deniederhut:
> Has there been consideration of splitting it into repos by year with a separate one for common tools/build system

It's split into branches by year, with a separate one for the build system.

deniederhut (approved):

PR #603, opened 2020-08-02
final state of 2020 proceedings

deniederhut:


stargaser (approved):
I viewed all the file changes, and I copy-pasted all the added YouTube and Zenodo URLs into a web browser to check them. I read through the Preface also. Looks good to me! Thanks for doing all this detailed work.

I didn't try building the proceedings -- @deniederhut let me know if I need to do that for this review.

stargaser:
To satisfy my own curiosity -- I see the slides PDFs are included here even when there is a link on Zenodo. The "carpenter" PDF added here is 96 MB so I noticed it! Would linking to the Zenodo URL be enough?

deniederhut:
Great question. The slides aren't in GitHub for archiving, it was just the least bad solution to keeping track of everything.
ivan-marroquin <notifications@github.com> wrote:
“To satisfy my own curiosity -- I see the slides PDFs are included here even when there is a link on Zenodo. The "carpenter" PDF added here is 96 MB so I noticed it! Would linking to the Zenodo URL be enough?”

PR #613, opened 2021-04-30
Collect author ORCIDs for presentations

stargaser:
As a last part of implementing #355, add information to the example JSON files for presentations.


stargaser:
Maybe we'll want to collect author email addresses too?

deniederhut:
This is exactly what I was imagining -- thanks for working on this! As discussed on the call right this very second, we might want to add affiliation as an option in the author field too

stargaser:
I've added the affiliations. For the example of a poster by @deniederhut , I used the affiliation on the poster and not the current one.

deniederhut (approved):
Thanks for doing this work David!

PR #648, opened 2021-07-16
ENH: add note about name changes

deniederhut:
In light of Theresa Tanenbaum's talk today, we should probably have some statement to the effect that our policy is to allow author name changes after publication. I don't know if this is too informal / too buried in an already giant README to be helpful, but it seems better than having nothing.


scoobies:

stargaser (approved):
Looks Good To Me!

mepa (approved):
LGTM

cbcunc (approved):

PR #691, opened 2022-05-24
Auto detect paper name if only one paper.

Carreau:
If there is a single paper in papers (excluding the examples from
stefan).

I _think_ that might make it a bit easier to use

process.

for more information.

*Editor*:  

*Reviewers*:

### DIFF
--- a/make_paper.sh
+++ b/make_paper.sh
@@ -2,10 +2,6 @@
 
 DIR=$1
 
-if [[ ! -d $DIR ]]; then
-  echo "Usage: make_paper.sh source_dir"
-  exit -1
-fi
 
 python publisher/build_paper.py $DIR
 if [ "$?" -ne "0" ]; then
--- a/publisher/build_paper.py
+++ b/publisher/build_paper.py
@@ -12,6 +12,7 @@ import shutil
 import io
 
 from distutils import dir_util
+from pathlib import Path
 
 from writer import writer
 from conf import papers_dir, output_dir, status_file, static_dir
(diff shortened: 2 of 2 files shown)


scoobies:

publisher/build_paper.py
```diff
@@ -224,14 +225,24 @@ def build_paper(paper_id, start=1):
...
 if __name__ == "__main__":
-    if len(sys.argv) != 2:
-        print("Usage: build_paper.py paper_directory")
-        sys.exit(-1)
+    if len(sys.argv) > 2:
+        sys.exit("Usage: build_paper.py paper_directory")
+
+    if len(sys.argv) == 1:
+        pths = list((Path(__file__).parent.parent / "papers").glob("*"))
+        pths = [p for p in pths if p.name not in {"00_bibderwalt", "00_vanderwalt"}]
```
cbcunc, line 233: This is no longer the correct list of papers to exclude. Check .publisher.conf.excludes.

make_paper.sh
```diff
@@ -2,10 +2,6 @@
 
 DIR=$1
 
-if [[ ! -d $DIR ]]; then
```
cbcunc, line 5: Why remove this?

cbcunc (changes requested):

make_paper.sh
Carreau: Because this is now handled on `publisher/build_paper.py` and that's the point of this PR, to not have to pass a folder name, but autodetect when possible.

publisher/build_paper.py
Carreau: Thanks, this PR is a bit over a year old, and I'm not publishing or coming this year. Feel free to reuse the code, but I'm unlikely to work actively on this. 

I'm going to turn this back into a draft. Feel free to reuse the code or close it.

PR #694, opened 2022-05-26
Add booktabs package to LaTeX preamble for formatting tables

ENM185:
This PR intends to update the build process to include the `booktabs` LaTeX package. This is to be used for formatting tables in a readable manner when they would otherwise be convoluted.

### DIFF
--- a/publisher/_static/scipy.sty
+++ b/publisher/_static/scipy.sty
@@ -61,6 +61,7 @@
 }
 
 \usepackage{multirow}
+\usepackage{booktabs}
 \newlength{\tablewidth}
 \setlength{\tablewidth}{\linewidth}
 


scoobies:

deniederhut:
Okay so I rebuilt the 2021 proceedings locally with booktabs included in the style file, and pulled the original 2021 proceedings off the website to compare and they look identical.

*No booktabs*
Screen Shot 2022-06-02 at 8 02 31 AM

*With booktabs*
Screen Shot 2022-06-02 at 8 02 39 AM

This shouldn't have been the case, so I dug around a bit, and deep in the rst->tex translation layer we have the following setting enabled in docutils:

https://github.com/scipy-conference/scipy_proceedings/blob/253b1245c769418b49b0a1f5815b5952bfdeef70/publisher/writer/__init__.py#L62

So it seems like we're doing booktabs-ish translation from the rst tables, but haven't made it possible to have the same behavior for tables written using the raw latex directive. It seems to be that we would want tables that look similar whether they are defined in rst or tex.

@mepa do you agree?

mepa:
@deniederhut I agree about wanting consistent-looking tables whether defined in rst or tex. It seems that adding the booktabs package to the preamble will allow for that, plus give authors flexibility to use booktabs commands.

mepa (approved):

deniederhut (approved):

deniederhut:
Thanks Meghann!

PR #788, opened 2023-04-13
Merge dev into 2023

deniederhut:


mepa (approved):

requirements.txt
```diff
@@ -9,3 +9,4 @@ pytest
 testpath
 ##### requirements with version specifiers
 docutils == 0.14
+pyyaml
```
mepa, line 12: looks good, though perhaps we can also remove the comment lines from requirements.txt

deniederhut: mkay I'll open a separate PR for that

PR #804, opened 2023-05-23
BUG: clear output dir before building

### LINKED ISSUE #797: Cannot rebuild already built latex paper
Hi!

Found a seemingly simple bug in the latex build. To reproduce:

1. Go to the repo root
2. Build the example paper (we assume python reqs and latex are already installed). All goes fine.

        ./make_paper.sh papers/00_texderwalt

3. Build the example paper again and observe the error

        ./make_paper.sh papers/00_texderwalt

Error text:

> Building: 00_texderwalt
> Traceback (most recent call last):
>   File "/home/ubuntu/projects/scipy_proceedings/publisher/build_paper.py", line 679, in <module>
>     build_paper(paper_id)
>   File "/home/ubuntu/projects/scipy_proceedings/publisher/build_paper.py", line 663, in build_paper
>     tex2tex(out_path, metadata_path)
>   File "/home/ubuntu/projects/scipy_proceedings/publisher/build_paper.py", line 532, in tex2tex
>     raise RuntimeError("Found more than one input .tex--not sure which "
> RuntimeError: Found more than one input .tex--not sure which one to use.
> Error building paper papers/00_texderwalt. Aborting.

Expected behavior is that the old paper render is overwritten with the new one. Confusing error, as there is certainly only one tex file in the `papers/00_texderwalt` dir! But turns out the issue is in the `output` directory. It can be worked around by removing the content of the output directory:

    rm -r output/00_texderwalt

But that seems unnecessary and probably easily fixed by the build script.

deniederhut:
When searching for the latex source file, we glob the output dir for anything that ends in '.tex'. When we build the paper, we inject the source file into our template and call the output paper.tex. If you try to build twice, we find both the source and the output ending in tex and raise an error.

Probably the real way to solve this problem is to have users specify the paper's file name in their metadata file. For now, we can clean the output dir before each build to avoid seeing too many .tex files.

Closes #797

### DIFF
--- a/publisher/build_paper.py
+++ b/publisher/build_paper.py
@@ -237,6 +237,8 @@ def detect_paper_type(in_path: str) -> str:
 def prepare_dir(in_path: str, out_path: str, start: int):
     """Copy required files to build dir
     """
+    # clear out whatever cruft might be in the output dir
+    shutil.rmtree(out_path)
     # copy the whole source folder to the build directory
     dir_util.copy_tree(in_path, out_path)
     base_dir = os.path.dirname(__file__)


mepa (approved):
Thanks, @deniederhut. This looks like a good quick fix.

PR #885, opened 2024-04-03
Update 2024 Template (LaTeX)

rowanc1:
Initial template changes for LaTeX, please wait on merging in as this will be an initial test of the new publishing infrastructure!

*Editor*: 

*Reviewers*:


papers/00_texderwalt/myst.yml
```diff
@@ -1,2 +1,43 @@
+version: 1
 project:
+  # Update this to match `scipy-2024-{folder}`
```
rowanc1: Rename folders!

cbcunc (approved):
Thank you, @rowanc1

rowanc1:
We should have a fix for the failing CI later today (i.e. know DOIs can be flagged so as not to raise a failure). I will merge at that point!

PR #908, opened 2024-05-28
Paper: Making Research Data Flow With Python

JBorrow:
If you are creating this PR in order to submit a draft of your paper, please name your PR with `Paper: <title>`. An editor will then add a `paper` label and GitHub Actions will be run to check and build your paper.

_Editor_: Charles Lindsey @cdlindsey

_Reviewers_: 
- Michael Sarahan @msarahan 
- Min Khant Zaw @sharyxx


papers/josh_borrow/paper.md
```diff
@@ -0,0 +1,527 @@
...
+
+In contrast to many other tools, and to the 'HERA Librarian' which preceeded the
+Librarian, files carry minimal science metadata along with them through the
+system. We found that metadata systems attached to data flow and management
+systems often lacked necessary features and caused significant complications
+(such as needing frequent database migrations) when coupled. As such, we chose
+to explicitly delegate metadata management to other subsystems within the
+project, focusing only on file transfers.
+
+#### User Management
```
msarahan: User management seems like an annoying detail, and perhaps not worth including in this paper. It doesn't add to your important ideas of unifying over-the-wire transfers with SneakerNet.

msarahan (commented):
This is a nice overview of a difficult problem. I appreciate the approach of not reinventing the wheel of data transfer, relying on Globus to do things that would be difficult to (re)engineer.

I like your diagrams. I think you might benefit from screenshots, to show what datasets "look" like - either just filenames, or thumbnails, or indexed metadata, or whatever.

I had a few questions, or perhaps maybe feature requests, that might be worth answering somehow, either by ruling them out, or by mentioning as potential new developments:

* Can you sync metadata representations without syncing data? The goal would be to avoid data transfer until necessary. Perhaps you use all data sets, and this on-demand functionality isn't useful.
* How portable is this to other sites? Other research fields? What does data have to look like in order to use your tool?

msarahan:
I'm going on holiday from Thursday, July 4 to Friday, July 12. If there are comments or edits in that time, I'll get back to you when I return.

JBorrow:
Hi @msarahan, we'll address your comments shortly, but have a wonderful holiday!

JBorrow:
Thank you so much for your review! In response to your individual points, I've made a few changes to the paper.

a) I've added an additional description of our data. It's basically raw binary data sat in folders, so having a picture, etc. isn't exactly the most helpful thing in the world. I included one of those in my presentation though!
b) Syncing metadata representations without syncing data is kind of antithetical to the librarian system. Our main goal here is to ensure tight control on the actual data _transfers_, rather than the data itself - because we need to carry stuff down the mountain on disks, our syncs tightly correspond to the data actually being moved. I've tried to highlight this in the paper, but please let me know if you can think of a section where beefing this up may help?
c) Our data management tool isn't super portable to other fields, which we've tried to highlight. It only really makes sense if you need to make use of the SneakerNet functionality as that's our real unique feature. I hope the small additions I've made have clarified what the data that goes into the tool looks like.
d) Re: user management. This is actually somewhat important for the librarian as we want to make sure that the remote sites are as locked down as possible. I've added extra information here as to why our varying permissions levels are important.

Thanks again!

cdlindsey:
Thank you for the good paper.  As I read through, I grew to understand the problem more clearly.  I do think it might be helpful to see some of the python used in the project, so readers can get a better feel for the interface.  Some mocked up example code may be good.

JBorrow:
Thanks very much @cdlindsey; whilst most of the client-facing actions we leave to the documentation, I have added a block showing the major component (uploading files).

msarahan (approved):

cdlindsey:
Screenshot 2024-09-03 122242

PR #933, opened 2024-06-01
Paper: Mandala: Persistent & Queriable Memoization for Easy and Powerful Scientific Data Management

amakelov:
If you are creating this PR in order to submit a draft of your paper, please name your PR with `Paper: <title>`. An editor will then add a `paper` label and GitHub Actions will be run to check and build your paper.

_Editor_:  Amey Ambade @ameyxd 

_Reviewers_:  Andrei Paleyes @apaleyes and Matthew Feickert @matthewfeickert


papers/scipy.tex
```diff
@@ -0,0 +1,295 @@
+\documentclass[11pt,twoside=semi,openright,numbers=noenddot]{scrartcl}
```
rowanc1: Could you please move these documents into a folder under papers, following the latex template provided. There must be a myst.yml with your author/title/affiliations information included as well as a unique ID matching your folder. There is more information in the [readme](https://github.com/scipy-conference/scipy_proceedings?tab=readme-ov-file#setting-up-your-environment) under author instructions.

amakelov: Thanks for the clarification, added this as requested along w/ the new version of the paper

ameyxd:
@amakelov can you recheck the DOIs that fail checks? As long as they are valid DOIs we should be okay. 

you may be able to add citation keys you want to ignore if their DOIs don't exist in myst.yml under error_rules:

```error_rules:
  - rule: doi-exists
    severity: ignore
    keys:
      - abc
      - def01

**Comment by ameyxd** [comment]
Hey @amakelov are you skipping DOIs when they exist? I don't recommend skipping explicitly just because they are failing checks. Please lmk

**Comment by amakelov** [comment]
Hi @ameyxd, I only skipped the DOIs for papers/other publications where I couldn't find a DOI after searching on [crossref](https://search.crossref.org/).

**Comment by ameyxd** [comment]
okay. can you add your ORCiD and affiliation? :)

**Comment by amakelov** [comment]
OK, this failed again - I guess I didn't follow the latex instructions? Is it correct that I must remove everything before and including \begin{document}?

File: papers/aleksandar_makelov/main.tex [hunk 2]
```diff
@@ -0,0 +1,574 @@
...
+
+\def\l{\left}
+\def\r{\right}
+
+\title{\texttt{mandala}: Compositional Memoization for Simple \&
+Powerful Scientific Data Management}
+
+\author{Aleksandar Makelov\\
+\texttt{aleksandar.makelov@gmail.com}
+}
```
fwkoch: Hey @amakelov - Taking a look at your failure, the check is picking up on the `\author` defined here in your `tex` document (which does not have an affiliation), rather than the author defined in `myst.yml`. Can you just delete this author entry here?

```suggestion
```

(This could also be addressed as a bug in the Check, but the suggested change here should resolve it.)

fwkoch:
I think removing everything before and including `\begin{document}` would also be fine - that will remove any other conflicting metadata.

Is there anything in there you are concerned about losing? Some of the `newcommand` stuff might need to come into `myst.yml` `project.math`. There are a number of errors in your preview build, linked in the comment at the top of your PR (e.g. cross-references). Once the affiliation check is passing, I'll take a deeper look there...

matthewfeickert:
:wave: Hi @amakelov. I will be one of your reviewers for your SciPy 2024 proceedings submission. You should expect a full first pass review from me in the form of a PR review with comments and suggestions before 2024-06-29. Please tag me here with any comments you have for me. :+1:

matthewfeickert (changes requested):
@amakelov Please do not edit any files outside of the `papers/aleksandar_makelov/` directory,

https://github.com/scipy-conference/scipy_proceedings/blob/1312327595fe187858f7194fa2ede453b88df2e9/README.md?plain=1#L199

 so please restore all files under `papers/00_tex_template/`.

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+and semantics} (Figures \ref{fig:basic-usage} and \ref{fig:cf}). The integration aims to be
+maximally transparent and unobtrusive, so that the user can focus on the
+\emph{essential complexity} (the scientific problem at hand), rather than on the
+\emph{accidental complexity} (the data management tools used to solve it)
+\citep{Brooks1987NoSB}.
+
+The rest of this paper presents the design and main functionalities of
+\texttt{mandala}, and is organized as follows: 
+\begin{itemize}
+\item In Section \ref{section:core-concepts}, we describe how memoization is
```
matthewfeickert, line 163: @amakelov There are multiple rendering issues on the current Curvenote build, such as Section references (like here), Figure references, and the `basic-usage` figure appearing much larger than expected.

@rowanc1 @fwkoch as this submission is written in LaTeX can you offer any suggestions on syntax changes and how to control figure size in the Curvenote render?

matthewfeickert (changes requested):

apaleyes:
Hi @amakelov! I will also be reviewing your paper, alongside @matthewfeickert . Looking forward to reading it!

papers/aleksandar_makelov/main.tex
amakelov: Sorry about that - was not familiar with the Curvenote system, I thought it's fine to just submit a .tex file that compiles. Is there something I can read that explains the dos and don'ts when writing .tex for Curvenote?

rowanc1: Hi @amakelov there are some instructions [in the readme](https://github.com/scipy-conference/scipy_proceedings?tab=readme-ov-file#instructions-for-authors); there is a video overview there. The system uses `mystmd` (https://mystmd.org) to compile. Curvenote is used for the hosting and checks only. There is some documentation on specifics of LaTeX here:
https://mystmd.org/guide/writing-in-latex

The readme has setup instructions to run/preview locally, including the two templates which you can follow!

For the image sizing, I can aim to take a look next week, I was under the impression that we already supported things like `\includegraphics[width=0.8\linewidth]{figure.png}`, but that might not be implemented.

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+  `interfaces' to their own intermediate results by retracing the memoized
+  calls; (2) all computations in a project form a single computational graph,
+  which can be explored, queried and manipulated in high-level ways through a
+  \emph{computation frame}, which is a natural generalization of a dataframe in
+  a way made precise in this paper. 
+
+  Several features implemented on top of the core memoization data structures --
+  such as natively and transparently handling Python collections, caching of
+  intermediate results, and a flexible versioning system with dynamic dependency
+  tracking -- turn \texttt{mandala} into a practical and radically simpler tool
```
apaleyes: "radically simpler" - simpler than what? Maybe just "simple"?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+  for managing and interacting with computational data.
+\end{abstract}
+
+\section{Introduction}
+\label{section:}
+
+Numerical experiments and simulations are growing into a central part of many
+areas of science and engineering \citep{hey2009fourth}. Recent trends in
+fields such as machine learning point towards ever-increasing complexity of
+computational pipelines, and increasingly safety-critical application domains
```
apaleyes: "increasingly safety-critical application domains" - do you mean these domains are increasingly more safety critical? Or that ML is more pervasive in these domains? Not very clear from how it's phrased

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+such as autonomous driving \citep{bojarski2016end}, healthcare
+\citep{ravi2016deep,abramson2024accurate} and large language model autonomous
```
apaleyes: LLM agents are hardly a safety critical domain in general

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+agents \citep{yang2024swe}.
+
+These developments impose opposing constraints on the tools used to manage the
+resulting computational artifacts: on the one hand, they should be simple and
+easy to use by researchers, with a minimal learning curve and unobtrusive syntax
+and semantics; on the other hand, they should be powerful enough to enable
+high-level operations \cite{wickham2014tidy}, full data \& code provenance
```
apaleyes: cite{wickham2014tidy} -> citep{wickham2014tidy}

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+auditing \citep{davidson2008provenance} and reproducibility
+\cite{ivie2018reproducibility} in complex projects.  Rules and practices that
```
apaleyes: citep again

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+\label{section:core-concepts}
+
+\subsection{Memoization and the Computational Graph}
+
+Memoization is a technique that stores the results of expensive function calls
+to avoid redundant computation. \texttt{mandala} uses \emph{automatic}
+memoization \citep{norvig1991techniques} which is applied via the combination of
+a decorator (\texttt{@op}) and a context manager which specifies the
+\texttt{Storage} object to use (Figure \ref{fig:basic-usage}). The memoization
+can optionally be made persistent, which is what you would typically want in a
```
apaleyes: What is "persistent" referring to? Hard disk storage as opposed to memory?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+long-running project. Any Python function can be memoized (as long as its inputs
+and outputs are serializable by the \texttt{joblib} library); there is no
```
apaleyes: joblib serialisation is actually a very important point, thanks for calling it out. Have you tried `mandala` on some non-trivial datatypes, e.g. numpy/tensorflow/pytorch/jax structures?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+composition of memoized functions that produced a given value -- which must be specified anyway -- is already a canonical reference to it. 
+\item it \textbf{organizes storage activities around a familiar and useful
+interface -- the function call}. This automatically enforces the good practice
+of partitioning code into functions, and eliminates extra `accidental' code to
+save and load values explicitly. Furthermore, it synchronizes failures between
+computation and storage, as the memoized calls are the natural points to
+recover from.
+\end{itemize}
+
+The trade-off is that referring to values without reference to code that
```
apaleyes: There is another trade-off here, I guess. Does modification of the source code invalidate the mandala's storage, and if so by how much?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+functions it calls. This avoids the need for static analysis of the code to find
+dependencies, which can result in many false positives and negatives, especially
+in a dynamic language like Python.
+\end{itemize}
+
+A full discussion of the versioning system is beyond the scope of this paper,
+but we refer the reader to a blog post by the author
+\citep{makelov2023practical} for details and motivation.
+
+\section{Related Work}
```
apaleyes, line 531: This section feels incomplete without discussion of tensorflow and function tracing of the computation graph: https://www.tensorflow.org/guide/function

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+mandala is similar to semver in that it allows you to make backward-compatible
+changes to the interface and logic of dependencies. It is different in that
+versions are still labeled by content, instead of by `non-canonical' numbers.
+The unison programming language \citep{lozano2017unison} represents functions by
+the hash of their content (syntax tree, to be exact).
+
+\section{Conclusion}
+\label{section:}
+
+While \texttt{mandala} is not very mature yet and still very much a work in
```
apaleyes: "not very mature" sounds weakening, there is no reason to leave the reader with an impression like that. How about more positive "is being actively developed"?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+that this paper has given a good overview of the core concepts of
+\texttt{mandala}, and that the reader will be interested in exploring the
+library further.
+
+
+% bibliography
+\bibliography{scipy}
+\bibliographystyle{arxiv_template}
+
+\end{document}
```
apaleyes: SHouldn't these lines appear after the main text is concluded?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+
+
+In this paper, we approach this tension from first principles, and propose to
+use \emph{memoization} of function calls as the basic building block of such a
+system:
+\begin{itemize}
+\item First, we propose a simple design for a decorator for Python functions
```
apaleyes: Readers can argue that [functools](https://docs.python.org/3/library/functools.html) are full of similar function cache decorators. Is it worth spending some time in the paper explaining how mandala's `@op` is different?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+computation frame are the topology of a computational graph, and the `rows' are
+individual computations following that topology (perhaps partially).
+Computation frames allow one to explore the stored computation graph
+interactively and in a way adapted to one's use case. Importantly, thanks to the
+availability of full data provenance in the storage, computation frames can
+perform high-level actions such as `delete all computations that depend on these
+values' with a single line of code.
+\end{itemize}
+
+\section{An Essential vs. Accidental Complexity Analysis}
```
apaleyes: This section serves as a reasonable motivation, and thus shall appear much earlier in the text. In fact, seeing a section that comes after "Conclusion" is a bit confusing

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+of the arguments or return values.
+
+\paragraph{\texttt{Call} and \texttt{Ref} objects, and content/history IDs.}
+\texttt{Ref}s and \texttt{Call}s are the two main data structures in
+\texttt{mandala}'s model of computations. When a call to an
+\texttt{@op}-decorated function \texttt{f} is executed inside a storage context,
+this results in the creation of 
+\begin{itemize}
+\item a \texttt{Ref} object for each input to the call. These wrap the `raw'
+values passed as inputs together with content IDs (hashes of the Python objects)
```
apaleyes: How does the Ref object deal with float point precision as well as randomized inputs? E.g. imagine a classic example of training a neural net: it's common to start with randomized initial weights - would mandala still be helpful in any way in that scenario?

apaleyes (commented):
`mandala` is an interesting library, and I feel the paper does a good job motivating it and explaining some inner workings. The concept of computation frame analogous to database views is very interesting!

The latex build definitely needs fixing. The preview build log is full or errors and warnings, and none of the references (to sections and figures) render correctly, as already pointed out before.

Given that ML is mentioned in the paper as one of the motivating examples, I feel there is a place for showcasing how exactly is the library helpful for ML researchers. Few questions below hint at this.

Finally, it would be great to see examples of the library in action, so that its benefits become very clear and apparent. For example, a quick summary or an excerpt from the logistic regression example, found in the library's repo, could be a valuable addition here.

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
+\documentclass{article} % For LaTeX2e
+\usepackage{arxiv_template}
+
+% Optional math commands from https://github.com/goodfeli/dlbook_notation.
+\input{math_commands.tex}
```
fwkoch: `math_commands.tex` causes many MyST build errors from unsupported `\def` usage.

However, I don't think you are actually using commands from there in your document, so the build errors do not lead to errors in your paper. Maybe you can just remove this line?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,571 @@
...
+hash of the identity of \texttt{f} and the content IDs of the input
+\texttt{Ref}s) and a history ID (by analogy, a hash of the identity of
+\texttt{f} and the history IDs of the inputs).
+\item a \texttt{Ref} object for each output of the call. These are again
+assigned content IDs by value, and history IDs by hashing the tuple (history ID
+of the call, corresponding output name\footnote{Since Python functions don't
+have designated output names, we instead generate output names automatically using the order in the tuple of return values.}).
+\end{itemize}
+
+\begin{wrapfigure}[23]{r}{0.5\textwidth}
```
fwkoch: MyST currently does not support `wrapfigure` so this figure and the subsequent figure that uses `wrapfigure` do not render.

For now, can you just replace these with normal `figure`:

```suggestion
\begin{figure}
```
etc.

papers/aleksandar_makelov/myst.yml
```diff
@@ -0,0 +1,44 @@
...
+        - semver
+        - lozano2017unison
+  # A banner will be generated for you on publication, this is a placeholder
+  banner: banner.png
+  # The rest of the information shouldn't be modified
+  subject: Research Article
+  open_access: true
+  license: CC-BY-4.0
+  venue: Scipy 2024
+  date: 2024-07-10
```
fwkoch, line 42: Regarding the broken section references in your MyST site that render as "??" - These are `\ref` commands looking for a section number, but your sections are not numbered.

Probably the easiest fix for this is to add to your `myst.yml`
```suggestion
  date: 2024-07-10
  numbering:
    headings: true
```

fwkoch (commented):
Here are some suggestions to fix your two missing figures and your unresolved cross-references. I believe if you make these changes, your content will be building correctly.

Currently, MyST does not support some of the subfigure sizing/arrangement from your tex file: https://github.com/executablebooks/mystmd/issues/1337

ameyxd:
Review reminders sent to @matthewfeickert and @apaleyes

papers/aleksandar_makelov/main.tex
fwkoch: I updated MyST to support this, so you may leave `wrapfigure` in your manuscript.

papers/aleksandar_makelov/myst.yml
fwkoch: (Numberings will also be fixed on the final scipy proceedings site, but adding the change I suggested will fix your preview build.)

papers/aleksandar_makelov/main.tex
amakelov: Wow, thanks a lot!

papers/aleksandar_makelov/main.tex
amakelov: I use it for all my ML work, which involves objects like `torch.Tensor`, `pandas.DataFrame`, as well as custom classes, e.g. `torch.nn.Module`. Have not used it with `jax` or `tensorflow` though.

My observations are the following:
- indeed it can be tricky with some types, for example, if you save a `torch.Tensor` with `joblib.dump` and then load it back with `joblib.load`, the `joblib.hash` of the object will be different (might be strictly a CUDA issue, haven't checked yet)
    - I use a workaround: when the given object is a tensor, or a (nested) Python collection of tensors, I first convert to `numpy.ndarray`, which does not suffer from this issue (I believe `joblib` is *specifically* designed to play nice with `numpy`)
    - But even then, custom classes that hide `Tensor`s inside are an issue (happens all the time with `nn.Module`). You could use some more heavyweight introspection to try to overcome this, but I haven't done it so far
- For a lot of other things it works pretty great though. E.g., a lot of ML analyses are just passing arrays around.

To get an idea, here's a content hash function from "the trenches":
```python
def get_content_hash(obj: Any) -> str:
    if hasattr(obj, "__get_mandala_dict__"):
        obj = obj.__get_mandala_dict__()
    if Config.has_torch:
        # TODO: ideally, should add a label to distinguish this from a numpy
        # array with the same contents!
        obj = tensor_to_numpy(obj) 
    if isinstance(obj, pd.DataFrame):
        # DataFrames cause collisions for joblib hashing for some reason
        # TODO: the below may be incomplete
        obj = {
            "columns": obj.columns,
            "values": obj.values,
            "index": obj.index,
        }
    result = joblib.hash(obj)  # this hash is canonical wrt python collections
    if result is None:
        raise RuntimeError("joblib.hash returned None")
    return result
```

papers/aleksandar_makelov/main.tex
amakelov: Great question. There's a versioning system (I hope to write something more informative about it in the paper during the review period) which is turned off by default. This means that you can change the code, and the storage won't be invalidated. 

If you choose to use the versioning system (which admittedly is experimental, and has some kind-of-out-there ideas), it works roughly as follows:
- The overall idea is that, for each memoized function, we store its code, as well as the code of its dependencies, and when a change is detected (in either the memoized function or any dependency), the user has the option to mark the change as a "cosmetic" one (so that it doesn't invalidate saved calls) or a "semantic" one (in which case, when using this new code, memoized calls that used the old code will be recomputed).
- Crucially, the dependencies are tracked **dynamically**, meaning that each individual call keeps track of the functions it called (via a custom call stack in the `mandala` internals that does not invoke `sys.settrace`). This is nice because if you change some dependency, you want only the calls that actually accessed it to be invalidated.
    - The "version" of a memoized function `f` for a call is then a combination of `f`'s code at the time of the call, plus the code of all dependencies. 
- The system is **truth-of-the-code** [in the terminology of Rich Hickey](https://youtu.be/oyLBGkS5ICk?feature=shared&t=2928) like `git`, meaning that
    - it's content-based: there's no human annotation of versions, the only exception being the ability to mark some changes as non-invalidating (which brings some subtle problems).
    - the storage can figure out which "world" it is in w.r.t. versions by just looking at the state of the code at the time of calling the memoized function. This is a really powerful property! You can just "travel in time" to a previous code state, and interact with the storage at that point in time. 
        - It works because when you see the inputs to a call to `f`, you can find what dependencies `f` used in the past on these inputs. If one of the memoized calls matches the current state of the code w.r.t. these dependencies, you can re-use the memoized call. If none of the calls match, you have to re-compute. The key property is that, **assuming the set of dependencies invoked during a given call is deterministic**, there can't be two or more saved calls that both match the current state of the code, which allows you to uniquely resolve the version to use (and hence the memoized call whose outputs to return). 

The ability to mark some changes as non-invalidating opens the door to some subtle issues. Suppose we have a memoized function `f`. We extract a method `g` out of `f`, refactor `f` to
use `g`, and then mark the change as non-invalidating. However, the storage doesn't know that `g` is now effectively a dependency of `f`'s past calls . Then, if we change `g` and call `f` on some previously seen inputs, the storage will think it's fine to reuse the memoized call. 

I currently have no good solution to this that doesn't require some manual work by the user.

papers/aleksandar_makelov/main.tex
amakelov: Yes, persistent = hard disk storage, by default `Storage()` is in-memory.

papers/aleksandar_makelov/main.tex
amakelov: Thanks for the pointer - what do you think should be said about `@tf.function` (I'm not familiar with it btw)? The tracing system used in `mandala` is a very simple one, based on a decorator for functions/methods that manipulates an internal call stack. Do you mean that `@tf.function` could be a natural point of comparison because it's very similar?

papers/aleksandar_makelov/main.tex
amakelov: Oops yes fixed!

papers/aleksandar_makelov/main.tex
amakelov: At present I removed the section (it was scrapped after `\end{document}`, but I guess it still showed up in the preview); but if you found some things from it particularly valuable I'm happy to incorporate them in the paper!

papers/aleksandar_makelov/main.tex
amakelov: Regarding precision: great question, haven't really thought about that much! Currently there is no dedicated logic for that. When a memoized function is called on a value, the value is (roughly speaking) passed to `joblib.hash`, and the result is the "content ID". I think this is not entirely impractical (e.g. I haven't ran into issues):
- most times, inputs are either very specific scalar (or otherwise "small" and "explicit") values humans provide intentionally as some fixed expression (e.g., `True`, `[1, 2, 3]`, `0.75`), or more complex values that are results of other memoized functions - where it's very unlikely to run into two values that differ on the level of floating point error
- however, there are still some exceptions. For example, suppose two calls return arrays of zeros of the same shape, except there are some rounding errors. We want to treat the two outputs as the same object, which currently won't work.

Regarding randomness: this may be somewhat opinionated, but to me the best practice when dealing with randomness in projects where you actually want to save and be able to reproduce results is to 
- encapsulate randomness inside the memoized functions, 
- and ideally give these functions some arguments that control the randomness (easiest one would be the appropriate variant of `random_seed` to apply inside the function). This way you get to control and track your randomness precisely
- even without these random seed arguments, you get the property that the first time you call a memoized function, it samples whatever randomness it needs, and after that you reuse that same sample.

For the neural net example specifically, you'd typically have some function like `init_model(...)` that inits your model using some "deterministic" human-readable parametrization (e.g., this many layers, such and such architecture, and `random_seed=...`). You can return the model if you want to save and examine the randomly initialized weights later on.

papers/aleksandar_makelov/main.tex
amakelov: Thanks, fixed

papers/aleksandar_makelov/main.tex
amakelov: The latter; it's been rephrased now.

papers/aleksandar_makelov/main.tex
amakelov: Haha yes true, at least for now :) removed.

papers/aleksandar_makelov/main.tex
amakelov: Thanks, replaced!

papers/aleksandar_makelov/main.tex
amakelov: Removed, thanks!

papers/aleksandar_makelov/myst.yml
amakelov: So after all the suggested changes, none of the figures are showing for me when I preview locally. I get the following type of errors:
```
convert-im6.q16: attempt to perform an operation not allowed by the security policy `PDF' @ error/constitute.c/IsCoderAuthorized/421.
convert-im6.q16: no images defined `_build/site/public/fig1-a9a37853ebabe041933dc9811dd4708b.png' @ error/convert.c/ConvertImageCommand/3229.
```
```
⛔️ main.tex:413:5 Cannot convert image "list-46402ef1b912376ea7cea258654f06ad.pdf" - may not correctly render.
   To convert this image, you must install imagemagick
⛔️ _build/site/public/fig5-12a209743562173f048a769ce611b87d.pdf Could not convert from PDF to PNG: _build/site/public/fig5-12a209743562173f048a769ce611b87d.pdf - Error: Command failed: convert -density 600 -colorspace RGB _build/site/public/fig5-12a209743562173f048a769ce611b87d.pdf -trim _build/site/public/fig5-12a209743562173f048a769ce611b87d.png
convert-im6.q16: attempt to perform an operation not allowed by the security policy `PDF' @ error/constitute.c/IsCoderAuthorized/421.
convert-im6.q16: no images defined `_build/site/public/fig5-12a209743562173f048a769ce611b87d.png' @ error/convert.c/ConvertImageCommand/3229.
```
```
⚠️  _build/site/public/fig3-340c6c379bf85d1626441d78f8180d55.pdf Could not convert to webp:
Error: Command failed: convert -density 600 -colorspace RGB _build/site/public/fig3-340c6c379bf85d1626441d78f8180d55.pdf -trim _build/site/public/fig3-340c6c379bf85d1626441d78f8180d55.png
convert-im6.q16: attempt to perform an operation not allowed by the security policy `PDF' @ error/constitute.c/IsCoderAuthorized/421.
convert-im6.q16: no images defined `_build/site/public/fig3-340c6c379bf85d1626441d78f8180d55.png' @ error/convert.c/ConvertImageCommand/3229.
```

papers/aleksandar_makelov/main.tex
amakelov: Actually, I've been doing some tests today, and there are indeed some problems around `joblib` hashing and serialization. For instance, consider the code below which trains a `RandomForestClassifier` and then makes a roundtrip through `joblib` serialization:

```python
import joblib
from sklearn.ensemble import RandomForestClassifier
from sklearn.datasets import load_digits
import random
import numpy as np
import io
from typing import Any

X, y = load_digits(n_class=10, return_X_y=True)
np.random.seed(42)
random.seed(42)

def serialize(obj: Any) -> bytes:
    buffer = io.BytesIO()
    joblib.dump(obj, buffer)
    return buffer.getvalue()


def deserialize(value: bytes) -> Any:
    buffer = io.BytesIO(value)
    return joblib.load(buffer)

X, y = load_digits(n_class=10, return_X_y=True)
model = RandomForestClassifier(max_depth=2, n_estimators=100, random_state=42).fit(X, y)
restored = deserialize(serialize(model))
joblib.hash(model), joblib.hash(restored)
```
You can see that not only are the two hashes (original and post-roundtrip) different, but they are non-deterministic (even after setting all the random seeds). Probably there's some memory address information in the random forest implementation that causes this.

This may not be a huge problem in practice, as long as you don't pass "pathological" objects like that as input to `@op`s - unless you computed these objects by another `@op` (using arguments without pathological behavior), in which case they are stored once and deserialized every time from the same stored representation. The issue might come up when you pass what is morally "the same" object as input to two different calls, but the hashing does not recognize that and recomputes the call.

apaleyes: Thanks @amakelov ! I really appreciate the way you dug into the library, and the elaborate response.

Something general I'd like to discuss here though. Most questions I've asked in this review are an opportunity to clarify stuff in the paper. For example this one - it clearly suggests paper could use "limitations" section to talk about serialization, and about code modification (below). Or the question above about storage really hints at rephrasing in the main text.

Are you planning to address these questions in the paper, rather than just stopping at (very detailed and elaborate) answers in the PR itself?

papers/aleksandar_makelov/main.tex
apaleyes: Yes, it is a very natural mechanism to compare to. Especially in ML context, where one would expect a lot of people to be aware of it.

papers/aleksandar_makelov/main.tex
apaleyes: It's totally fine to just remove it, thanks

papers/aleksandar_makelov/main.tex
amakelov: Yes, certainly - once the discussion here is resolved, based on the consensus and feedback I'll update the paper and post a summary of the changes here.

papers/aleksandar_makelov/main.tex
apaleyes: ok, maybe worth clarifying this in the paper too

papers/aleksandar_makelov/main.tex
apaleyes: perfect, thanks! As mentioned above, I think your findings here warrant a "known limitations" section. Or "future work and improvements", to sound a bit more positive. Hopefully the idea is clear - to make sure paper presents the library in a fair way, and lets the potential users know what the limitations are.

papers/aleksandar_makelov/main.tex
apaleyes: > I hope to write something more informative about it in the paper during the review period

so here's the golden chance to do it! this sounds both like something to add to the section about main mechanics, and to the "limitations" suggested above

papers/aleksandar_makelov/main.tex
apaleyes: Saving and reproducing results is important, no doubt. But the entire point of randomized weights is to make sure results are not dependent on a random draw. Yes, fixing seeds is a common practice, but if some method does not work without the seed fixed - it just does not work. I thus think randomized inputs is quite an important aspect of ML research. It's totally fine if the library has no great answer for it at the moment. Again, this is why i suggested "limitations" section.

On precision - perhaps there is time left to run some tests and report results in the paper? It may either reveal that library can handle these precision differences just fine, or that it cannot and then it's a future improvement that users need to be aware of. Who knows, maybe that'll lead to some library improvement even before the paper is accepted to the proceedings!

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+\label{section:core-concepts}
+
+\subsection{Memoization and the Computational Graph}
+
+Memoization is a technique that stores the results of expensive function calls
+to avoid redundant computation. \texttt{mandala} uses \emph{automatic}
+memoization \citep{norvig1991techniques} which is applied via the combination of
+a decorator (\texttt{@op}) and a context manager which specifies the
+\texttt{Storage} object to use (Figure \ref{fig:basic-usage}). The memoization
+can optionally be made persistent, which is what you would typically want in a
```
matthewfeickert:
```suggestion
can optionally be made persistent to disk, which is what you would typically want in a
```

I agree that would be a helpful clarification. :+1:

papers/aleksandar_makelov/main.tex
matthewfeickert: Thanks very much @amakelov and @apaleyes for already having this discussion. I think it encapsulates the questions that I had in my review quite well, so I would agree with @apaleyes that some including some context on difficult serialization cases could be quite useful. It doesn't have to be in this section, but discussing it in the paper would be good.

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+\end{wrapfigure}
+
+
+The \texttt{Ref}s and the \texttt{Call} are then stored in the storage backend,
+and the next time \texttt{f} is called on inputs that have the same
+\emph{content} IDs, the stored \texttt{Call} is looked up to find the output
+\texttt{Ref}s, which are then returned (possibly with properly updated history
+IDs, if the call exists in storage by content ID only). The combination of all
+stored \texttt{Call}s and \texttt{Ref}s across memoized functions form the
+\textbf{computational graph} represented by the storage (Figure \ref{fig:comp-graph}). Importantly, the
```
matthewfeickert: You may discuss this elsewhere more, but is it possible to export the computational graph and inspect it, similar to how you might be able to optimize and visualize task graphs in Dask?

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+incremental computation and recovery from failures. Here, we add more
+computations while transparently reusing already computed values.}
+\label{fig:basic-usage}
+\end{figure}
+
+In this paper we present \texttt{mandala}, our proposal for such a system. At
+its core, it \textbf{integrates best data management practices} -- such as full
+data provenance tracking, idempotent \& deterministic computation,
+content-addressable versioning of code and its dependencies, and declarative
+high-level data manipulation -- \textbf{into Python's already familiar syntax
```
matthewfeickert:
```suggestion
its core, it \textbf{integrates best data management practices} --- such as full
data provenance tracking, idempotent \& deterministic computation,
content-addressable versioning of code and its dependencies, and declarative
high-level data manipulation --- \textbf{into Python's already familiar syntax
```

In LaTeX em-dashes are made by `---`. `--` gets you an en-dash, but for the punctuation you're using you want an em-dash. This applies throughout the whole document, so following here I'll denote this same comment by just commending "em-dash". :+1:

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+pointing to the value to be saved and the address where it should be saved
+(whether this is some kind of name or a file path). Basing data management on
+memoization means that the `address' of a value is now implicit in the code that
+produced it, and the value itself is stored in a shared storage backend. This has several advantages:
+\begin{itemize}
+\item it \textbf{eliminates the need to manually name artifacts}. This
+eliminates a major source of accidental complexity: names are arbitrary,
+ambiguous, and can drift away from the actual content of the value they point to
+over time. On the other hand, names are not strictly necessary, because the
+composition of memoized functions that produced a given value -- which must be specified anyway -- is already a canonical reference to it.
```
matthewfeickert:
```suggestion
composition of memoized functions that produced a given value --- which must be specified anyway --- is already a canonical reference to it.
```

em-dash

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+memoization means that the `address' of a value is now implicit in the code that
+produced it, and the value itself is stored in a shared storage backend. This has several advantages:
+\begin{itemize}
+\item it \textbf{eliminates the need to manually name artifacts}. This
+eliminates a major source of accidental complexity: names are arbitrary,
+ambiguous, and can drift away from the actual content of the value they point to
+over time. On the other hand, names are not strictly necessary, because the
+composition of memoized functions that produced a given value -- which must be specified anyway -- is already a canonical reference to it. 
+\item it \textbf{organizes storage activities around a familiar and useful
+interface -- the function call}. This automatically enforces the good practice
```
matthewfeickert:
```suggestion
interface: the function call}. This automatically enforces the good practice
```

I think you might actually want a colon here? Though an em-dash would work as well.

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+computation graph, and we're excited to explore their full potential in future
+work.
+
+\section{Some Extra Features}
+\label{section:extra-features}
+
+\subsection{Data Structures}
+\label{subsection:data-structures}
+
+Python's native collections -- lists, dicts, sets -- can be memoized
```
matthewfeickert:
```suggestion
Python's native collections --- lists, dicts, sets --- can be memoized
```

em-dash

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+  `interfaces' to their own intermediate results by retracing the memoized
+  calls; (2) all computations in a project form a single computational graph,
+  which can be explored, queried and manipulated in high-level ways through a
+  \emph{computation frame}, which is a natural generalization of a dataframe in
+  a way made precise in this paper. 
+
+  Several features implemented on top of the core memoization data structures --
+  such as natively and transparently handling Python collections, caching of
+  intermediate results, and a flexible versioning system with dynamic dependency
+  tracking -- turn \texttt{mandala} into a practical and simple tool for
```
matthewfeickert:
```suggestion
  Several features implemented on top of the core memoization data structures ---
  such as natively and transparently handling Python collections, caching of
  intermediate results, and a flexible versioning system with dynamic dependency
  tracking --- turn \texttt{mandala} into a practical and simple tool for
```

em-dash

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+\subsection{Retracing as a Versatile Imperative Interface to the Stored Computation Graph}
+\label{subsection:retracing}
+
+The compositional nature of memoization makes it possible to build complex
+computations out of calls to memoized functions, turning the entire computation
+into an end-to-end-memoized interface to its own intermediate results. The main
+way to interact with such a computation is through \textbf{retracing}, which
+means stepping through memoized code with the purpose of resuming from a
+failure, loading intermediate values, or continuing from a particular point with
+new computations. A small example of retracing is shown in Figure
```
matthewfeickert: This is more of a general question than anything that you need to address in the paper, but this seems somewhat similar to checkpointing in automatic differentiation. Is this a fair comparison? (I'm asking in earnest, as I'm not a computer scientist).

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,509 @@
...
+\newtheorem{theorem}{Theorem}[section]
+\newtheorem{lemma}[theorem]{Lemma}
+
+\begin{document}
+\maketitle
+
+
+\begin{abstract}
+  We present
+  \texttt{mandala}\footnote{\url{https://github.com/amakelov/mandala}}, a Python
```
matthewfeickert, line 64: Instead of only giving the GitHub URL for the project, if in https://github.com/amakelov/mandala you add a [`CITATION.cff` file](https://citation-file-format.github.io/) and optionally hook up the project repo to a DOI minting archive service like [Zenodo](https://zenodo.org/) then you could fully cite `mandala` here in the paper. :)

To be clear, I don't think this is something that is a problem for the paper as is, but just a suggestion as I have been thinking on software citation more in recent years: https://arxiv.org/abs/2309.14571

matthewfeickert (commented):
## @matthewfeickert Initial review

@amakelov, thanks for the nice proceedings submission! I overall agree with @apaleyes's review and have added some small suggested changes and asked some questions of my own.

Overall, I think this is a very strong proceedings submission that does an excellent job of balancing technical detail with accessibility to a broader scientific audience! I enjoyed reading it and I look forward to discussing any questions on the review as is. :+1:

papers/aleksandar_makelov/main.tex
amakelov: Indeed, it is analogous to checkpointing in the sense that you choose the "boundaries" at which results are checkpointed by the way you define the functions decorated by `@op`. 

As a general guide, you should make something an `@op` if you want to save its
outputs, e.g. if they take a long time to compute but you need them for later
analysis. Since `@op` encourages composition, you should aim to have `@op`s work on the outputs of other `@op`s, or on the collections and/or items of outputs of other `@op`s. There's a bit of WIP documentation on collections [here](https://amakelov.github.io/mandala/05_collections/)

papers/aleksandar_makelov/main.tex
amakelov: **Regarding randomness**: I may be opinionated about this, but I think that randomization should not really be automated by the storage layer, because it expresses some non-trivial (and sometimes crucial in a subtle way) intent on part of the user. By enforcing the explicit use of randomness as a resource (if you want multiple random samples, you should parametrize this somehow as an argument to the `@op`), I think `mandala` actually encourages best practices. 

For example, it will be quite visible, looking at memoized code, whether multiple seeds were tested or only a single random instance was implicitly created when there was no random seed argument explicitly passed into the memoized call.

**Regarding precision**: This is a great point, currently the library is very sensitive to often negligible differences between numerical types:
```python
from mandala.utils import get_content_hash

print(get_content_hash(42))
print(get_content_hash(42.0))
print(get_content_hash(42.00000000001))
```
```
d922f805b5eead8c40ee21f14329d6c7
ca276c58eef17e13c4f274c9280abc1e
b61bb24b62bf6b1ab95506a62843be08
```
This is not insurmountable in principle, but it could be hard to have an automatic solution that always works:
- The easy case is when handling "simple" objects like numbers. There could be a custom numerical type, `RoundedFloat` or whatever, which is insensitive to precision differences by passing to a canonical representation before hashing. If an input/output to a function is annotated with `RoundedFloat`, the storage will save only this canonical representation. 
- A similar solution could apply to specific array types, like `np.ndarray`.
- Though I imagine it'll get out of control if more complex types are to be covered, which is the real issue. If someone gives you two huge ML models that differ by 1e-19 in one of their parameters, it seems reasonable to treat them as practically the same. But also, such a situation is unlikely to arise in most practical scenarios, where you'll be handling models that differ in much more significant ways. 
- Finally, if there's a use case where one really cares about this kind of issue, it's probably best to write a custom hash function for the type in question.

I'll certainly add a mention of this in the paper!

papers/aleksandar_makelov/main.tex
amakelov: Added a discussion of these topics in "Limitations"

papers/aleksandar_makelov/main.tex
amakelov: Following our discussion, I extended the discussion of the versioning system in the main body, and added to the limitations section.

papers/aleksandar_makelov/main.tex
amakelov: There's now a figure illustrating this (Figure 2(c)) in the paper

papers/aleksandar_makelov/main.tex
amakelov: Great point, added to related work!

apaleyes:
@amakelov looks like you've put quite a lot of work into updating the paper. Thanks! Do you think it is ready for another pass, or not yet?

amakelov:
> @amakelov looks like you've put quite a lot of work into updating the paper. Thanks! Do you think it is ready for another pass, or not yet?

@apaleyes indeed, the current version incorporates all the feedback as far as I'm aware - so it's a great time for another pass! 

In the meantime, in case anybody is interested, I've also greatly updated the documentation of the library in preparation for a next release:
- readme: https://github.com/amakelov/mandala
- docs: https://amakelov.github.io/mandala/

In particular, the "Caveats" notebook contains some of the things we discussed here: https://amakelov.github.io/mandala/tutorials/gotchas/

ameyxd:
@apaleyes, @matthewfeickert - can you confirm your feedback has been incorporated by the author and give me a thumbs up.

papers/aleksandar_makelov/main.tex
```diff
@@ -0,0 +1,641 @@
...
+\begin{itemize}
+\item Representing Python collections in a way transparent to the storage, so
+that the membership relationships between a collection and its items are
+propagated through the saved computational graph;
+\item Caching of intermediate results to speed up retracing and memoization;
+\item A flexible versioning system with automatic dynamic dependency tracking.
+\end{itemize}
+\end{itemize}
+
+Finally, we give an overview of related work in Section \ref{section:related-work}.
```
apaleyes, line 186: Not final anymore, with "limitations" now in place!

apaleyes:
@ameyxd yes, I am happy with the changes. Great job on the paper @amakelov , thanks for the insightful discussions!

matthewfeickert (approved):
@ameyxd Sorry for the delayed final review, but all LGTM. Thanks for the nice work here @amakelov. :+1:

matthewfeickert:
Congratulations @amakelov on having your paper accepted and merged! :rocket: Nice work. :)

PR #1022, opened 2024-10-01
🎓 Add dois and shared journal-level myst config

fwkoch:
This also did some auto-formatting on the yaml!


rowanc1:
Taken over by #1029.

PR #1032, opened 2024-10-25
Paper: add acknowledgements to CyFi paper

ejm714:
Apologies for the late addition, but we're hoping to add an acknowledgements section to document funding sources, thank the data providers, and provide direction on email correspondence.

cc: @rowanc1


rowanc1:
Thanks @ejm714!

PR #1139, opened 2025-09-11
Poster: Carmina

infinite0x20:
Added the poster and myst.yml file for Carmina


fwkoch:
Thank you @infinite0x20!

PR #1151, opened 2025-09-20
fix: Fix typo in path to parent proceedings.

mihaimaruseac:
In the given template for submitting slides, etc. to the proceedings, the path under

```yaml
extends: ../../proceedings.yml
```

should actually be `../../../proceedings.yml` since the slides would be in a directory following the `presentations/slides/<name>` pattern, so there are 3 directories to go up to reach the `proceedings.yml` file at the root of the directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,7 +525,7 @@ Editors should come to a final 'ready', 'unready' decision before the **Final Ed
 
 ```yaml
 version: 1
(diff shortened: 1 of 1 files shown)


fwkoch:
Thanks for catching this - will make a difference next year. 🚀

PR #1153, opened 2025-09-22
Slides: Python is all you need: an overview of the composable, Python-native data stack

deepyaman:


fwkoch:
Thank you @deepyaman!

PR #1166, opened 2025-09-26
Poster: Fast and scalable general geospatial regridding

keewis:
This adds my poster to the proceedings.


fwkoch:
Thanks again @keewis!

PR #1206, opened 2026-06-01
Paper: Derivations, Not Just Simulations: Teaching Applied Mathematics with Scientific Python

mzargham:
This PR submits *"Derivations, Not Just Simulations: Teaching Applied Mathematics with Scientific Python"* by Michael Zargham (Dynamical Systems Group) for the SciPy 2026 Proceedings.

The article is the Jupyter notebook `papers/michael_zargham/paper.ipynb`, configured as MyST's primary export source (the same notebook-primary architecture as Kadie's 2025 SciPy paper on solvable and unsolvable equations). The paper presents a SymPy-driven port of Isaacs's 1951 *homicidal chauffeur* differential game as a worked exemplar for teaching applied mathematics with the SciPy stack, then integrates the optimal feedback backward with `scipy.integrate.solve_ivp` and renders the value function V*(x) as a matplotlib heat map.

`papers/michael_zargham/README.md` documents the architectural choices and a 3-line reviewer recipe: `uv sync; uv run pytest tests/; uv run jupyter lab paper.ipynb`. All 35 tests pass locally; the notebook re-executes cleanly via `jupyter nbconvert --execute`. Reproducibility is pinned via `uv.lock`.

Generative-AI tool use is disclosed in the paper's Acknowledgments cell per `genai_policy.md` (upstream commit 02399182).

@scipy-conference/2026-proceedings — please apply the `draft` label so the build checks fire.


mzargham:
Screenshot 2026-06-01 at 2 21 08 AM
I was able to test out the build locally with MyST but it appears the `paper.ipynb` based submission doesn't trigger the Curvenote Preview, per the above. Let me know what if anything else I need to do to have this work considered.

rowanc1:
@mzargham it was missing the PR label to trigger the workflow!

Can you run `myst init --write-toc` and order the toc in your `myst.yml` so that the ipynb paper comes first? The other one (I think) can be removed.

The references and acknowledgement sections can remove the `{.unnumbered}` which isn't picked up. These blocks will be pulled to be bottom by default anyways!

mzargham:
@rowanc1 thanks! looks like i have to run down the build issue but i'll work on it soon(tm); probably over the weekend.

mzargham:
I managed to get it compiling, i got my local environment matched to the one used for the tests and then it was relatively smooth sailing to green checks. That said, i am now seeing some formatting weirdness in the pdf, which seems to come from the way the sympy generated latex gets rendered. I'll continue to iterate for clarity and aesthetics but i think it's in a place now where it earns the attention of reviewers. (thanks again for your help)

chongshenng:
Hello @mzargham, I am Chong Shen Ng (@chongshenng). I will serve as the editor for your submission. Please reach out to me if any assistance needed 😊.

ktzhao:
Hello! Just leaving a quick comment to confirm I'll be reviewing this submission. It's an honor to help out, and I'm really looking forward to reading your work.

mzargham:
Great to meet you @ktzhao and @chongshenng. this week I am preparing the talk. if you want to discuss next week at SciPy in Minnesota let me know. I really like this approach to publishing via jupyter/myst/curvenote but i am new to it. I love the power of have the math and code directly available but i'll admit i was a little unsure about content balancing between literature review, narrating **why, what, how** in prose and providing the code and visuals. Thanks in advance for your feedback!

chongshenng:
@mzargham Glad you like it! I won't be attending SciPy this year unfortunately, perhaps next time.

ktzhao:
Hi @mzargham, nice to meet you as well! Unfortunately, I won't be attending SciPy in person this year, but I wish you the best with your presentation.

JuanCab:
I can review this paper...

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+    "  transformations, optimality criteria, and invariances live as\n",
+    "  symbol-manipulation under formal rules \u2014 typically stays on paper.\n",
+    "  SymPy is the SciPy-ecosystem tool that brings it into the same\n",
+    "  execution environment, and once all three worlds are executable in\n",
+    "  one notebook the glass-box principle (Buchberger; Drijvers; Artigue)\n",
+    "  becomes practicable for working scientists.\n",
+    "\n",
+    "  The worked exemplar is Isaacs\\'s 1951 homicidal-chauffeur differential\n",
+    "  game. *Naive heuristics catch the evader in 11.3 seconds; the\n",
+    "  saddle-point optimal play catches in 12.1 seconds, but its evader\n",
```
JuanCab, line 25: I am bothered by 'timing' without a reference to the system used. As I read the paper I realized it meant elapsed time in the game based on the specific speeds and parameters used in the example.  It might make sense to state that "given the exemplar's settings" or to simply drop the timings and make reference to the fact that the optimal play does manage to evade longer.  The exact times are not really relevant to this thesis.

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+    "\n",
+    "SymPy closes that gap. The central claim of this paper is that with SymPy in the notebook alongside NumPy/SciPy and matplotlib, the *glass-box principle* [@buchberger1990] (refined by Drijvers and Artigue [@artigue2002; @tall1991amtcomputer]) finally becomes practicable: the reader can read the seam between a derived expression, the function it lambdifies into, and the figure it renders \u2014 and verify the whole chain, rather than trust one opaque step. SymPy is what unlocks the third world for scientific-computing practice. Proprietary computer-algebra systems (Maple, Mathematica) and other open-source ones (Maxima, Sage) have offered symbolic capabilities for decades; the novel pivot here is having *all three Tall worlds executable in the same open-source Python kernel and notebook*, alongside the NumPy/SciPy/matplotlib idiom the working reader already uses.\n",
```
JuanCab, line 65: "read the seam" threw me for a bit.  In part because three items should have more than one "seam"  Does "read the seams" work?  Also, most seams as far as I know don't get "read."  Is this some mathematics-specific verbiage that should get re-expressed for a broader audience?

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+    "\n",
+    "That \u2605 is the question the rest of the paper answers. The second figure (`naive_vs_optimal`) compares two heuristic evaders against the same heuristic pursuer at the same initial condition: a \"run straight away from the pursuer\" evader (left, captured at 11.3 seconds) and a \"run perpendicular to the closing direction\" evader (right, captured at 6.2 seconds). The saddle-point evader plays differently \u2014 captured at 12.1 seconds, with the \u2605 kink. The saddle-point theorem says no heuristic evader can do better than the optimal strategy against an optimal pursuer; the 12.1 seconds value is a *bound*, not a behavioural guess. **The question:** where does the \u2605 come from? Why does the evader's optimal path kink at that specific point? Sections 4 through 8 derive the answer; \u00a73 is the puzzle the rest of the paper earns.\n"
```
JuanCab, line 99: I am confused a bit by the phrasing "the puzzle the rest of the paper earns", how does the paper earn the puzzle?  I could see the goal of the paper being we can earn the use of sympy to help us solve the puzzle, but that is earning the right to the tool, not the puzzle.

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+    "\n",
+    "Two concepts from optimal control are load-bearing for the derivations that follow. We sketch them in one paragraph each. The Notation quickstart below carries the symbols; the supporting material holds the full glossary.\n",
+    "\n",
+    "A **differential game** is the optimal-control generalisation of a saddle-point problem to continuous time. Two players control inputs `u_P` and `u_E` that jointly determine the state's evolution `\u1e8b = f(x, u_P, u_E)`. They have opposed objectives \u2014 one minimises a running cost `L(x, u_P, u_E)`, the other maximises it \u2014 and we look for a *saddle point* `(u_P*, u_E*)` where neither player can improve unilaterally. Isaacs's 1951 contribution was to formulate HC this way and to show that saddle points exist for many adversarial pursuit scenarios; the existence question is the hard part, and the answer depends on the singular-surface structure \u00a77 visits.\n",
```
JuanCab, line 176: Please indicate `P` means "pursuer" and `E` means "evader" in these subscripts.  It may be obvious to you, but I was pondering over it.   Also, given the goal of making the reader intuit what is going on, you should describe PRECISELY what `x` and `u` are meant to represent.  Based on your talk at SciPy I gather they are relative position of the two participants and angular velocity (e.g. turn rate) of the specific participants.

Actually, upon reading the "Notation quickstart" I see you included descriptions there of the variables, but I would still include it in the text as read.

Also, if I understand the notation, x is likely multidimensional and u could be (if the field of play was more than 2-D, say 3-D). The convention I am aware of suggests vectors like this should be boldfaced to indicate they are multidimensional.  You have constrained the game to be 2-D in your examples, so only `x` might be multidimensional.  This is a formatting issue which really comes down to what the proceeding's editors want.  I noticed the notation in the "quickstart" is consistent with this, but it would be nice if the rendered notation in the text was the same format.

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+    "\n",
+    "The **Pontryagin maximum principle** (PMP) is the calculus-of-variations machinery that locates the saddle point. It introduces a costate vector `p(t)` (the same dimension as the state) and a Hamiltonian `H(x, p, u_P, u_E) = p \u00b7 f + L`. The optimal controls minimise (resp. maximise) `H` over their feasible sets, and the state-costate pair evolves under Hamilton's equations `\u1e8b = \u2202H/\u2202p`, `\u1e57 = -\u2202H/\u2202x`. For HC the Hamiltonian turns out to be linear in the pursuer control `\u03c6`, which makes the optimal pursuer control \"bang-bang\" (always at the boundary of its feasible set, with sign chosen by a *switching function*). This linearity is the structural fact \u00a76 exploits to derive the switching function in two SymPy calls.\n",
```
JuanCab, line 178: Since p here is stated to be the costate VECTOR should it be boldfaced?  Also, it might made sense to define what a costate vector is for non-mathematicians struggling with these concepts (like me).  I did have to google it to realize why momentum is the co-state vector of position, which is the context I had seen this in.

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+    "\n",
+    "The notation quickstart names the symbols this paper uses; each one is also defined at first technical use, and the supporting material holds the complete glossary including the body-frame dynamics and singular-surface vocabulary. With these handles in place, \u00a75 turns to the HC-specific work: collapsing the 5-DOF lab-frame setup to the 2-DOF body-frame system every subsequent derivation operates on.\n"
```
JuanCab, line 180: This is repetitive with the statement 3 paragraphs back.  Either (i) strike this 'introduction' to what we are about to see in the notation quickstart here and move this up or (ii) strike the earlier mention and keep this here (which I think reads better).  If you go with option (ii) I would probably insert "following" before "notation quickstart names..."

Also "DOF" should be defined as Degrees of Freedom on this first pass.  I understand it is a standard abbreviation, but the style guide suggests "Acronyms should be spelled out upon first mention."

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   "metadata": {},
+   "source": [
+    "**Notation quickstart.**\n",
+    "Just the abstract objects of a Pontryagin / differential-game setup so the prose has handles to reach for. The HC-specific symbols ($x_1, x_2, \\phi, \\psi, \\sigma, \\ldots$) are introduced where they first appear in \u00a75\u2013\u00a78; the full notation reference is in the **supporting material**.\n",
+    "\n",
+    "| Term | Notation | Quickstart description |\n",
+    "|---|---|---|\n",
+    "| State vector | $\\mathbf{x}$ | The system's evolving variables; in HC, the 2-DOF body-frame relative position |\n",
+    "| Player controls | $u_P,\\ u_E$ | The two players' choices; in HC, pursuer turn rate $\\phi$ and evader heading $\\psi$ |\n",
+    "| Costate / adjoint | $\\mathbf{p}$ | Pontryagin's adjoint vector; equals $\\nabla V^*$, satisfies $\\dot{\\mathbf{p}} = -\\partial H / \\partial \\mathbf{x}$ |\n",
```
JuanCab, line 195: I understand "costate" and "adjoint" here, but you refer to "costate vector" earlier and "adjoint vector" here.  I would suggest you stick to one convention (although the term being "Costate/adjoint vector" makes sense in the column, just suggesting you stick to one convention in the text.

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "sec5",
+   "metadata": {},
+   "source": [
+    "## Reducing HC to body-frame coordinates (the symbolic-view warm-up)\n",
+    "\n",
+    "The HC state is naturally described in the lab frame with five degrees of freedom: pursuer position `(x_P, y_P)` and heading `\u03b8`, evader position `(x_E, y_E)`. The pursuer's kinematics are `\u1e8b_P = cos \u03b8`, `\u1e8f_P = sin \u03b8`, `\u03b8\u0307 = \u03c6` (with `|\u03c6| \u2264 1` the normalised turn-rate control, `v_P = 1`, and minimum turning radius `R_min = 1`). The evader's kinematics are `\u1e8b_E = w cos \u03c8_lab`, `\u1e8f_E = w sin \u03c8_lab` where `\u03c8_lab` is the evader's lab-frame heading. Five DOF, two players, two controls \u2014 well-posed but ungainly.\n",
```
JuanCab, line 209: I will note for the last time that it would be nice to have the notation match the style of the notation quickstart, but that is an editorial decision I suppose.  

I do think it might have been nice to have most of the content of this paragraph describing the HC state come before the first paragraph of the previous section, simply because it defines some the variables nicely (you could have replaced u_P and u_E with theta and psi and dropped the u_P and u_E which never get mentioned again anyway)  Then you could say solving for the optimal solution for the evolution of x requires calculus of variations and go into that discussion, where you use those variables.  It would have helped me to see these variables defined like this before going into describing the mathematical techniques needed to solve the problem.  I don't think you can entirely drop this paragraph here but I think it is better to define the variables before using them.

Given the more prominent role $w$ plays later in the paper, I think it would have made sense to call it out here.  Right now, it looks to me like a velocity for the evader (as a ratio to pursuer, I assume), but it is never defined.

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "sec5-postderivation",
+   "metadata": {},
+   "source": [
+    "SymPy did the algebra Isaacs would have done by hand. The reader can re-run with a swapped convention (say, evader $\\sin \\leftrightarrow \\cos$) and watch the simplified form change.\n",
+    "\n",
+    "Coordinate changes that exploit symmetry to reduce state-space dimension are a standard move in dynamical-systems analysis \u2014 and often necessary in differential games, where the full state may be too high-dimensional for a value-function field to be tractable. The reduction is also *invertible* given the pursuer's initial lab pose `(x_P(0), y_P(0), \u03b8(0))`: composing the body-frame trajectory with the integrated lab-frame pursuer state recovers the full lab-frame chase. That is why \u00a73's `chase_demo` (lab frame) and \u00a77's `dispersal_crossing` (body + costate + strategy) render the *same* trajectory from different coordinate views.\n",
```
JuanCab, line 490: I am not used to references to future code earlier in the paper, but I will admit I mostly get the point. I am more used to something stating "We will show that this change of coordinates renders the same trajectory in section 7".  Not necessarily asking for any correction here, just stating it doesn't read like most papers I have read (I don't read mathematical theory papers much).

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+    "\n",
+    "Two figures support the reduction. `problem_geometry` (the lab-frame schematic) names the symbols visually: pursuer heading `\u03b8`, turning circle of radius `R_min`, capture circle of radius `\u2113` around the pursuer, evader velocity vector at angle `\u03c8_lab` to the lab `+x` axis. `coordinate_progression` (the pair of panels) shows *why* the reduction is interesting: three different lab configurations \u2014 pursuer here-facing-east, pursuer there-facing-north, pursuer elsewhere-facing-northwest \u2014 all collapse to the *same point* on the body-frame panel. Three lab states, one game state.\n",
```
JuanCab, line 492: There are an infinite number of lab configurations that will collapse into the same point on the body-frame panel... is this not obvious? That's like saying that there an infinite number of potential origins we can use, but changing the origin doesn't change the fundamental differential equation problem. Not trying to sound snarky, its just I first read this and the *why* the reduction is interesting seemed focused on "three" different lab configuration... the idea is that the body-frame really can reduce the complexity of the problem (as you say in the next paragraph we go from 5 to 2 DOF).  Can you rephrase this a bit to make it clear that the figure just shows three of the infinite possible lab states corresponding to the game state?

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
...6V/ND1WW3bt1cUFSDSZx77rlumoULF7rPaPk1vd4/8sgjg2bWbdy40R5//HE79thj3XatYKi2h+OOO869N2rUKFdfCtbpNWWyyYoVK9x3LFu2zNWFvuPAAw+01q1bu/e3bt3qmobqNW1va9ascXVy2GGHFdRXINWBptN3vPfee66O9H+vjjzaLr777ju3nGlpaa4ODjjggIhse0BJkTkGAOUkvWqadd2/lXXZt6XtzMozS/rnwggAAERuhEoFIuKdmg6eddZZ7v9q5njllVfaKaec4v5WIERBIWWen3baaS7Qo2CNAgiiYICCBgoEnXDCCe7zClb85z//CXldoe/SdyqwoO+68MILC75LAQpvXnpfgQwFIjyqUwVF9B3/+te/7KCDDnLBiRdffNF9n8p49tlnu2k/+eSTgs/99ttv9tVXX9ngwYPtoosucgGmb775xhYtWhR2tpECMwMGDHDz148CLJ999lmh6bVcCgYpWHXmmWe6QNkXX3wR8jsUKFLZFWDUMp9//vnWv39/18zW6ydMwSL1Wad66tGjhwvwKENNFPRR8Eff2adPH7vgggtcIElNZEXr7dJLL3UBHgUZVd+qg8DAn75DATgFnbQO9TkFAr3l1/e9++67LmCk97X8vXv3tqpVq4asO5kwYYKra61z1d9HH31kv/76qwtMnX766W66SZMmFXxOAVEFX/WevkfBKPXjFjjfyZMnu7KqvhSE0nxDbW9r165126malarM+ox/HYmCwq+88oq1bdvWva+AnbY9bTdl3faA0iBzDADKWYO2tdyPzP1lmSWnJVvrfo0sOYXnFQAAJMIIlQqEKCihjCP/gImXcbPHHnvYkCFD3GvKatJ0CgqIso2UZaMMIc1H9t9/f9exvrLDgmXPZGVluXkrw8YLAnmBD2XyKOghykhS5pD3XQokKTNN8/fv10sBKs1L2UOeffbZx15++WX3PSqHgj/KgNL8RJlTyqJSkE+fDeQtu3999OzZs9A0CpS9//77hcqvulFQT3UiCkR9/vnnIetegTNll6n+PKpjLxijQJICLsrk8+an7C4FbhSk0Xeq3o8++uiCZVNfYn///XfBulUGloJDCgipKaU3by/TShlpCoSddNJJLlPPW4cKWnnLryCigoG77bZbQVPMUJmB/nUxYsSIggCavn/x4sUuyOfVj5ZLQTOPlxHnUXa
```
JuanCab, line 1298: When I looked at this section, it also rendered with the section number at the front (and the "{.unnumbered}" as part of the section title).

JuanCab, line 1288: When I looked at this section, it rendered with the section number at the front (and the "{.unnumbered}" as part of the section title). I don't know if this will be able to fixed later by Curvenote before final rendering online.

JuanCab, line 572: This may be math terminology, but keeping in mind the audience are not supposed to be mathematicians, can you clarify "The plan is backward characteristics."?  The rest of the paragraph does explain some of this, but "characteristics" appears to be domain-specific and I am mostly understanding it from context... which makes me uncertain I understand it. 

It might make sense to state "Our approach is called backward characteristics" and then explain the what a "characteristic" is in mathematics. 

By the way, this opening almost reads like how Claude (AI) likes to describe plans of attack for the problems it solves.

**Other Notes**:

Given the way you used formatting earlier in the paper, was this sentence meant to be boldfaced as a sort of 'subsection indicator'?

"V^* is its viscosity solution" Huh?  Again, is this mathematical terminology because viscosity has a completely different meaning in physics.

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "sec4",
+   "metadata": {},
+   "source": [
+    "## A crash course in *differential game* and *calculus of variations*\n",
+    "\n",
+    "Calculus of variations and optimal control are load-bearing tools across applied mathematics \u2014 used routinely in aerospace, robotics, control engineering, mathematical biology, economics, and quantitative finance, though rarely taught outside narrow specializations at the undergraduate level. We chose a differential game rather than a single-player optimal-control problem because pursuit-evasion is essentially the children's game of *tag*: every reader has embodied experience of chasing and being chased, and that intuition is a place to hang the formalism on.\n",
```
JuanCab, line 172: Do you want to mention "control" here is referring to the player controls?  Took me a while to reason that out since I never head of this class of problems before.

papers/michael_zargham/paper.ipynb
JuanCab, line 750: Is $\alpha$ or $w$ defined in this paper, I couldn't find it.  I know $w$ is mentioned  in the equation for the evader's velocity, but it is not explicitly defined! The lack of definition made it hard to determine what the heck is meant by this statement. 

Given this is the crux of what is going on in section 8, it seems like this needs a clearer introduction/definition.  I don't think we should be expected to read Merz 1971 to get a definition.

papers/michael_zargham/paper.ipynb
```diff
@@ -0,0 +1,1586 @@
...
+   "source": [
+    "**Notation quickstart.**\n",
+    "Just the abstract objects of a Pontryagin / differential-game setup so the prose has handles to reach for. The HC-specific symbols ($x_1, x_2, \\phi, \\psi, \\sigma, \\ldots$) are introduced where they first appear in \u00a75\u2013\u00a78; the full notation reference is in the **supporting material**.\n",
+    "\n",
+    "| Term | Notation | Quickstart description |\n",
+    "|---|---|---|\n",
+    "| State vector | $\\mathbf{x}$ | The system's evolving variables; in HC, the 2-DOF body-frame relative position |\n",
+    "| Player controls | $u_P,\\ u_E$ | The two players' choices; in HC, pursuer turn rate $\\phi$ and evader heading $\\psi$ |\n",
+    "| Costate / adjoint | $\\mathbf{p}$ | Pontryagin's adjoint vector; equals $\\nabla V^*$, satisfies $\\dot{\\mathbf{p}} = -\\partial H / \\partial \\mathbf{x}$ |\n",
+    "| Hamiltonian | $H(\\mathbf{x}, \\mathbf{p}, u_P, u_E)$ | The Pontryagin function; its gradient drives the state and costate ODEs |\n",
```
JuanCab, line 196: This is the first use of ODE.  Per the style guide "Acronyms should be spelled out upon first mention."

papers/michael_zargham/paper.ipynb
JuanCab, line 762: Is there not a better way to refer to the cell than its ID number which isn't visible to me in the online paper?  

Also, this sentence with the cell reference is not grammatically correct.  At a minimum "state" is missing an article (e.g. "the state"). Also, since bang-bang is the control, I think it makes more sense to say "the bang-bang control phi^*" to clarify, especially since you use an article to describe the velocity components.

JuanCab, line 764: Again, the use of a code rather than figure number here is confusing (I'd just drop everything in parentheses). Is this left over from an AI?

JuanCab, line 768: Do you want to explain "singular surface analysis" to the non-domain specialist?  If not, do you want to drop the mention of this as a "dispersal surface"?  

I suspect your point is to say that this a nice illustration of the mathematical concept, but it needs a bit more support to bring me, a non-mathematician, in to a place where it seems important.  This entire paper is about how sympy can be educationally valuable, so you can spend a few more words explaining what each of these figures brings to that argument.

JuanCab, line 770: Again, what is an "equivalence-class representative" in non-domain specific language?

JuanCab, line 780: The glass-box seems a bit smoky to me. What do you mean by "seams between coordinate systems"?  A seam is usually a boundary, these coordinate systems are not TOUCHING, they are simply differing transformations of the visualization of the system.  The use of "seam" seems a mismatch to its common use, for one thing, we can cross a seam, and one can not 'cross' between transformations.  Each panel holds the full ODE solution.

JuanCab, line 873: Again, the language is slipping here to be a bit 'idiomatic'.  Instead of "Section 8 has two jobs" instead say the slightly more format "This section has two goals". and I would NUMBER the goals to make them clear  I will admit this may just be me preferring  slightly more formal style. 

I do believe the use of "seam" in this context is a bit confusing for the reasons I cited above (is there a boundary to be crossed between SymPy and NumPy?).

"walk the figure" is also idiomatic, although not as hard to follow.

JuanCab, line 958: It is not clear to me at all why a smoothed interpolation would automatically over-report the capture time.  Can this be clarified?

"Same chase, two sections, one number." Huh? It is not clear what this is referring to. What "sections"? What "number"?

JuanCab, line 960: I don't see any dark grey no-capture region in this figure except in its legend.  I assume that is because there are no "no-capture" regions in this figure..

JuanCab, line 962: I can guess what you are referring to as the "kidney", but giving its rough x_1, x_2 coordinates would help.

"artefact" is the British spelling of "artifact".  no big deal, but given the paper is full of American-flavored idioms, its striking.

JuanCab, line 1103: This somewhat defines "the seam"... this definition should either have come sooner, explaining the interpretation difference between numpy (numeric) and sympy (symbolic), or just been dropped as confusing.

JuanCab, line 1149: Again, is "viscosity" here domain-specific language?  Certainly it isn't the physics definition.

JuanCab, line 1159: The conclusions are STRONG.  This part basically explains what the point of the entire paper was and it only uses mathematical language that was introduced earlier.

JuanCab, line 1171: The last sentence is weird, it didn't need the "A honest moment" prefix to the idea that "glass-box doesn't mean omniscient"

JuanCab, line 1260: This use of "seams" actually appears appropriate, as it is the 'boundary' between the different worlds mentioned by Tall.

JuanCab, line 1274: Its a bit funny to state "personal motivation" for this paragraph and then start with "we"?  Maybe just "Motivation."?

JuanCab, line 1290: I guessed by some of the stylistic choices in the text this was Claude.  I recognized its fingerprints in the prose style.

papers/michael_zargham/supporting_material.ipynb
```diff
@@ -0,0 +1,666 @@
...
+    "\n",
+    "Companion to the SciPy 2026 paper by Michael Zargham. Contains the full **notation reference** (Appendix A) and the **verification audit** (Appendix B): the test catalog, a live in-notebook pytest run, and the symbolic-vs-numerical conservation cross-check. Run from `papers/michael_zargham/` with the paper's `uv` environment.\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "appendix-a-notation",
+   "metadata": {},
+   "source": [
+    "## Appendix A \u2014 Notation reference {.unnumbered}\n",
```
JuanCab, line 17: Just a note that this section does appear numbered and with the text "{.unnumbered}" appearing in the title when I preview it in Curvenote.

papers/michael_zargham/supporting_material.ipynb
```diff
@@ -0,0 +1,666 @@
...
+    "| **Singular surfaces** | | |\n",
+    "| Dispersal surface | $\\{\\sigma = 0\\}$ | Locus where $\\phi^*$ flips sign; chase_demo crosses this at the \u2605 |\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "appendix-b-intro",
+   "metadata": {},
+   "source": [
+    "## Appendix B \u2014 Verification audit {.unnumbered}\n",
```
JuanCab, line 64: Same as the previous section, this section does appear numbered and with the text "{.unnumbered}" in it when viewed in Curvenote.

papers/michael_zargham/supporting_material.ipynb
```diff
@@ -0,0 +1,666 @@
...
+    "| **Costate / adjoint** | | |\n",
+    "| Adjoint variables | $p_1, p_2$ | Pontryagin's costate; equal the gradient $\\nabla V^*$ of the value function |\n",
+    "| Costate magnitude | $\\|\\mathbf{p}\\| = \\sqrt{p_1^2 + p_2^2}$ | Conserved along every optimal characteristic (see \u00a78) |\n",
+    "| **Controls** | | |\n",
+    "| Pursuer turn rate (bang-bang) | $\\phi \\in [-1, 1]$ | Normalised; $\\|\\phi\\| \\le 1$ is the minimum-turning-radius constraint |\n",
+    "| Evader heading (body frame) | $\\psi$ | Lab-frame heading minus pursuer's $\\theta$ |\n",
+    "| Optimal pursuer control | $\\phi^* = -\\operatorname{sign}(\\sigma)$ | Bang-bang law from $H$ being linear in $\\phi$ |\n",
+    "| Optimal evader heading | $\\psi^* = \\operatorname{atan2}(p_1, p_2)$ | Evader points velocity along the costate direction |\n",
+    "| **Parameters** | | |\n",
+    "| Evader-to-pursuer speed ratio | $w = v_E / v_P \\in (0, 1)$ | The pursuer is always faster in this paper's regime |\n",
```
JuanCab, line 41: Would have been nice to see this in the paper. I understand the role of the appendix to help, but when crucial to the argument of the prose, it should go there as well!

papers/michael_zargham/supporting_material.ipynb
JuanCab, line 1: This supporting material section seems quite reasonable and readable.

JuanCab (changes requested):
I did review all the code provided by the author.  The code ran without error as did all the tests provided.  Their results are reproducible.

I want to start by saying that I think the idea behind this paper is excellent.  I like the idea of using SymPy to allow math students or those exploring mathematical concepts to skip the tedium of the "process" they already know (algebra, derivatives, integrals) and that can be error-prone when done by hand and instead focus on the steps in a derivation that require some human intelligence (e.g. picking which substitutions to make).  

I also want to emphasize I am NOT a mathematician.  I am an astrophysicist by training.  I felt I could follow most of the mathematical notation and ideas, once I dug into them.  However, the paper, with its very heavy use of domain-specific language, made it difficult for me to follow the logic in some places.  The [review criteria](https://github.com/scipy-conference/scipy_proceedings/blob/2026/review_criteria.md) specifically state that "Due to the interdisciplinary nature of SciPy, highly domain-specific jargon should be avoided or explained where possible."  I do not feel this paper did that.  Certainly, I think the ideas CAN be communicated to a more general audience of scientifically-literate and math-capable readers who understand differential equations but are not necessarily spending their professional lives exploring them.  I would be much happier with this paper if it was reworked with an eye toward that interdisciplinary audience. For example, I imagined using SymPy in an advanced undergraduate Physics course (e.g. Advanced Electricity and Magnetism) once students have learned about how to set up the mathematics with all the bounds, to actually work through the tedium of the integrations/derivatives.  This paper gave me a way to express the pedagogical motivation for that.  Unfortunately, it was a battle to understand how I could apply its ideas in my discipline because of its choice of jargon.  Explaining more of that jargon to the non-mathematican would have helped a lot.

Assuming the journal still is aimed at an interdisciplinary audience, **I would not accept the paper until such changes are made.**  That said, I really hope the author puts in the effort to rework this, it is a very interesting idea to me as an educator and I want to see this published!


**Other general issues with the paper:**

The author is honest about using an AI to draft this.  I see a lot of places where that AI's fingerprints appear to be showing. Its not offensive or anything, but I would have preferred to see more of the authors voice (their talk was quite good at SciPy).  

Another issue throughout the paper, possibly related to the use of AI, was the preponderance of unicode characters instead typesetting with LaTeX for variables/equations in the text.  Given the use of LaTeX typesetting in other places in the paper, the mixture is jarring.

Finally, I think with a paper like this that introduces a fairly complex mathematical idea to an interdisciplinary audience, care needs to be taken to define all terms and all variables when they are first seen.  There are many cases where equations are presented in the text and the variables are only defined much later in the text or in the "supporting material".  If a paper is going to communicate to an audience new to the subject matter, it has to be very clear about definitions of jargon terms and mathematical variables.  I have tried to provide notes about where I noticed this issue, but I am sure I didn't catch every mention of some new variable/idea that wasn't explained.

**AI Use by reviewer:** As a reviewer, I did use AI (Claude) to check a few ideas I had about grammar and to re-express some of the ideas from the paper so I could comprehend them better. I never fed it more than 2 sentences from the paper and I wrote every editorial comment by hand. I was tempted to ask Claude to highlight every situation in which a complex mathematical idea or variable was introduced without explanation, but I leave it to the author to consider doing that.

ktzhao:
I have reviewed the manuscript. The accompanying test suite and interactive notebook workflows run smoothly.

I largely agree with JuanCab's overall assessment. The core concept of the paper represents a significant pedagogical advancement. Specifically, enabling learners to offload tedious algebraic manipulations to a Computer Algebra System (CAS)—while maintaining high-level control over derivation steps, such as constructing a Hamiltonian or selecting coordinate transformations—is highly valuable and merits publication.

Coming from an astrophysics background, I can fully appreciate the dense technical terminology used in the paper. Given that the case studies are explicitly framed within the context of optimal control and differential games (specifically, Isaacs's "Homicidal Chauffeur" problem), these terms constitute standard, essential concepts rather than obscure jargon.

JuanCab mentioned the application of SymPy in an "Advanced Electromagnetism" course. By analogy, a paper demonstrating SymPy in that context would naturally employ terms like "vector potential," "gauge transformation," or "spherical harmonics." While these concepts are fundamental to physicists, they may be unfamiliar to readers from other disciplines. Therefore, we should not dismiss or penalize a paper simply because a domain-specific case study employs terminology unfamiliar to a broader audience.
I consider it acceptable to place supplementary content in the supporting materials to keep the main text concise; this should certainly not be grounds for rejection. However, during the revision process, I do advise the authors to improve the paper's overall readability and eliminate some of the AI-generated phrasing.

Reviewer's Statement on AI Usage: As a non-native English speaker, I utilized AI tools solely for grammatical polishing and vocabulary refinement to ensure accuracy. All evaluative comments and substantive viewpoints are entirely my own.

mzargham:
Thanks for the reviews. This is very helpful since sometimes its hard to separate jargon out when you've been in a field for a long time. i know the review cycle ends tomorrow so i'll be spending the next few hours (or however long it takes) working through the document on jargon clean up duty. I will aim to strike a balance between terms which are necessary for the topic. I would also like to add as evidence that the talk in person at SciPy was well attended and well received. Though, in keeping with your feedback. i did thin the content out a little bit to keep it in the time window. the recording should be up on the scipy youtube channel soon(tm).

papers/michael_zargham/paper.ipynb
mzargham: Qualified in b3a6cb0c: the abstract now opens the timing claim with "With this exemplar's parameter settings", and section 3 states the settings (w = 0.45, l = 0.5) where the comparison figure is introduced.

papers/michael_zargham/paper.ipynb
mzargham: Reworded in b8f34fda. It now reads "follow the chain from a derived expression, to the function it lambdifies into, to the figure it renders — and verify every link." No more seam to read.

papers/michael_zargham/paper.ipynb
mzargham: Reworded in b8f34fda: "This section poses the puzzle; Sections 4 through 8 derive the answer."

papers/michael_zargham/paper.ipynb
mzargham: Done across c39804b2, e6ca9d5e, and da7749cf: section 4 now names the pursuer P and evader E, ties u_P/u_E to the concrete turn rate and heading, and the prose math is typeset in LaTeX with bold vectors matching the notation table.

papers/michael_zargham/paper.ipynb
mzargham: Defined at first use in c39804b2: the costate is introduced as the analogue of momentum in classical mechanics, measuring how sensitive the optimal cost is to each state component. The vector is now bold.

papers/michael_zargham/paper.ipynb
mzargham: Went with your option (ii) in b8f34fda: the earlier announcement is gone and the later one reads "The following notation quickstart...". DOF is spelled out at first use (c39804b2, da7749cf).

papers/michael_zargham/paper.ipynb
mzargham: Standardised on "costate" in the prose (c39804b2). The table row label keeps "Costate / adjoint" as the one bridge between the two conventions.

papers/michael_zargham/paper.ipynb
mzargham: Adopted in full in da7749cf: the lab-frame state paragraph now sits in section 4 ahead of the differential-game and PMP paragraphs, with the kinematics set as a display equation (31300c2c). w is defined in section 3 prose and in the notation table (c39804b2).

papers/michael_zargham/paper.ipynb
mzargham: Understood, and left as is per your note. The forward reference is a deliberate part of the executable-paper style.

papers/michael_zargham/paper.ipynb
mzargham: Rephrased in c39804b2: "three of the infinitely many lab configurations sharing one relative geometry ... Infinitely many lab states, one game state."

papers/michael_zargham/paper.ipynb
mzargham: Fixed in 8674bdba: the attribute is removed and the bibliography block supplies its own References heading, which renders unnumbered.

mzargham: Fixed in 8674bdba: attribute removed; Acknowledgments renders unnumbered.

mzargham: Reworded in c39804b2: "The plan is the method of characteristics, run backward: along special curves (the characteristics — here, the optimal trajectories themselves), a partial differential equation (PDE) reduces to ordinary differential equations we integrate directly." The viscosity solution is explained in place as the generalised-solution notion for PDEs with corners, with a note that the term is historical and unrelated to physical viscosity. The sentence-fragment style you noticed is also reworked (b8f34fda).

papers/michael_zargham/paper.ipynb
mzargham: Clarified in c39804b2 and da7749cf: controls are identified as the players' steering variables (turn rate and heading) at first use.

papers/michael_zargham/paper.ipynb
mzargham: Both defined now. alpha is introduced where it appears as the angular position on the capture circle where each characteristic terminates; w is defined in section 3 prose (evader-to-pursuer speed ratio) and in the notation table (c39804b2). No trip to Merz required.

papers/michael_zargham/paper.ipynb
mzargham: Spelled out at first use in c39804b2: "ordinary differential equations (ODEs)".

papers/michael_zargham/paper.ipynb
mzargham: Fixed in b8f34fda: the cell-id reference is gone and the sentence now reads "the state evolves under the closed-loop dynamics with the bang-bang control phi* = -sign(sigma)".

mzargham: Dropped the parenthetical in b8f34fda.

mzargham: Addressed in c39804b2: the dispersal surface is introduced in place as "a boundary where two distinct optimal paths become available at once", and section 3 defines the singular-surface taxonomy as "the map of where optimal strategies switch, split, or merge". The "singular-surface analysis" phrasing at this spot is gone.

mzargham: Plain language now (c39804b2): "the one body-frame picture that stands in for every lab-frame chase sharing the same relative geometry".

mzargham: Agreed, and reworded in b8f34fda: "The transformations between coordinate views (lab, body, strategy, costate) are now executable cells". The seam wording is reserved for the one sense you flagged as appropriate, and its remaining use (the SymPy-to-NumPy interface) is now defined at first appearance (c39804b2).

mzargham: Reworked in b8f34fda: "This section has two goals: (1) ... and (2) ...", and "walk the figure" is gone. The SymPy-to-NumPy seam is defined where it first appears (c39804b2).

mzargham: Explained in 58a78d19: the backward fan does not pass exactly through this point, and linear interpolation blends in samples from neighbouring, longer-time characteristics, so the map's value there reads nearly three seconds high; the directly measured capture time is used instead. The cryptic closer now reads "The same chase thus appears in section 3 and section 8 with the same capture time" (b8f34fda).

mzargham: Correct, there are none in this figure. The prose now says so explicitly (58a78d19): such cells "would render in dark grey (the legend's unreachable entry); none occur within this figure's extent."

mzargham: Coordinates added in 58a78d19 (roughly |x_1| <= 3, -2 <= x_2 <= 4). Artefact is now artifact throughout the paper and supplement (b8f34fda).

mzargham: Agreed on the timing: the definition now comes at the term's first appearance in section 8 (c39804b2), so this passage elaborates an already-defined term. The confusing earlier uses in section 1 and section 7 were reworded (b8f34fda).

mzargham: Yes, it is standard terminology in the viscosity-solutions literature. The first use now carries a short definition (the generalised-solution notion for PDEs with corners; a historical term, unrelated to physical viscosity), and this second mention reuses the HJI acronym (c39804b2).

mzargham: Thank you. No change made here.

mzargham: Dropped in b8f34fda, along with the fragment above it. The passage now ends "Glass-box does not mean omniscient; it means inspectable", extended in 927820c1 with a note on where human judgment enters (choosing tactics) and how the choice is preserved as version-controlled, verifiable code.

mzargham: Noted, thanks. That use is kept; the other senses were reworded so the word now carries one meaning.

mzargham: Changed to "Motivation." in b8f34fda.

mzargham: The prose has been reworked toward plainer, less idiomatic phrasing throughout (b8f34fda, c8d7ca63), and all prose math is now typeset in LaTeX (e6ca9d5e).

papers/michael_zargham/supporting_material.ipynb
mzargham: Fixed in 8674bdba: attribute removed; renders unnumbered.

papers/michael_zargham/supporting_material.ipynb
mzargham: Fixed in 8674bdba: attribute removed; renders unnumbered.

papers/michael_zargham/supporting_material.ipynb
mzargham: Agreed. The speed-ratio definition is now in the paper as well: section 3 prose and a notation-table row (c39804b2).

papers/michael_zargham/supporting_material.ipynb
mzargham: Thank you. The supplement also picked up small fixes in this revision (test catalog completed, notation typos, its live test run refreshed).

mzargham:
Thank you both for the careful reviews. The revision is on the branch (18 commits, CI green). Each commit message names what it addresses, and I have replied to each inline comment with the commits that resolve it. A summary of the changes:

For the accessibility concern at the center of @JuanCab's review: every flagged term now carries a one-clause plain-language definition at first use (costate, method of characteristics, viscosity solution, dispersal surface, the singular-surface taxonomy), every symbol is defined before it is used (w, alpha, tau, f_1/f_2, the capture radius), acronyms are spelled out at first mention, and the homicidal-chauffeur state is introduced concretely, with its kinematics as a display equation, before the optimal-control machinery that uses it. Per @ktzhao's assessment, the standard terminology itself is retained and explained rather than removed, and the supplement keeps its role carrying depth, with the one definition crucial to the argument (the speed ratio) now also in the paper.

Both reviews asked for plainer prose: idiomatic constructions were replaced with direct statements, punctuation now favors conventional marks, and all prose mathematics is typeset in LaTeX with bold vectors matching the notation table.

I also received useful feedback presenting this work at SciPy and accommodated it: figures now precede their panel-by-panel narration, and the saddle-point Hamiltonian is reduced symbolically to the exact form cited from Isaacs, with the equivalence asserted by `sp.simplify` in the notebook. Alongside these, I reran my consistency review of the revised manuscript against the code and cited sources corrected the remaining prose bugs it surfaced: the test count (36, not 35), the parameter dependence of the capture guarantee (now scoped by Isaacs's criterion), panel-narration details, and stale docstrings.

The code and tests you verified are unchanged apart from figure-title wording and comment/docstring text; the full suite passes after every such change, and the supplement's live test run reflects the current suite.

mzargham:
i was just doing another final cleanup of my local materials and i realized i was "behind" due to changes on the main repo so i integrated them now to derisk future errors. looks like CI came back clean.

chongshenng:
Many thanks @mzargham, for carefully addressing the reviewer comments.

Additionally, thank you @ktzhao and @JuanCab for the detailed review -- is the revised manuscript acceptable in the current state?

JuanCab:
> Many thanks @mzargham, for carefully addressing the reviewer comments.
> 
> Additionally, thank you @ktzhao and @JuanCab for the detailed review -- is the revised manuscript acceptable in the current state?

I need time to review the new draft, I have pencilled in time this coming weekend (Sept. 5).  Is that fast enough?

chongshenng:
> > Many thanks @mzargham, for carefully addressing the reviewer comments.
> > Additionally, thank you @ktzhao and @JuanCab for the detailed review -- is the revised manuscript acceptable in the current state?
> 
> I need time to review the new draft, I have pencilled in time this coming weekend (Sept. 5). Is that fast enough?

Thank you @JuanCab, yes Sept 5 is fine. The decision deadline is on Sept 6.

ktzhao (approved):

JuanCab (approved):
Thanks for all the changes and addressing my concerns.  My second pass through the paper was a lot less "challenging" for me as much of the domain-specific terms were defined and the notation stayed in one consistent format.  Thank you @mzargham for carefully addressing my concerns.  I approve this paper for publication.

chongshenng (approved):

PR #1215, opened 2026-06-09
Paper: PanelBox - A Comprehensive Python Library for Panel Data Econometrics

guhaase:
Submission for the SciPy 2026 Proceedings, accompanying the talk *PanelBox: A
Comprehensive Python Library for Panel Data Econometrics*. Authors: Gustavo Coelho Haase, Paulo
Henrique Dourado, Osvaldo Candido.

Hi @guhaase, this is Sanhita Joshi (@sanhitamj). I will serve as the editor from the Proceedings Committee. Let me know if there is anything I can help with.


papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,424 @@
...
+  correction), discrete choice, count data, spatial econometrics, stochastic
+  frontier analysis, quantile regression, panel VAR/VECM, censored and
+  selection models, and the first comprehensive Python implementation of panel
+  unit-root and cointegration tests. The library adds 50+ diagnostic tests,
+  11 standard-error estimators, four bootstrap methods, an integrated
+  HTML/LaTeX/Markdown reporting system with interactive Plotly visualizations,
+  and 103 bundled datasets, all on top of the scientific Python stack
+  (NumPy, pandas, SciPy, statsmodels). We validate PanelBox against Stata's
+  `xtabond2` and R's `plm`, obtaining coefficient differences below 0.01%, and
+  illustrate it with a canonical dynamic labor-demand application. With 3{,}900+
```
rowanc1: I think this is maybe copied from latex? `3{,}900` -> `3,900`.

```suggestion
  illustrate it with a canonical dynamic labor-demand application. With 3,900+
```

apaleyes:
Hi everyone! I have been assigned as a reviewer to this paper. Looking forward to it! Preview looks fairly complete and ready, but @guhaase please let me know if there are still pending changes.

guhaase:
Hi! Thanks @apaleyes  for reviewing our paper!

Yes, this is the final version of the paper and there are no pending changes. We really appreciate your time and look forward to your feedback.

nil68657:
@guhaase - I am an assigned reviewer for the paper. Let me know if there things I need to be aware of apart from what's already explained in README.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,424 @@
...
+NumPy [@harris2020numpy], pandas [@mckinney2011pandas], and
+SciPy [@scipy] — has lagged conspicuously behind in panel econometrics.
+
+Existing Python libraries cover only fragments of the workflow.
+`linearmodels` [@linearmodels2019] provides static fixed- and random-effects
+models and some instrumental-variables support but no dynamic GMM; `pyfixest`
+excels at high-dimensional fixed effects but offers neither GMM nor dynamic
+panels; and `statsmodels` [@seabold2010statsmodels] has time-series tools but
+no panel-specific GMM. Critically, *none* of them implement panel unit-root or
+cointegration tests — routine in Stata (`xtunitroot`, `xtcointtest`) and R
```
apaleyes: Not quite sure em dash is correctly placed here. Perhaps ", which are routine..." would sound a bit more natural

apaleyes:
After reading first 2 sections of the paper, I have to ask. Was any Gen AI used to write it? It is not bad at all and is permitted, but according to Scipy's policy has to be disclosed: https://github.com/scipy-conference/scipy_proceedings/blob/2026/genai_policy.md

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,424 @@
...
+a validated dynamic-GMM implementation with the Windmeijer finite-sample
+correction; the first comprehensive panel unit-root and cointegration testing
+suite; and an intelligent algorithm for unbalanced panels that retains far more
+data than naive complete-case approaches. The library is open source (MIT),
+integrates natively with pandas, and ships with interactive visualization,
+publication-ready reporting, and 103 datasets for immediate experimentation.
+
+## Library overview and design
+
+@fig:architecture summarizes the architecture. A common data layer wraps a
```
apaleyes: I think the figure could use some improvement. Right now it does not really correspond to this paragraph. Figure suggests 6 almost disjoint components in the library, which clearly is not the case. What are their relationships? Any dataflow between the components worth highlighting? What is user-facing and what is internal?

guhaase:
Hi @nil68657, thank you for reviewing our paper!

There isn't anything beyond what is already described in the README that you need to be aware of. The current submission is our final version, and all supporting material should be covered there. If any questions come up during your review, I'd be happy to clarify them.

guhaase:
@apaleyes Thank you for raising this, and for the pointer to the policy. Yes — generative AI (Anthropic's Claude) was used to assist with drafting and refining the prose, editing the LaTeX/MyST markup, and suggesting code for parts of the library and the replication scripts. It was not used to produce the numerical results, the statistical validations (e.g., against xtabond2 and plm), or the references, all of which come from actual runs of PanelBox and were verified by me. I've added a "Generative AI disclosure" section after the Conclusion stating the tool, how it was used, and for what purpose, in line with the SciPy Gen AI policy. The updated version is in the latest commit.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+self-contained HTML, LaTeX, or Markdown.
+
+:::{figure} figures/fig1_architecture.png
+:label: fig:architecture
+High-level architecture of PanelBox: the 11 model families build on a shared
+data layer and feed a common inference, diagnostics, visualization, and
+reporting infrastructure.
+:::
+
+A minimal estimation looks like idiomatic scientific Python:
```
apaleyes: Does `idiomatic scientific Python` actually exist? I think this is highly framework-dependent, e.g. numpy-based and jax-based idiomatic scientific code will look quite different. 

What it actually looks like to me is sklearn-like interface. And it's ok to just say that directly, given how well known sklearn is in the community.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+```
+
+where $\alpha_i$ are entity effects and $\lambda_t$ are time effects. Fixed
+effects are computed by within-transformation using sparse matrix operations
+and pandas `groupby`, so the estimator scales to large $N$ and supports
+unbalanced panels with entity-specific $T_i$.
+
+## Dynamic panel GMM
+
+The flagship capability is dynamic panel GMM. With a lagged dependent variable,
```
apaleyes: This section completely destroys the story flow of the paper for me. I am having hard time understanding how this follows from the previous section, what is panel GMM and how does it relate to the package.

I would suggest a much gentler approach here that considers readers with no subject matter expertise - introduce the core problem this feature solves, give minimal intuition, and then explain what is offered by the package.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+:::{figure} figures/fig2_unbalanced_algorithm.png
+:label: fig:unbalanced
+The unbalanced-panel algorithm: instruments are validated per observation and
+entity-specific instrument sets are assembled into a block-diagonal matrix,
+retaining observations that fixed-template approaches discard.
+:::
+
+## Panel unit-root and cointegration tests
+
+A capability unique to PanelBox among Python panel libraries is comprehensive
```
apaleyes: By contrast, this section reads much friendlier. I would maybe add another sentence to explain what these testing techniques are normally useful for, again for the benefit of less familiar readers

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+from panelbox.validation.unit_root import LLCTest, IPSTest
+
+llc = LLCTest(data, variable="log_gdp",
+              entity_var="country", time_var="year", trend="ct")
+print(llc.test().pvalue)
+```
+
+Together these enable a complete workflow — test for unit roots, test for
+cointegration when variables are $I(1)$, then estimate the appropriate model —
+illustrated in the growth application below.
```
apaleyes: Cross-reference to the section?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+estimator, which guarantees non-crossing quantiles. Panel VAR/VECM provides
+impulse-response functions with bootstrap bands, forecast-error variance
+decomposition, and Granger-causality tests.
+
+## Inference, diagnostics, and reporting
+
+Valid inference is central. All standard errors use the sandwich form
+$\widehat{\mathrm{Var}}(\hat{\boldsymbol{\beta}}) =
+(\mathbf{X}'\mathbf{X})^{-1}\,\hat{\boldsymbol{\Omega}}\,(\mathbf{X}'\mathbf{X})^{-1}$,
+differing in the meat $\hat{\boldsymbol{\Omega}}$: HC0–HC3 for
```
apaleyes: `in the meat` is that a standard term?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+```python
+pb.comparison_table(
+    [ols_result, fe_result, diffgmm_result, sysgmm_result],
+    labels=["Pooled OLS", "FE", "Diff-GMM", "Sys-GMM"],
+)
+```
+
+## Application: dynamic labor demand
+
+We reproduce the canonical application of @arellano1991: a dynamic labor-demand
```
apaleyes: As an example of a simple adjustment to enhance reading flow: why not say here "To illustrate capabilities of the library, we use it to reproduce ..."

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+with system GMM at $\hat{\alpha} = 0.694$ (standard error 0.137) versus
+difference GMM at $0.686$ (0.152), confirming the efficiency gain from the
+additional moment conditions. The wage elasticity is consistently negative
+($\hat{\beta}_w \approx -0.57$) and capital enters positively
+($\hat{\beta}_k \approx 0.38$). Diagnostics validate the specification: the
+Hansen $J$ test does not reject ($p = 0.172$), AR(1) rejects as expected
+($p = 0.003$), AR(2) does not ($p = 0.671$), and the 91 instruments stay well
+below the 140 entities. The results match @arellano1991 to within 0.01%.
+
+:::{figure} figures/fig5_coefficient_comparison.png
```
apaleyes, line 459: What do shaded red and blue regions show?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+`plm` [@croissant2008plm], estimating identical specifications and computing
+relative differences $\Delta = |\text{PanelBox} - \text{reference}| /
+|\text{reference}|$. For the Arellano–Bond system-GMM specification, every
+coefficient and standard error matches `xtabond2` to four decimal places, and
+the Hansen $J$, AR(1), and AR(2) statistics are identical (@tbl:validation).
+Across 25 specifications on 10 datasets — difference and system GMM, one- and
+two-step, with and without collapse, balanced and unbalanced — the mean
+coefficient difference is 0.0003% and the maximum 0.008%. Against `plm`, fixed-
+and random-effects coefficients, the Hausman statistic, and LLC/IPS unit-root
+statistics agree exactly.
```
apaleyes: This is so interesting! Just out of curiosity, can you explain how is this possible? No numerical jitter, no stochasticity, no small implementation fluctuations make any difference?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,438 @@
...
+
+Current limitations include spatial-model scalability beyond $N \approx 2{,}000$
+(mitigated by Chebyshev approximation), GMM restricted to linear dynamic panels,
+and no GPU acceleration. Planned work includes FMOLS/DOLS for cointegrated
+panels, double/debiased machine learning for panel data, GPU acceleration via
+JAX/CuPy, and Bayesian panel models. By eliminating the need to leave Python for
+sophisticated panel analysis, PanelBox makes empirical economic research more
+reproducible, more integrated with the machine-learning ecosystem, and more
+accessible. It is available on PyPI (`pip install panelbox`) and on
+GitHub [@panelbox2025], with documentation, tutorials, and full replication
```
apaleyes: I would actually suggest to use direct github link here, instead of citing Zenodo page. The goal of this link is to get the interested reader to the repo as quickly as possible, right?

apaleyes (commented):
My review comes with a caveat - I have 0 expertise in the subject matter of PanelBox, and most of the terms were completely new to me. Thus, some parts of the paper were probably more confusing to me than to intended target audience.

PanelBox is a collection of models for panel analysis of econometric data, all adhering to sklearn-like interface. Plus many useful utilities for plotting, testing, data loading, error handling. The majority of the paper describes core models implemented in PanelBox, and there are two sections that validate these implementations (with replication study and with statistical testing). I think the paper does a good job of conveying the purpose of the package to someone with deep expertise in panel analysis.

However, for someone like myself, with no knowledge of the domain, the paper is not particularly friendly, and does almost no introduction into concepts behind key features. I have left some comments below in few places suggesting how to mitigate this. But again, I am not sure if this is something to be fixed at all, as scipy papers are not expected to be tutorial-like.

While reading this i found it quite clear that the paper was written with GenAI assistance, and authors are acknowledging at much. Concretely, sections 2-6 all read as bullet point lists transformed into prose, and lack steady storytelling flow. However, I struggle to suggest concrete proposals here, so it would be interesting to know if other reviewers or editors feel the differently. It could be just me! Also, sections 7-9 read much more natural.

I would also like to suggest authors to spend some time explaining to users how the library can be used, extended and adapted to their use case. What is the expected user-facing flow? What customisations are, and are not, allowed? Anything about adapting to data sources not natively supported, and what about those that are supported? In other words, consider the fact that this is not only about replication of existing methods, but also about advertising your package for potential users, and convincing them it's easy and useful enough to try.

papers/gustavo_haase/main.md
guhaase: Agreed — we'll change it to ", which are routine in Stata (`xtunitroot`,
`xtcointtest`) and R (`plm::purtest`)". Thanks.

papers/gustavo_haase/main.md
guhaase: Good point.... the current figure shows the components but not how they relate. We will redesign it to show the actual dataflow: a pandas DataFrame enters the shared data layer, flows through the estimators, and the resulting objects feed the inference, diagnostics, and reporting layers. We will also visually distinguish the user-facing surface (data loading, model classes, result objects, reports) from internal infrastructure (data layer, covariance engines). This also pairs naturally with the new user-facing-flow section we are adding, so figure and text will tell the same story.

papers/gustavo_haase/main.md
guhaase: Fair  "idiomatic scientific Python" is too vague. We'll name the conventions directly: the API follows the `statsmodels` conventions (R-style formulas, `fit()`, `summary()` on a results object), which we chose because they are the lingua franca of econometrics in Python, with an sklearn-like construct-then-fit pattern. We'll rephrase the sentence accordingly.

papers/gustavo_haase/main.md
guhaase: Agreed, and this is the clearest symptom of the flow problem you flagged in the general review. In the revision this section will open with the problem and the intuition before any estimator details, roughly: *when today's outcome
depends on yesterday's (employment adjusts slowly, GDP growth persists), the model must include the lagged outcome as a regressor; but that lag is mechanically correlated with the unobserved entity effect, so OLS and fixed
effects are both biased — in opposite directions. The GMM estimators of Arellano–Bond and Blundell–Bond solve this by using older lags as instruments, and this is the machinery PanelBox implements and validates.* Only then will
we present what the package offers. The bias intuition also connects directly to Figure 5, where the OLS-above / FE-below / GMM-in-between pattern is visible.

papers/gustavo_haase/main.md
guhaase: Thanks... will do. We'll add a motivating sentence up front: these tests answer "does this variable have a stable long-run level, or does it wander?" regressing one wandering (nonstationary) series on another produces spurious correlations, so testing for unit roots and cointegration is the standard safety check before estimating relationships between macroeconomic variables such as GDP, prices, or exchange rates.

papers/gustavo_haase/main.md
guhaase: Yes... we'll add an explicit cross-reference to the growth application in the Application section.

papers/gustavo_haase/main.md
guhaase: It is, surprisingly - the sandwich covariance estimator's outer matrices are called the "bread" and the inner matrix the "meat" in the literature (e.g., Zeileis 2006, "Object-Oriented Computation of Sandwich Estimators", *JSS*, whose R package literally exports `bread()` and `meat()` functions). But given your broader accessibility point, we'll add a brief gloss so readers outside econometrics aren't left wondering: "differing in the middle 'meat' matrix $\hat{\boldsymbol{\Omega}}$ (the term of art for the inner matrix of the sandwich form)".

papers/gustavo_haase/main.md
guhaase: Agreed - we'll adopt this phrasing, and apply the same kind of connective framing at the start of the other sections as part of the flow revision.

papers/gustavo_haase/main.md
guhaase: They were meant to indicate the ranges where the upward-biased (OLS, red) and downward-biased (FE, blue) estimates typically fall, but you're right that they are not defined anywhere in the caption  and on reflection they are redundant with the bars and annotations. We'll remove the shaded bands and let the bars, error bars, and the dashed consistent-value line carry the message.

papers/gustavo_haase/main.md
guhaase: Great question, it's possible because these particular estimators are closed-form, deterministic linear algebra: the fixed- and random-effects estimators solve a least-squares problem with an exact solution $(\mathbf{X}'\mathbf{X})^{-1}\mathbf{X}'\mathbf{y}$, and the Hausman and LLC/IPS statistics are deterministic functions of those estimates. There is no optimization loop, no random initialization, no stochastic component anywhere.
Both PanelBox and `plm` perform the same computation in IEEE-754 double precision (via LAPACK underneath NumPy and R respectively), so the results agree to all reported digits; residual differences from floating-point operation ordering exist but sit at the ~1e-12 level, below any reported precision. "Agree exactly" means exactly that — we'll clarify it in the text as "agree to machine precision". The GMM comparisons in the same paragraph show tiny nonzero differences (max 0.008%) precisely because there the implementations make different internal choices (e.g., ordering of operations in weighting-matrix construction), which accumulate slightly different rounding.

papers/gustavo_haase/main.md
guhaase: Agreed, we'll put the direct GitHub URL in the text so interested readers get to the repo in one click, and keep the Zenodo citation in the references for archival/DOI purposes.

guhaase:
Thank you @apaleyes  for the careful reading, the perspective of a non-specialist is genuinely useful, since SciPy proceedings reach well beyond econometrics. We (my co-author and I) discussed your review and agree with all three points. Before we submit the revision, we'd like to share the plan, since it involves restructuring rather than local fixes.

On reflection, your comments on flow and motivation point to the same root cause. PanelBox originated in our day-to-day work as model validation specialists, validating credit-risk models in production Python environments, where econometric models had to be exported to Stata/R for estimation and validation. My co-author and I are both specialists in model validation, and this practical experience directly shaped the design of PanelBox. The library's modules (validation, estimation, inference, interactive reporting) mirror that production workflow. The current draft buried this origin story and instead enumerated features, which is why sections 2–6 read as lists while the validation sections (7–9) flow naturally: validation is where the library came from.

Concretely, we plan to:

1. Rewrite the introduction around this motivation, adding an intuition-first explanation of the core concepts a general SciPy reader needs (what panel data is, why unobserved heterogeneity matters, whydynamic models require GMM), addressing your accessibility comments.

2. Restructure sections 2–6 as a workflow narrative** (specify → estimate → test assumptions → report) rather than a feature catalogue, keeping the in-depth treatment for the three genuinely novel contributions.

3. Summarize the full breadth (70+ models across 11 families) in a single comparative table against Stata/R equivalents, instead of enumerating in prose.

4. Add a user-facing section covering the expected end-to-end flow, what is and isn't customizable, how to use PanelBox with your own data (any long-format DataFrame with entity/time columns), and pointing to the official documentation's gallery of example notebooks for the complete feature set.

We reply to each inline comment on its thread below. If you see any problem with this direction, please let us know — otherwise we expect to push the revised version by Friday. Thanks again for the constructive review.

apaleyes:
@guhaase that sounds wonderful, i would be happy to read the updated version. No notes on this plan from my side.

papers/gustavo_haase/main.md
apaleyes: Fantastic. It's refreshing to see closed-form calculations in today ML-heavy world. Thanks for the explanation

papers/gustavo_haase/main.md
apaleyes: This is quite funny, thanks for explaining

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,569 @@
...
+  (NumPy, pandas, SciPy, statsmodels). We validate PanelBox against Stata's
+  `xtabond2` and R's `plm`, obtaining coefficient differences below 0.01%, and
+  illustrate it with a canonical dynamic labor-demand application. With 3,900+
+  tests and an MIT license, PanelBox lets researchers run sophisticated panel
+  analysis entirely in Python.
+---
+
+## Introduction
+
+PanelBox did not begin as a general-purpose econometrics library. We work in
```
sanhitamj, line 31: The whole first paragraph in Introductions reads weird, and seems like an after thought. The story of how PanelBox started is necessary here and the rest of the material belongs to the 2nd paragraph here. It gives a good background of what strata and R able to do.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,569 @@
...
+science and increasingly in quantitative social science thanks to
+NumPy [@harris2020numpy], pandas [@mckinney2011pandas], and
+SciPy [@scipy] — has lagged conspicuously behind in panel econometrics.
+
+Existing Python libraries cover only fragments of the workflow.
+`linearmodels` [@linearmodels2019] provides static fixed- and random-effects
+models and some instrumental-variables support but no dynamic GMM; `pyfixest`
+excels at high-dimensional fixed effects but offers neither GMM nor dynamic
+panels; and `statsmodels` [@seabold2010statsmodels] has time-series tools but
+no panel-specific GMM. Critically, *none* of them implement panel unit-root or
```
sanhitamj: Remove the emphasis. Proceedings are supposed to be dry, free from passion.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,569 @@
...
+pooled OLS, fixed effects (within), random effects (feasible GLS), the
+between estimator, and first differences for the model
+
+```{math}
+:label: static
+y_{it} = \mathbf{x}_{it}'\boldsymbol{\beta} + \alpha_i + \lambda_t + \epsilon_{it},
+\qquad i = 1,\dots,N,\; t = 1,\dots,T,
+```
+
+where $\alpha_i$ are entity effects and $\lambda_t$ are time effects. Fixed
```
sanhitamj: Very short description, say a sentence each, of what are entity effects and time effects would be useful for a reader who doesn't work in the field.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,569 @@
...
+capability, and previously had no validated Python implementation.
+
+**Difference GMM** [@arellano1991] first-differences {ref}`dynamic` to remove
+$\eta_i$ and exploits the moment conditions
+$\mathrm{E}[y_{i,t-s}\,\Delta\epsilon_{it}] = 0$ for $s \ge 2$, so lagged levels
+instrument the differenced regressors. **System GMM** [@blundell1998] augments
+the differenced equations with the level equations, instrumenting them with
+lagged differences; this restores efficiency and reduces finite-sample bias
+when $\alpha$ is close to unity and lagged levels are weak instruments.
+PanelBox implements both, in one-step and two-step variants, and generates the
```
sanhitamj: Again, a short explanation about what are these 2 types of variants wrt the topic would help an outsider.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,569 @@
...
+de facto standard for dynamic panel GMM, and R offers the mature `plm`
+package [@croissant2008plm] for linear panels and specialized packages such as
+`pdynmc` [@fritsch2019pdynmc] for dynamic ones. Python — now dominant in data
+science and increasingly in quantitative social science thanks to
+NumPy [@harris2020numpy], pandas [@mckinney2011pandas], and
+SciPy [@scipy] — has lagged conspicuously behind in panel econometrics.
+
+Existing Python libraries cover only fragments of the workflow.
+`linearmodels` [@linearmodels2019] provides static fixed- and random-effects
+models and some instrumental-variables support but no dynamic GMM; `pyfixest`
```
sanhitamj: For an outsider, explaining the following jargon in a few words would highly increase readability:
static fixed, random=effects models, instrumental-variables support, dynamic GMM

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,569 @@
...
+  of our model-validation practice in credit risk, where econometric models
+  running in production Python systems had to be exported to Stata or R for
+  estimation and validation. We introduce
+  PanelBox, the first all-in-one Python library for panel data econometrics,
+  implementing more than 70 models across 11 families — static linear models,
+  dynamic GMM estimators (Arellano–Bond, Blundell–Bond with the Windmeijer
+  correction), discrete choice, count data, spatial econometrics, stochastic
+  frontier analysis, quantile regression, panel VAR/VECM, censored and
+  selection models, and the first comprehensive Python implementation of panel
+  unit-root and cointegration tests. The library adds 50+ diagnostic tests,
```
sanhitamj, line 18: Is the word cointegration standard in the field? I would find co-integration easier to read. Also needs a short explanation for an outsider.

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,569 @@
...
+pb.comparison_table(
+    [ols_result, fe_result, diffgmm_result, sysgmm_result],
+    labels=["Pooled OLS", "FE", "Diff-GMM", "Sys-GMM"],
+)
+```
+
+(sec:using)=
+## Using and extending PanelBox
+
+PanelBox makes no assumptions about where data come from. Any long-format
```
sanhitamj, line 388: You made me happy by using data as a plural! 💃

apaleyes:
Hi @guhaase! Just wanted to check if you are still planning to update the rest of the paper? Dates say we have about 10 days to submit the last review.

guhaase:
> Hi @guhaase! Just wanted to check if you are still planning to update the rest of the paper? Dates say we have about 10 days to submit the last review.

Hi! Yes, we’re currently working on the improvements based on your guidance. I’m planning to send the updated version tonight.

papers/gustavo_haase/main.md
guhaase: Agreed, fixed. The first paragraph now tells only the origin story. The part about the tooling gap was moved to the second paragraph, where it fits the Stata/R background.

papers/gustavo_haase/main.md
guhaase: Done. We also removed similar emphasis in other parts of the text. Italics were kept only where a term is being defined, like "panel data" and "unit root".

papers/gustavo_haase/main.md
guhaase: Added right after the equation. Entity effects absorb what is specific to each entity and constant over time (for example, a firm's management quality). Time effects absorb shocks common to all entities in a given period (for example, a recession year).

papers/gustavo_haase/main.md
guhaase: Added a short explanation: one-step uses a fixed weighting matrix, two-step re-estimates the weights from the first step residuals to gain efficiency. This also motivates the Windmeijer correction discussed next.

papers/gustavo_haase/main.md
guhaase: Done. Each term now has a short parenthetical explanation in that paragraph, and dynamic GMM points to the section where it is explained in full.

papers/gustavo_haase/main.md
guhaase: Yes, the unhyphenated form is the standard spelling in the current literature and in the tools we compare against (Stata's `xtcointtest`, R's `plm`), so we kept it. Curiously, the original Engle and Granger (1987) paper did use the hyphen. A short plain-language explanation was added in the unit-root section.

papers/gustavo_haase/main.md
guhaase: Glad you noticed. Thanks for the careful read!

guhaase:
@apaleyes  and @sanhitamj fixed in ca9978a1

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
+---
+# Ensure that this title is the same as the one in `myst.yml`
+title: "PanelBox: A Comprehensive Python Library for Panel Data Econometrics"
```
apaleyes, line 3: Few minor suggestions for the abstract:
* Use more of present tense, so "has lacked" -> "lacks", "have had to fall back" -> "have to fall back"
* Swap around sentences "PanelBox grew out..." and "We introduce PanelBox", or maybe skip the former entirely and leave this story to the main text

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+model validation for credit risk, where econometric models estimated on panel
+data must be independently re-estimated, stress-tested, and documented before
+they are allowed into production. Our production systems run on Python; the
+reference implementations of the methods we needed did not. Every validation
+cycle meant exporting data to Stata or R, re-importing results, and
+reconciling discrepancies by hand — slow, error-prone, and impossible to
+automate end to end. What started as an internal toolkit for validating panel
+models in pure Python grew, model family by model family, into PanelBox.
+
+The data at the center of this workflow are *panel data* — repeated
```
apaleyes: "this workflow" -  a bit unclear which workflow. Maybe something like "production econometrics workflow like ours"?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+tooling for these methods is unevenly distributed across computing
+environments. Stata's `xtabond2` command [@roodman2009stata] is the
+de facto standard for dynamic panel GMM, and R offers the mature `plm`
+package [@croissant2008plm] for linear panels and specialized packages such as
+`pdynmc` [@fritsch2019pdynmc] for dynamic ones. Python — now dominant in data
+science and increasingly in quantitative social science thanks to
+NumPy [@harris2020numpy], pandas [@mckinney2011pandas], and
+SciPy [@scipy] — has lagged conspicuously behind in panel econometrics.
+
+Existing Python libraries cover only fragments of the workflow.
```
apaleyes, line 65: This is a really long sentence that is also really easy to break up. I'd suggest doing so

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+visualization, publication-ready reporting, and 103 datasets for immediate
+experimentation. The paper follows the workflow a researcher actually
+traverses — specify and estimate a model, probe its assumptions, and report
+the results — with the depth reserved for the three novel contributions;
+@sec:breadth summarizes the full catalogue, and @sec:using describes how the
+library is used and extended in practice.
+
+## Library design: one workflow from data to report
+
+PanelBox is organized around the estimation workflow rather than around a
```
apaleyes: "PanelBox is organized around the estimation workflow rather than around a catalogue of methods." - this makes sense as a contrast to the first version of the paper draft, but on its own i think can be confusing. How about something like "PanelBox is designed to model a standard estimation workflow common in econometrics."?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+catalogue of methods. The user starts from a pandas `DataFrame` in long
+format; a common data layer wraps it together with entity and time
+identifiers, and everything downstream consumes that layer. Estimators accept
+R-style formulas via `patsy` and return a result object following the
+`statsmodels` conventions — `summary()`, `conf_int()`, `to_latex()`,
```
apaleyes: maybe a bit more accurately "object that follows `statsmodels` conventional API - ...."

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+`to_html()` — so that the same result flows into the shared inference layer
+(11 standard-error types, four bootstrap methods), the diagnostics layer (50+
+tests), and the reporting layer that renders self-contained HTML, LaTeX, or
+Markdown. @fig:architecture summarizes this pipeline.
+
+:::{figure} figures/fig1_architecture.png
+:label: fig:architecture
+PanelBox as a dataflow. A long-format pandas `DataFrame` enters the shared
```
apaleyes: Great improvement on the figure here! One small suggestion would be to make emphasis in "user input" rather than dataformats. For example: "Input data, represented as long-format pandas DataFrame, enters...", and same for the figure title

Some arrows look untidy, but it's really minor detail

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+y_{it} = \mathbf{x}_{it}'\boldsymbol{\beta} + \alpha_i + \lambda_t + \epsilon_{it},
+\qquad i = 1,\dots,N,\; t = 1,\dots,T,
+```
+
+where $\alpha_i$ are entity effects — absorbing everything specific to entity
+$i$ that is constant over time, such as a firm's management quality — and
+$\lambda_t$ are time effects, absorbing shocks common to all entities in
+period $t$, such as a recession year. Fixed
+effects are computed by within-transformation using sparse matrix operations
+and pandas `groupby`, so the estimator scales to large $N$ and supports
```
apaleyes: Any quantification of "large N"? At least orders of magnitude?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+:label: dynamic
+y_{it} = \alpha\, y_{i,t-1} + \mathbf{x}_{it}'\boldsymbol{\beta} + \eta_i + \epsilon_{it}.
+```
+
+That innocent-looking addition breaks both estimators of the previous
+section. The lagged outcome is itself driven by the entity effect $\eta_i$, so
+it is correlated with the composite error: pooled OLS is biased upward, and
+the within estimator downward (the Nickell bias, @nickell1981) — the true
+dynamics lie somewhere in between, and neither estimator can find them. The
+resolution, due to @arellano1991 and @blundell1998, is to transform away
```
apaleyes: "due to" -> "according to"

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+causing severe sample loss.
+
+PanelBox instead validates instruments observation-by-observation. For each
+entity $i$ and period $t$ it (1) identifies the required instrument lags, (2)
+checks their availability, (3) builds an entity-specific instrument set from
+whatever lags exist, and (4) assembles the block-diagonal instrument matrix
+$\mathbf{Z}$ from variable-sized blocks (@fig:unbalanced). An observation is
+kept whenever valid instruments exist, rather than discarding the whole
+entity. On the Arellano–Bond employment data this retains 72% of observations,
+against roughly 40% for `xtabond2` and 0% for naive complete-case analysis,
```
apaleyes: "and 0% for naive complete-case analysis" - this is quite harsh! I assume this is referring to "many practitioners drop every entity with any missing value, causing severe sample loss", which sounds quite bad but not 100% data loss bad. Is this indeed the intended meaning?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+:::{figure} figures/fig2_unbalanced_algorithm.png
+:label: fig:unbalanced
+The unbalanced-panel algorithm: instruments are validated per observation and
+entity-specific instrument sets are assembled into a block-diagonal matrix,
+retaining observations that fixed-template approaches discard.
+:::
+
+## Testing assumptions: unit roots and cointegration
+
+Estimation is only the middle of the workflow; its validity rests on
```
apaleyes: "Estimation is only the middle of the workflow" - a bit awkwardly formulated. Perhaps what was meant is "Real-life econometrics workflow does not end when estimation is done"?

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+first question is whether each variable has a stable long-run level or
+wanders without one (a *unit root*): regressing one wandering series on
+another produces convincing-looking but spurious correlations, so testing for
+unit roots — and, when variables do wander, for a genuine shared long-run
+relationship (*cointegration*) — is the standard safety check before
+estimating relationships among variables such as GDP, prices, or exchange
+rates. This capability is unique to PanelBox among Python panel libraries,
+despite being standard in macroeconomics and finance.
+
+Three panel unit-root tests build on the entity-specific augmented
```
apaleyes: I think this paragraph is missing "PanelBox implements all these tests" message somewhere

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+(sec:breadth)=
+## Breadth: eleven model families
+
+Linear panels are only part of applied practice: economic data routinely
+involve binary outcomes, counts, censoring, spatial dependence, and
+efficiency frontiers. PanelBox covers these in the same interface;
+@tbl:families maps each family to its closest Stata and R equivalents, and we
+highlight below only the capabilities not available elsewhere in Python. The
+complete catalogue, with an executable example notebook per family, is in the
+official documentation.
```
apaleyes: A link to said documentation would be handy here

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+specification, call `fit()`, receive a results object with the same methods
+everywhere. Comparing a fixed-effects model with a system-GMM model is a
+one-line change of class, and `comparison_table` aligns any set of results.
+Within this contract, the supported customization points are the covariance
+estimator and bootstrap scheme (per `fit()` call), the instrument design for
+GMM (`gmm_lags`, `collapse`), arbitrary variable transformations through
+`patsy` formulas, marginal-effects options for nonlinear models, and the
+output format of every report (HTML, LaTeX, Markdown).
+
+Two honest boundaries: there is currently no public API for user-defined
```
apaleyes: This reads really LLMy: "Two honest boundaries: ... ". How about rewriting it a bit to sound more human like? For example "At the moment there are two limitations in PanelBox that we hope to overcome in the future. First .... Second ..."

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+difference GMM at $0.686$ (0.152), confirming the efficiency gain from the
+additional moment conditions. The wage elasticity is consistently negative
+($\hat{\beta}_w \approx -0.57$) and capital enters positively
+($\hat{\beta}_k \approx 0.38$). Diagnostics validate the specification: the
+Hansen $J$ test does not reject ($p = 0.172$), AR(1) rejects as expected
+($p = 0.003$), AR(2) does not ($p = 0.671$), and the 91 instruments stay well
+below the 140 entities. The results match @arellano1991 to within 0.01%.
+
+:::{figure} figures/fig5_coefficient_comparison.png
+:label: fig:coefficients
```
apaleyes, line 481: Confidence intervals (i assume that's what these are) for both GMM methods are so wide! Is that expected?

Also, note that currently legend covers one of them slightly, perhaps it's worth playing with bar widths and legend position to fix it. Doesn't really affect the figure, so it's a minor detail

papers/gustavo_haase/main.md
```diff
@@ -0,0 +1,575 @@
...
+  - −2.93 (0.003)
+  - 0.00
+* - AR(2)
+  - −0.42 (0.671)
+  - −0.42 (0.671)
+  - 0.00
+```
+
+PanelBox is implemented in pure Python on NumPy/SciPy yet remains within
+20–30% of Stata's compiled `Mata` for system-GMM estimation and runs roughly
```
apaleyes: "yet remains within 20–30%" is that in terms of running time? Worth clarifying

apaleyes (commented):
Thanks a lot for the update @guhaase! The paper reads much much better now. There is a steady storyline, implicit lists are replaced with narratives, and features are highlighted where and when appropriate. It is also friendlier to an average reader. I have no more major notes or suggestions, some (honestly nitpicky) comments are below

papers/gustavo_haase/main.md
guhaase: Done. Switched to present tense and removed the origin sentence from the abstract. The story stays in the introduction.

papers/gustavo_haase/main.md
guhaase: Adopted your wording.

papers/gustavo_haase/main.md
guhaase: Split into three sentences.

papers/gustavo_haase/main.md
guhaase: Good catch, the contrast only made sense against the old draft. Now reads "PanelBox is designed to model the standard estimation workflow of applied econometrics."

papers/gustavo_haase/main.md
guhaase: Changed.

papers/gustavo_haase/main.md
guhaase: Thanks. The input box now reads "Input data (long-format pandas DataFrame with entity & time identifiers)", the figure title and caption follow the same wording, and the arrows are now all straight.

papers/gustavo_haase/main.md
guhaase: Added: the estimator scales to panels with thousands of entities and tens of thousands of observations, with a pointer to the benchmarks in Figure 8, which go up to N = 2,500 entities.

papers/gustavo_haase/main.md
guhaase: We went with "proposed by Arellano and Bond and Blundell and Bond", since "according to" would read as if they were reporting the result rather than authoring it.

papers/gustavo_haase/main.md
guhaase: You were right to question this, and thank you: the number was wrong. We recomputed everything on the Arellano and Bond employment data. The panel has 140 firms observed for 7 to 9 years (1,031 observations), and only 14
firms have all 9 years. Complete-case analysis keeps those 14 firms, 126 observations, or 12.2%. PanelBox keeps 751 observations (72.8%), which is every observation that still has a valid instrument after one lag and one
difference are consumed. The text, Figure 2, and the conclusion now carry these numbers. We also removed the unverified xtabond2 retention figure that was in the old draft, and fixed the description of the dataset in the application section, which wrongly called it balanced.

papers/gustavo_haase/main.md
guhaase: Adopted your wording.

papers/gustavo_haase/main.md
guhaase: Added as the opening sentence of the paragraph.

papers/gustavo_haase/main.md
guhaase: Added: https://panelbox.readthedocs.io/ (also in the "Using and extending" section and the conclusion).

papers/gustavo_haase/main.md
guhaase: Fair. Rewritten with your structure.

papers/gustavo_haase/main.md
guhaase: One correction to my earlier reply: the numbers quoted there (0.14 for GMM against 0.013 for OLS) came from the affected results. With the corrected library the standard errors are 0.234 (difference GMM) and 0.194 (system GMM) against 0.008 for OLS, so the point stands, and the intervals in the new Figure 5 are wider still. The figure now also shows the OLS-FE bracket of Bond (2002), which makes the logic of the comparison visible: a consistent estimate should fall inside the band, and both GMM estimates do.

papers/gustavo_haase/main.md
guhaase: Follow-up: that sentence is gone. The Stata timing could not be re-run, so the paragraph now reports measured execution times against plm on the same machine: 1.1 s versus 0.5 s at N = 500 and 4.6 s versus 3.1 s at N = 2,500 for two-step system GMM, T = 10. PanelBox is within a factor of 1.5 to 2 of plm for N >= 500, with a fixed overhead that dominates on small panels. Figure 8 was regenerated from those runs.

guhaase:
@apaleyes  Small follow-up: the 12.2% I gave earlier counted observations kept in the sample; the comparable number is observations actually used in the differenced equation, which is 98 (9.5% of the panel) for complete-case analysis against 751 (72.8%) for PanelBox. Text, Figure 2 and the conclusion now use those.

guhaase:
Hi @apaleyes, one more update, and an important one. While preparing the replication scripts for this revision we found a bug in PanelBox's GMM code. Year dummies were entering the regressions but not the instrument matrix, and the instrument builder was dropping sparsely observed lags. Both affected the estimates reported in the application and validation sections, so the numbers in the previous version were not reproducible with the library as published.

We fixed the library (PanelBox commits 536ca53, 56ee3b5 and c45314a, with regression tests pinned to external reference values), re-ran every result in the paper with the corrected version, and updated the text, the validation table and Figures 2, 5 and 8 in a44d6810. The qualitative story is unchanged: OLS is biased upward, fixed effects downward, and both GMM estimators fall in between. The point estimates did change (system GMM alpha is now 0.563 rather than 0.694, with 18 collapsed instruments rather than 91), and the paper now reports them as they are.

Two consequences for the validation section. First, Stata is proprietary and we could not re-run the full comparison against xtabond2 within the revision window, so my co-authors and I decided to drop the Stata comparison and validate against R's plm, which any reader can reproduce: difference GMM, fixed effects and the IPS unit-root test now match plm to three or more decimals, and LLC within 2%. Second, plm and xtabond2 are known to weight the level equations of system GMM differently, and PanelBox follows xtabond2, so for system GMM we compare against pydynpd, an open Python replica of xtabond2 (Wu, JOSS 2023); agreement is within 1 to 2% on coefficients and exact on the Hansen test. A direct Stata comparison is stated as planned work.

While re-running the Penn World Table example we also noticed that the bundled dataset was a synthetic stand-in, so that paragraph now uses the real PWT 10.01 and reports what it gives, including a Hansen rejection that the earlier text did not show. The timing benchmark was likewise redone against plm on the same machine and the Stata timing claims removed.

We would rather report this ourselves than have a reader find it. Apologies for the extra round, and thanks again for the careful reading that led us to check these numbers in the first place.

apaleyes (approved):
@guhaase what can I say, thank you for the additional diligence and honest reporting objective of your results. Glad the review process was useful, it's a great illustration of why Scipy's open model is such a great idea. I quickly skimmed through the recent changes and everything looks fine, happy to recommend it for proceedings.

nil68657:
Approved with the recommended changes from reviewers

PR #1223, opened 2026-06-14
Paper: Toward Reliable Localization of Free and Open Source Software

leo062644:
This PR adds the SciPy 2026 paper submission for YuLin-Chen.


leo062644:
Hi @freemansw1,

I would like to check on the reviewer assignment for this paper. I noticed that reviewers have not yet been assigned to the PR.

Could you please let me know whether the reviewer assignment is still in progress, or if any additional action is required from me?

Thank you for your help.

freemansw1:
> Hi @freemansw1,
> 
> I would like to check on the reviewer assignment for this paper. I noticed that reviewers have not yet been assigned to the PR.
> 
> Could you please let me know whether the reviewer assignment is still in progress, or if any additional action is required from me?
> 
> Thank you for your help.

Hi @wendy062644 , We're working on this now. Thanks for your patience.

kalyan678:
Hi ,

I would like to review the paper.

papers/yu_lin_chen/main.md
```diff
@@ -0,0 +1,547 @@
...
+| Condition | Masking | Glossary hints | Candidates | Purpose |
+|---|---:|---:|---:|---|
+| C0 (Direct) | no | no | 1 | Direct translation without masking |
+| C1 (Complete workflow) | yes | yes | 3 | Production condition |
+| C2 (Direct + glossary) | no | yes | 3 | Tests glossary/candidates without masking |
+| C3 (Masked without glossary) | yes | no | 3 | Tests masking without glossary hints |
+| C4 (Single candidate)  | yes | yes | 1 | Tests the value of generating multiple candidates |
+```
+
+C0--C4 are experimental conditions for the same workflow over a fixed stratified subset of 3,000 message segments. Our evaluation on the complete corpus focuses only on C1, however, as it performs better than others over the fixed subset of 3,000 messages. Scaling up all experimental conditions to the complete corpus would primarily increase cost and review burden rather than change our deployment priority: whether the selected production workflow will remain structurally safe in order to release to the QGIS Traditional Chinese user community.
```
freemansw1: general comment- it would be better to replace -- with - here.

freemansw1 (approved):
I have one minor formatting comment, but otherwise @wendy062644 this is in excellent shape! Thank you for this fantastic contribution!

papers/yu_lin_chen/main.md
leo062644: Thank you! I've made the suggested change.

papers/yu_lin_chen/main.md
```diff
@@ -0,0 +1,547 @@
...
+        brace_placeholders: ["{feature_id}"]
+       }
+)
+```
+
+The splitter then produces this alternating sequence for $s$:
+
+```text
+text_0   = "Open "
+token_1 = "&A"
```
kalyan678: With token_1="&A" and text_1="Attribute Table", the stated decomposition does not reconstruct the original source Open &Attribute Table; it would produce Open &AAttribute Table. Since the mnemonic is later moved to (&A), would it be clearer to model mnemonic information separately from the literal protected-token sequence, or adjust the formal definition/example? This matters because the current example conflicts with the equation defining the source as an alternating token/text sequence. Let me know what do you think on this?

papers/yu_lin_chen/main.md
```diff
@@ -0,0 +1,547 @@
...
+The complete-corpus result confirms that the subset finding scales up to the complete QGIS Qt source file: all three backends produce zero observed structural failures under C1 for the token classes we have checked. The local TAIDE backend is structurally safe under the workflow, but its rule-based content completeness QA is much lower. This indicates that the local backend is a viable offline deployment path for structural preservation, while requiring substantially more review efforts than the other two cloud backends.
+
+---
+
+(sec-reproducibility-artifact)=
+## Reproducibility and Artifact Availability
+
+The accompanying reproducibility artifact is available in the [qgis-llm-localization-workflow repository](https://github.com/wendy062644/qgis-llm-localization-workflow). The current edition of the artifact contains the scripts, configurations, archived outputs, and evaluation reports needed to reproduce this paper's analyses.
+
+```{table} Reproducibility levels.
```
kalyan678, line 505: The reproducibility section mentions a larger extended archive containing the 3,000-segment and 28,924-message outputs. Could you add a direct persistent link/version/DOI to that archive? The currently linked public repository appears to contain the 100-segment mini reproduction by default, so a reviewer may not know where to obtain the full archived outputs used for the paper table

kalyan678 (commented):
Please look into these two comments and let me know if we are good to make those changes. Otherwise, it is really great paper. Thoroughly enjoyed reviewing it!

papers/yu_lin_chen/main.md
leo062644: Thanks! We added a direct persistent Depositar link to the complete 3,000-segment and 28,924-message experimental outputs.

papers/yu_lin_chen/main.md
leo062644: Thanks for catching this. We revised the example and formalization so that Qt mnemonics are handled separately as metadata rather than as literal protected tokens.

kalyan678 (approved):
Thanks for making the changes. Looks good to me.

Wishing you the best on your paper @leo062644 !!

piyushbag (approved):

piyushbag (commented):
Left a few notes on specific spots. Not asking for another revision. Accept as-is.

papers/yu_lin_chen/main.md
```diff
@@ -0,0 +1,564 @@
...
+$$
+
+Here, $\hat{t}_{u,j}$ is the $j$-th candidate for translation unit $u$. In this strict masked path, every emitted validation issue blocks automatic acceptance.
+
+If no generated candidate passes strict validation in a masked condition, the source string is written as a source-preserving fallback and the translation unit is logged as review-required; such a fallback is not counted as completed localization. In the current implementation, this status is recorded in the external run log rather than in the `.ts` `unfinished` attribute, so the log must accompany the generated artifact during review. In the direct ablations C0 and C2, the selected raw candidate and its detected issues are retained so that structural failures remain observable. Review state is not used as a key quality metric because it may have several operational causes: no valid candidate, backend exception, malformed JSON, or intentionally conservative rejection. Detailed review-state counts are reported in the accompanying reproducibility artifact. This paper focuses on structural failure rate, deterministic rule-based QA diagnostics, and MQM semantic quality.
```
piyushbag: If review-required only lives in the external run log and not in the `.ts` unfinished attribute, then someone who just takes the generated `.ts` will treat those fallback rows as finished translations. Would it be cleaner to write unfinished back onto those rows, or is the log meant to always travel with the file? This matters because the 0.00% structure number includes those fallbacks. Let me know what you think.

papers/yu_lin_chen/main.md
```diff
@@ -0,0 +1,564 @@
...
+| C2 (Direct + glossary) | no | yes | 3 | Tests glossary/candidates without masking |
+| C3 (Masked without glossary) | yes | no | 3 | Tests masking without glossary hints |
+| C4 (Single candidate)  | yes | yes | 1 | Tests the value of generating multiple candidates |
+```
+
+C0--C4 are experimental conditions for the same workflow over a fixed stratified subset of 3,000 message segments. C0 and C2 disable the structural hard lock and retain selected raw outputs so that direct-generation failures remain measurable. C1, C3, and C4 use masking, rule-based reassembly, strict validation, and a source-preserving fallback for rejected candidates. The conditions therefore compare complete operating configurations rather than isolating masking from the fallback policy. Our complete-corpus evaluation focuses on C1 because it is the selected production configuration, combining masking, glossary hints, and three candidates. Scaling up all experimental conditions would increase cost and review burden without being necessary to test whether this selected configuration produces structurally valid final artifacts for the QGIS Traditional Chinese localization workflow.
```
piyushbag: Makes sense not to run every condition on all 28k messages. I was more wondering why C3 didn't at least stay in the main subset table. You say it doesn't show a distinct structural failure pattern, which is useful, but then the glossary's effect on Rule QA / MQM ends up in the artifact. Not blocking, just harder to see whether glossary hints actually matter for picking C1.

papers/yu_lin_chen/main.md
```diff
@@ -0,0 +1,564 @@
...
+
+### MQM-style semantic review metrics
+
+Semantic quality is evaluated with MQM-style scoring. For segment $u$, let $P_u$ be the set of MQM penalties assigned by the judge. The weighted MQM error rate is:
+
+$$
+MQM\text{-}ER = 1000 \times \frac{\sum_{u=1}^N\sum_{p \in P_u} p}{\sum_{u=1}^N |s_u|}.
+$$
+
+Lower values indicate fewer weighted semantic errors per 1,000 source characters. The fixed MQM penalties are 0, 1, 5, and 25 for neutral, minor, major, and critical findings, respectively. The auxiliary MQM score on a 0--100 scale is reported for readability; the error rate is the main semantic metric. For each backend-condition pair, Grok 4.3 judged three random samples of 200 segments, yielding 600 judgments. The same sampled segment positions were used across backends and conditions within each repeat. The reported approximate 95% half-width is $1.96s/\sqrt{3}$, where $s$ is the standard deviation of the three run-level MQM-ER values; with only three repeats, this interval is descriptive rather than a precise population estimate.
```
piyushbag: Grok is a translation backend here and also the only MQM judge, so it scores its own outputs. You mention self-preference later. With only 3 repeats of 200 the interval on Grok C1 is already 9.005 ± 2.807, so I'm not sure how much we should read into the MQM ranking. The structural checks feel more solid. Let me know if I'm reading the judge setup right.

papers/yu_lin_chen/main.md
```diff
@@ -0,0 +1,564 @@
...
+:name: tab-complete-corpus-c1
+
+| Backend | Messages checked | Structure failure % ↓ | Structure score ↑ | Rule QA ↑ | Avg. valid candidates | Possibly untranslated |
+|---|---:|---:|---:|---:|---:|---:|
+| Grok 4.3 | 28,924 | 0.00 | 100.000 | 92.11 | 2.938 | 1,394 (4.82%) |
+| Gemini 3.1 Flash-Lite | 28,924 | 0.00 | 100.000 | 90.09 | 2.935 | 1,825 (6.31%) |
+| TAIDE 12B | 28,924 | 0.00 | 100.000 | 65.13 | 2.052 | 6,674 (23.07%) |
+```
+
+The zero rates in {numref}`tab-complete-corpus-c1` describe the final `.ts` artifacts and include source-preserving fallbacks flagged for review: 591 of 28,924 rows for Grok (2.04%), 1,396 for Gemini (4.83%), and 7,265 for TAIDE (25.12%). These rows preserve structure but remain incomplete localization rather than accepted translated content.
```
piyushbag, line 494: Just checking I have this right. The 0.00% on the full corpus includes the fallback rows (591 / 1396 / 7265). For TAIDE that's about 25% still untranslated and queued, so zero structural failure and done localizing are not the same thing. Would it help to put that in the table caption so people don't just skim the 0.00? Let me know.

papers/yu_lin_chen/main.md
leo062644: Yes, that's correct. In the current implementation, review-required fallback rows are tracked in the external run log rather than through the `.ts` `unfinished` attribute, so the log is intended to accompany the generated `.ts` artifact. I agree that marking these fallback rows as `unfinished` directly in the `.ts` would be cleaner, and this would be a useful improvement for a future update.

papers/yu_lin_chen/main.md
leo062644: Thanks, that's a good point. I have added C3 back to the main subset table so that the effect of glossary hints on Rule QA and MQM can be seen directly alongside C1.

papers/yu_lin_chen/main.md
leo062644: Yes, that's the correct interpretation. Grok 4.3 is both one of the translation backends and the sole MQM judge, including for its own outputs. Because of that, and given the limited number of repeats, I treat the MQM results as descriptive rather than as a strong model ranking. The structural checks are the more robust part of the evaluation.

papers/yu_lin_chen/main.md
leo062644: Yes, exactly. The 0.00% structural failure rate includes the source-preserving fallback rows, so it should not be interpreted as complete localization. I have added this clarification to the full-corpus table caption to make that distinction more explicit.

piyushbag: Thanks for clarifying and updating the caption. That addresses my concern.

PR #1225, opened 2026-06-14
Paper: Automated Data Enrichment for Police Accountability: Where Agentic Judgment Earns Its Place

hongsupshin:
First-draft submission for the SciPy 2026 proceedings.

**Paper:** Where Agentic Judgment Earns Its Place in High-Stakes Data Enrichment
**Author:** Hongsup Shin (Texas Justice Initiative)

A pipeline built on LangGraph and the Anthropic API that recovers missing fields in the Texas Justice Initiative's officer-involved-shooting databases, pairing a deterministic core with three bounded LLM judges of graduated authority (block / null / advise), each gated by an offline, reward-hacking-resistant evaluation. This is a first draft and will be refined during review.


spraveengupta:
HI @hongsupshin - I am the assigned reviewer. I am happy to review the paper and share my feedback. 
Thanks
Praveen

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+
+  We study this on the Texas Justice Initiative's police shooting databases, where nearly two thousand records are missing the weapon, the subject's race, or the outcome, whose fields volunteers typically recover by hand, fifteen to thirty minutes each. Our LangGraph pipeline, deterministic in its control flow, searches, validates, extracts, and escalates hard cases to a human. It completes 92% of officer and 70% of civilian records and invents zero facts across twenty fabricated incidents. The recovery itself came from a deterministic prompt fix without any agent. An autonomous agent pointed at the same fabricated incidents, with more freedom, commits a wrong-article fabrication the pipeline escalates.
```
spraveengupta: This statement would benefit from clarification. The phrase "invents zero facts across twenty fabricated incidents" may be misread as meaning that all 20 incidents were fabricated. 

If 20 refers to the sample size, it could explicitly be stated as the following "across a sample of 20 test incidents." Clarification on the “invents zero facts” can be helpful as it means fabricated names produced by the source LLM were rejected by the judge, rather than accepted into the final records.

Suggested rewrite - It completes 92% of officer records and 70% of civilian records, while introducing no unsupported facts across a sample of 20 test incidents.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+
+  If the deterministic core does the recovering, the agentic layer earns its place by making those recovered values trustworthy. Agency lives only in this thin judgment layer above extraction, and its components act in one of two ways. One acts on the pipeline's control flow: a relevance judge reads the retrieved articles and, when none actually report this incident, routes the record to a human instead of completing it. The other two pass judgment on what extraction produced: one deletes a value the source never states, and the other explains to the reviewer why the sources disagree on a value. Extraction calls an LLM too, but because it only proposes values for these judges to rule on, we do not count it as agentic. Every judge had to clear a reward-hacking-resistant evaluation gate before it shipped. The main contribution of this paper is a discipline, an "earn-it" protocol, for drawing the line between what a high-stakes pipeline should settle deterministically and where it is worth granting agentic judgment.
```
spraveengupta: The enumeration in this paragraph could be clearer. The paragraph first says the components act in "one of two ways," then describes "one" component and later refers to "the other two," which may confuse readers about whether the authors are describing two categories or three components. It would help to explicitly separate the two categories: one control-flow judge that routes records to human review, and two output-evaluation judges that assess the extracted values.

Also, I thought there are "three evaluation judges" based on this statement - "whereas three small bounded judges hold authority to veto, null, or flag" in p2.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+
+1. **Agentic judgment belongs to the evidence layer, above extraction.** A deterministic prompt change produced the single largest gain in this project; the judges add faithfulness and leave recall to retrieval.
```
spraveengupta: Showing details around how overall output is being "deterministic" can provide extra clarity on the methodology as LLMs are generally "probabilistic" in nature.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+
+## Results
+
+### Held-out accuracy
+
+A record *completes* when it reaches the Complete node with at least one field extracted; otherwise it escalates to human review. On the held-out set (100 records per dataset; {ref}`tbl:holdout`, {ref}`fig:overview`), the pipeline completes 70% of civilian records (95% Wilson interval 60–78%) and 92% of officer records (85–96%). Among extracted values, aggregate exact-match precision is 77% for civilians and 71% for officers, rising to 89% and 86% under fuzzy match. No record escalates on a conflict alone, since partial completion commits the agreed fields and routes only the contested one to review.
```
spraveengupta, line 345: For the cases that were escalated to human review, it would be helpful to report what the system’s precision would have been if those cases had not been escalated (will this be the same as autonomous agent result?). This would clarify how much the escalation step contributes to the final precision.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+  - civilians_shot
+  - officers_shot
+* - Completion rate
+  - 70% (70/100) [60–78]
+  - 92% (92/100) [85–96]
+* - Escalation rate
+  - 30% (30/100)
+  - 8% (8/100)
+* - Aggregate exact
+  - 77% (210/272) [72–82]
```
spraveengupta, line 362: Explicit clarity on why this is higher than 100 can be helpful.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+* - time_of_day
+  - 92
+  - 39
+  - 82%
+  - 82%
+* - location_detail
+  - 100
+  - 44
+  - 16%
+  - 91%
```
spraveengupta, line 409: This is the only place where exact and fuzzy results differs. what is the ground truth for location detail? would it have been better to extract the specific location from LLM using structured output for comparison?

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+
+### Adversarial robustness
+
+The adversarial suite tests whether the system refuses to answer when it should. On its 20 fabricated incidents, **all 20 escalated; none completed; zero fabricated names appeared in any extracted field.** Seventeen escalated at search, two at the relevance judge, and one as insufficient sources ({ref}`tbl:outcomes`). Because seventeen were stopped at search before any extraction ran, the agentic defenses were exercised on only the two records that retrieved real articles: the deterministic front-end does much of the safety work, with the agentic layer as the deeper net that caught every case in this small but deliberately hard probe.
```
spraveengupta: consider changing "names" -> "data" for extreme clarity.  The current phrasing confuses that there is a "name" field where we do not have any fabrications with the proposed framework.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+
+### An autonomous agent fabricates where the pipeline escalates
+
+The cheaper-model probe tested a smaller model on the same suite; this baseline tests dropping the workflow altogether. We ran the autonomous-agent baseline on the same twenty incidents and scored it identically ({ref}`tbl:baseline`). The pipeline completed none of the twenty and escalated each with a reason a reviewer can act on. The agent completed one incident, 99913, in all three runs, each time committing six fabricated fields and marking the record done with no signal that anything was wrong; it declined the other nineteen.
+
+:::{list-table} The autonomous-agent baseline against the shipped pipeline on the 20-incident adversarial probe. Agent figures are per-run means over three runs. Cost is for this probe, where 18 of 20 incidents escalate before any extraction runs; the pipeline's holdout average, which includes extraction, is higher (\$0.20).
+:label: tbl:baseline
+:header-rows: 1
+
+* - Metric (20-incident probe)
```
spraveengupta, line 514: Curious to know why we are doing the comparison with baseline using a different dataset? (not using the same dataset with 100 records that was used earlier).  Adding that rationale can make the narrative unambiguous.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,658 @@
...
+::::
+
+Autonomy was also more expensive. The agent ran 14.9 Tavily searches per incident to the pipeline's 2.8 and added open-web fetches the pipeline does not make, costing about \$1.00 per incident against the pipeline's \$0.05 on this probe, and the extra effort bought no safety: the additional searching on 99913 surfaced more of the adjacent Austin coverage the agent drew on. Two qualifications matter. The agent declined nineteen of twenty, so a careful prompt does much of the work, and the judge closes the remaining trap where adjacent coverage let the agent convince itself. One of those declines was also a near-miss: on a San Antonio trap the agent chased a real overnight-shooting story until the turn cap stopped it, so the single committed fabrication is a floor. The cheaper-model and autonomous-agent probes point the same way, with the adversarial suite and the gate catching the tempting shortcut before it ships.
```
spraveengupta: I am curious why the agent performs a higher number of searches. Could this be reduced through prompt optimization or other changes to the search strategy?

spraveengupta (commented):
Overall, this is a strong submission. I have a few comments that I believe could help improve the clarity and presentation of the work.

chongshenng:
Hello @hongsupshin! Great to see a submission from you -- as you can already tell, I will serve as the editor for your submission. Please reach out to me if any assistance needed 😊.

dinkellabs:
Hi @hongsupshin,
I am the second assigned reviewer for your submission. I will have a read tonight and share my thoughts with you this week. The title definitely looks interesting to me.
Cheers,
Dennis

papers/hongsup_shin/main.md
hongsupshin: Thanks for the comment! The fact that this sentence supports two readings, which makes it sound confusing, means it needs the fix, though I want to flag that the intended meaning is closer to the opposite of the suggested rewrite. The 20 incidents are not a sample of real test records. All 20 are incidents we fabricated (invented names attached to real Texas cities and dates) as an adversarial probe, described under Evaluation design. For a fabricated incident the only correct behavior is to complete nothing, so "invents zero facts" means the pipeline committed no values for any of the twenty and escalated every one to human review, not that a judge filtered fabricated names out of otherwise-accepted records. I've rewritten the sentence to make both points explicit: 

"On an adversarial probe of twenty incidents we fabricated, where the only correct behavior is to refuse, it completes none and asserts nothing, escalating every record to a human."

papers/hongsup_shin/main.md
hongsupshin: Agreed. The paragraph asked the reader to compute "three" from "one plus the other two." I've restated it to give the count and the categories upfront: "three bounded judges in two categories," then one sentence per category, naming each judge (relevance judge on control flow, race verifier and conflict annotator on extracted values). Naming all three in the abstract also ties it directly to the Introduction's "three small bounded judges," which, as you inferred, is the same set. Thanks for pointing this out!

papers/hongsup_shin/main.md
hongsupshin: Agreed here as well. The paper was using **deterministic** for two different things and only defined one of them. I've made the distinction explicit in two places. In Methods, the design-choices paragraph now states directly that deterministic describes the _control flow_, not the model outputs: every LLM call remains probabilistic (run-to-run variance is acknowledged in Limitations and in the reproducibility statement), but which nodes run, in what order, and what happens on each outcome is fixed code that no model can redirect. And in this numbered claim, "a deterministic prompt change" was misleading shorthand for "a change in the fixed, non-agentic layer"; it now reads "a change to the fixed extraction prompt, with no agent involved," which claims nothing about output determinism.

papers/hongsup_shin/main.md
hongsupshin: Good question. I've added a paragraph in Results making it explicit. For most escalations no counterfactual precision exists: 22 of 30 civilian and 4 of 8 officer escalations are retrieval gaps where no article survived validation, so an _un-escalated_ pipeline would still have had nothing to extract from. The counterfactual is real only for the relevance-judge vetoes (7 civilian, 3 officer): without the judge those records complete (officer completion 95% rather than 92%), and our audit of every veto found the retrieved articles reported a different incident, so those completions would have committed fields describing the wrong event. 

On your question in the parenthesis, yes, that is the same wrong-article failure the autonomous agent commits on incident 99913 ({ref}fig:agent-trace), except on real holdout records the wrongly committed values would be facts about a different shooting rather than fabricated ones.

papers/hongsup_shin/main.md
hongsupshin: Added. In the original table, the first two rows are per record, the aggregate rows are per extracted field value, and each completed record contributes several fields (70 completed civilian records yield 272 scored values, where a value is scored only if its ground truth is non-null). The caption now states this explicitly.

papers/hongsup_shin/main.md
hongsupshin: The ground truth is the database's city (with county as a fallback), while the pipeline extracts street-level detail such as "100 block of Couch Court, Springtown, Parker County" against a ground truth of "Springtown". So the exact match fails on the granularity gap by construction, and the fuzzy approach is the meaningful one for this field. 

Regarding the structured output, the extraction already returns structured output, so the mismatch is one of granularity and purpose rather than format. Constraining the field to city-level would raise exact match but would discard the enrichment value (**the street-level detail is what the database lacks**) and would be circular (i.e., the city is part of the pipeline's own search anchor). 

I think your question actually exposed two imprecise points, so I've applied the fix. The Results sentence had the "street-level ground truth", and the Evaluation design section overstated location as a "hidden" field (the city anchors search and validation), so the fix now resolves this and the manuscript should describe the location score as a consistency check rather than a blind holdout.

papers/hongsup_shin/main.md
hongsupshin: I acknowledge the suggestion, but I kept one nuanced bit. Since none of the twenty completed, the committed-data claim is absolute, and the text now says so plainly ("the pipeline committed no data at all"). I kept a sharpened version of the names claim as a second sentence because the planted names do specific work. They are the only invented content in each probe record (the cities and dates are real by design, so genuine articles can collide with them), which makes their absence from every extraction the strictest fabrication check the probe supports. The rewrite states that role explicitly instead of leaving "names" to read as an arbitrary field.

papers/hongsup_shin/main.md
hongsupshin: Ah yes, the rationale here was not explicit in the manuscript. I've added it where the baseline is introduced. To be specific, I added three reasons:
1. The baseline exists to test the safety claim, not the accuracy claim: the paper's argument concerns fabrication under uncertainty, and the probe is built to measure that, whereas a holdout comparison would grade the agent on _accuracy_ while leaving refusal untested. 
2. Ground truth on the probe is absolute. Every incident is fabricated, so any _completion_ is a fabrication, with no fuzzy field-level scoring involved. 
3. Cost. At roughly $1.00 per incident for the agent, three runs over the 100-record holdout would spend five times as much to answer the wrong question. 

One clarification on "different dataset": both systems ran the identical twenty incidents (the pipeline's side is the adversarial column of Table 3), so the comparison itself is matched. It only differs from the holdout used for the accuracy results.

papers/hongsup_shin/main.md
hongsupshin: Excellent question! I think the higher count is actually structural rather than a tuning artifact. The pipeline's searches are bounded by code. We have the fixed three-rung ladder, which caps the number of searches, (hence 2.8 per incident), but the agent decides for itself when it has searched enough. On a fabricated incident no query can succeed, so it just keeps reformulating, and the search volume is essentially its unresolved uncertainty made visible. 

Prompting could certainly lower the count, but I was intentional about that because that path illustrates the paper's central distinction. A search budget stated in prose is an instruction of the same kind the agent already overrode on 99913, and the extra searching bought no safety. Enforcing the budget reliably means _moving it into code_, which recreates the workflow this baseline deliberately removed. I think this would be a good point to add in the discussion, so I've added it.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,660 @@
...
+  ],
+  "outcome_summary": "Escalated incident 75: no retrieved article reports this 2018 incident"
+}
+```
+
+The two terminal outputs (field names verbatim from the pipeline's JSON schema; values abridged). The completion (top) logs each field's confidence and sources; the escalation (bottom) is the relevance judge blocking an article that passes every rule-based check but reports a different (2013) case.
+::::
+
+Two design choices are deliberate. First, **the orchestration is deterministic**: the Coordinator is `if`/`match` logic over state fields and the retry ladder is a fixed list. *Deterministic* here describes the control flow, not the model outputs: every LLM call remains probabilistic (a variance we return to in Limitations), but which nodes run, in what order, and on which outcomes is fixed code that no model can redirect. We chose this over an LLM router because, in a high-stakes domain, predictable control flow is itself a safety property. Second, the system is **human-in-the-loop (HITL) by construction**. It never writes back to the source database, and escalation is a designed terminal outcome that hands the record to a human.
```
dennisbakhuis: "Probabilistic" isn't quite right. At temperature 0 sampling is
deterministic in principle; please state the decoding config
(temperature/seed). The variance you cite in Limitations comes largely
from API-side nondeterminism (batching, floating-point effects, model
updates), not sampling randomness, which also explains why the eval
reports aren't bit-for-bit reproducible even though a seed would fix
truly probabilistic calls. Consider "nondeterministic and unverified"
instead: correctness, not stochasticity, is what the architecture
defends against.

Minor: "no model can redirect" sits uneasily next to the
Introduction, where the relevance judge's veto is acknowledged as a
model-attributable routing outcome. Carry that caveat over here.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,660 @@
...
+
+### Adversarial robustness
+
+The adversarial suite tests whether the system refuses to answer when it should. On its 20 fabricated incidents, **all 20 escalated and none completed, so the pipeline committed no data at all.** The planted names, the only invented content in each probe record (the cities and dates are real), are the sharpest tracer of fabrication, and they never appeared in any extraction. Seventeen escalated at search, two at the relevance judge, and one as insufficient sources ({ref}`tbl:outcomes`): the deterministic front-end does much of the safety work, and the agentic defenses, exercised only on the two records that retrieved real articles, caught every case in this small but deliberately hard probe.
```
dennisbakhuis: This result is based on only 20 incidents. Using your own rule of reading rates through a 95% Wilson interval, 0/20 still leaves an upper bound of about 16%. So it's really "no fabrication seen in this small probe", not "fabricated nothing". This section hedges nicely, but the abstract and conclusions don't. You're careful about this for the agent baseline ("the single committed fabrication is a floor"), so I'd just apply the same care here, or grow the probe.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,660 @@
...
+4. **A consensus resolver** (`field_normalizers.py`) commits an `outcome` or `time_of_day` value when at least two articles agree on the same canonical form; relatedly, a record can **complete partially**, keeping the fields the sources agree on while routing only a genuinely conflicting field to review.
+5. **A race taxonomy** (`race_taxonomy.py`) maps race terms to TJI's buckets for matching while preserving the raw value and flagging divergences for review.
+
+{ref}`tbl:examples` shows several of these behaviors on concrete cases.
+
+### Agentic judges with graduated authority
+
+Only after extraction do three LLM judges reason about what rules cannot, each granted authority **calibrated to the stakes** of its decision, from block for the highest-confidence check down to advise for the fuzziest. Every judge is deliberately bounded: a single structured-output call (no loops, no tools), *fail-open* (an error is logged and the pipeline proceeds as if the judge had not run), and read-only with respect to the database. The worst case any judge can produce is one extra human review, never a corrupted record.
```
dennisbakhuis: Minor comment, but "fail-open" here means the judge is skipped and the process just continues. If the relevance judge doesn't run, what happens next? Even though a human reviews the output, the completed record looks like any other, so the reviewer can't tell the check was skipped. Did this happen during the eval runs?

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,660 @@
...
+  - Declined (cost scales with k; variance already absorbed by the prompt fix)
+```
+
+### Evaluation design
+
+Two measurement signals feed the multi-objective gate ({ref}`tbl:gate`): an accuracy dev set measures field-level correctness, and an *adversarial suite* measures whether the pipeline refuses when it should.
+
+The accuracy comes from fields that already exist in the database but are *hidden* from the pipeline during enrichment (age, race, weapon, location, time, outcome). The pipeline never sees them; we compare its independently extracted values against them afterward, a natural held-out test. Location is the one partial exception: the pipeline necessarily knows the city, which anchors search and validation, so the `location_detail` comparison asks whether the street-level string it extracts names the right city, a consistency check rather than a blind holdout. We draw a deterministic, year-stratified sample of 100 records per dataset, with field-appropriate comparators: exact match for age, race, and outcome; category normalization for weapon; fuzzy string match (`rapidfuzz` [@rapidfuzz], threshold 80) for location; and a ±2-hour comparison for time of day.
```
dennisbakhuis, line 333: Here you use the database values as ground truth, however before you stated that for example race is "frequently mischaracterized" and that the "harm" field is "sometimes stale".  Here it seems that you say that these errors are pipeline errors, while it could also be database errors as you stated before.

It would be worth spot-checking a few disagreements against the source articles and being explicit about this circularity, since some "errors" may actually be the pipeline correcting the database, which would be a selling point rather than a flaw.

papers/hongsup_shin/main.md
```diff
@@ -0,0 +1,660 @@
...
+  - 11%
+  - 97%
+* - civilian_race
+  - 99
+  - 9
+  - 67%
+  - 67%
+```
+
+Denominators differ by field because each is scored only where ground truth exists. Small cells warrant caution: `civilian_race` accuracy is over the 11 values the verifier committed, and its 95% Wilson interval (62–98%) is wide enough to span the 65% gate-off figure reported below. We therefore read the 65%→91% change as the verifier declining to assert unstated races, not a real accuracy gain. Each extracted value also carries a self-reported confidence label, and it is usefully calibrated: high-confidence extractions are markedly more accurate than medium-confidence ones (roughly 93% vs 68% exact on civilians, 86% vs 54% on officers), a usable triage signal for a reviewer.
```
dennisbakhuis: Minor, but when seeing numbers like "roughly 93% vs 68%" I am always curious about the counts.

dennisbakhuis:
Hi @hongsupshin,

I enjoyed reading your paper and I agree with @spraveengupta that this is a strong submission.

The core idea, a discipline for deciding where agentic judgment earns its way into a high-stakes pipeline, is well argued and well evidenced, and the "earn-it" framing with the table of rejected ideas is something I expect readers to reuse. The twenty fabricated incidents are a really nice piece of evaluation design: a negative control most LLM pipeline papers simply don't have, made concrete by running the autonomous agent on the same probe and releasing the transcripts. My comments are mostly about precision of claims rather than the design or the results, and none of them should be hard to address.

papers/hongsup_shin/main.md
hongsupshin: Good question, To answer directly, the pipeline leaves temperature unset on the ChatAnthropic client, so calls use the Anthropic API default of 1.0, and there is no seed (the Messages API doesn't expose one). The calls are genuinely sampled, so "probabilistic" is meant literally rather than loosely.

That said, your deeper point is right. Sampling isn't the only source of run-to-run variance; API-side nondeterminism like batching, floating-point compounds it, and we can't verify that easily, so the eval reports aren't bit-for-bit reproducible. I've added the decoding config to the Methods passage and reworded it so it names both sources and makes explicit that the architecture is built to catch wrong values, not to eliminate variance.

On the minor point, agreed, that was an inconsistency. I've clarified that the control flow is model-independent except for the one model-attributable route, the relevance judge's veto, which the Coordinator reads to escalate.

papers/hongsup_shin/main.md
hongsupshin: You're right, the honest claim is "no fabrication observed on a small, hard probe," not "fabricated nothing." I've applied the fix in two places: added the explicit Wilson bound to the Results paragraph, and scoped the abstract and conclusion so neither reads as a guarantee. I've taken the wording route rather than growing the probe for this revision, and I note a larger adversarial suite as the way to tighten that bound.

papers/hongsup_shin/main.md
hongsupshin: I think your point about observability gap is real. A completed record shows a judge's veto (as an escalation reason) but not whether the judge actually ran, so a fail-open looks like a normal completion.

On whether it happened in these runs, the judges' effects are all present in the saved reports, and the relevance judge's vetoes appear as `irrelevant_sources` escalations (7 civilian, 3 officer), and the race verifier's output is the 11 committed civilian races. A frequent fail-open would have erased these, so the judges were clearly running. I can't certify a hard zero, though because the harness logs each fail-open as a *run-time warning* rather than to the report, so the exact count isn't recoverable from the saved artifacts. To address this, I've added a sentence to Methods making this explicit.

papers/hongsup_shin/main.md
hongsupshin: Yes, I agree that it's a real circularity, and TJI actually acknowledges this (rarely but we sometimes find errors in our DB). So a mismatch scored against the pipeline can be a database error. Among the examples I examined, most of the weapon and time mismatches are taxonomy or comparator artifacts ("rifle" vs. "firearm", "just after midnight" vs. 00:07, etc.), several ages are off by one, and the residual conflicts land on race, which TJI warns as often mischaracterized. To reflect your point, I've reframed the comparison in Evaluation design: the database is an imperfect reference, so counting every mismatch against the pipeline makes the reported accuracy a conservative lower bound, and I flag the one committed civilian-race disagreement (Black where the database says Hispanic) as a likely database error than possibly an extraction error.

hongsupshin:
@spraveengupta @dennisbakhuis Thank you so much for the thoughtful and thorough review. I have addressed your comments and updated the manuscript. Thanks @chongshenng for assigning these reviewers!

spraveengupta:
The manuscript looks great with all the revisions.  @hongsupshin 
Thanks for the responses and capturing nuances called out in the comments above. Thank you.

chongshenng:
Thanks @hongsupshin for efficiently and carefully addressing the reviewer comments! Thank you also @spraveengupta and @dennisbakhuis for reviewing this submission.

spraveengupta (approved):
Thanks for addressing all the comments.  Approving the request.

dennisbakhuis (approved):

dennisbakhuis (approved):

chongshenng (approved):

PR #1227, opened 2026-06-15
Paper: Feel the model: Sensory Transduction of Neural Activations as a Human-in-the-Loop Safety

virajsharma2000:
Submitting corresponding paper for Scipy 2026 accepted Poster Presentation - poster titled "XAI - Feel the Model"

Hi @virajsharma2000 , I am Sanhita Joshi (@sanhitamj) from SciPy Proceedings. I will serve as the editor on this paper. Let me know if I can help in any way.


papers/viraj_sharma/main.md
```diff
@@ -0,0 +1,244 @@
+---
+# Ensure that this title is the same as the one in `myst.yml`
+title: "Feel the model: Sensory Transduction of Neural Activations as a Human-in-the-Loop Safety"
+abstract: |
+  Current mechanistic interpretability methods - sparse autoencoders, activation classifiers, and natural language autoencoder, share a basic assumption: that the safety-relevant content of a model's internal representations can be faithfully represented through human-understandable text. I find this to be limiting, as it is not merely lossy, but directionally biased as we are discarding signal that does not conform to language categories. Using GPT-2 and published sparse autoencoders, I compare classifiers trained directly on model activations with classifiers trained on SAE-derived features. I will also show that cases of classifier disagreement, where SAE features predict one behavioral class while raw activations predict another, correlate with ground-truth misclassification in the SAE-based classifier, identifying a concrete failure mode I term interpretive displacement, wherein the text label assigned to an activation pattern actively misdirects human interpretibility.
```
sanhitamj: Explain what does SAE stand for for the first use in abstract, and then again the first occurence in the rest of the text.

virajsharma2000: Hello @sanhitamj  - thank you for the review. I have attempted to explain acronym before its used in the abstract. please check if it is acceptable.

I ran into curvenote issues for mp4 video display and thumbnail, so I had to fix those also.

sanhitamj: Tagging @fwkoch for Curvenote issues. 

I have also noticed that neither of the assigned reviewers have commented here. I will follow up with them, and if necessary find more reviewers.

aditya14896:
Hi all, I'm Aditya Singh (@aditya14896). I'll be one of the reviewers for this paper — looking forward to reading it and will get a review posted soon.

aditya14896:
Review of PR #1227, "Feel the model: Sensory Transduction of Neural Activations as a Human-in-the-Loop Safety"

Reviewer: Aditya Singh (@aditya14896). Reviewed at commit 7aad798, against the Curvenote build of Jul 12.

Apologies for the delay, and thank you for your patience, @virajsharma2000.

Decision: approve for the 2026 Proceedings, conditional on Conditions 1 through 6 below.

@scipy-conference/2026-proceedings, one process note. All six conditions are addressable by editing text, with the single exception noted in Condition 5, so I believe they fit comfortably before the sign-off. I am happy to **re-review** quickly.

The topic, scope and Python tooling are a good fit for the Proceedings, and the AST proposal is a legitimate contribution. My conditions are almost entirely about aligning the stated claims with what the code actually does.

## Condition 1. Substantiate or remove the disagreement-analysis claim. This is the one blocking item.

The abstract states that cases of disagreement between the two classifiers "correlate with ground-truth misclassifications in the SAE-based classifier," and names this interpretive displacement, which the paper presents as its central contribution. The Experiment 2 text at line 68 then narrates that analysis in the first person: "I looked for cases where the two classifiers disagreed... In those disagreement cases, I checked which classifier matched ground truth. The raw classifier was correct more often than the SAE classifier."

The linked notebook for Experiment 2 contains no classifier, no disagreement set, and no ground-truth comparison. It loads GPT-2 and the SAE, computes the output-distribution entropy and a softmax entropy over SAE activations, and prints four means. Those four means are the only numbers the section reports.

So the paper's headline finding is described as having been performed but has no supporting analysis or numbers anywhere in the submission. Either is fine as a resolution:

- Run the analysis and report it: the number of disagreement cases, the proportion in which the raw-activation classifier was correct, and a confidence interval. Note that on a 72-row test set with a 6-item accuracy difference the disagreement set will be small, so please report its size plainly.
- Or remove the claim from the abstract and from line 68, and let the entropy comparison stand on its own with the caveats in Condition 3.

## Condition 2. Relabel Experiment 1. It does not measure text mediation.

The paper's thesis is that safety-relevant information is lost when activations are converted into human-readable text. Experiment 1 compares a logistic-regression probe on raw residual-stream activations against a probe on `sae.encode` outputs. No feature labels, annotations, auto-interp descriptions, or text of any kind appear anywhere in the pipeline.

Please change:

- The Table 1 row "SAE features (text)" to "SAE features".
- The row "Information lost by SAE" to "Accuracy gap", since an accuracy difference is not a measure of information.
- Line 44, "The 8% gap is information that existed in the model's internals but got lost when I converted it to text features", and line 60, "a direct measurement of behavioral signal destroyed by the SAE's text-mediated encoding", to describe loss under sparse ReLU encoding rather than under conversion to text.

Two related points worth a sentence each. `sae.encode` is a deterministic function of the raw activation, so raw performing at least as well as SAE features is expected by construction rather than a discovery. And a probe on 24,576 SAE features with 168 training rows sits in a very different regularization regime from a probe on 768 raw dimensions, so a matched-capacity control would strengthen the claim considerably. If a control is out of scope now, please say so in a limitations note.

## Condition 3. Report the sample size and uncertainty, and caveat the SAE entropy measure.

On Experiment 1: with 240 prompts and `test_size=0.3` at `random_state=42`, the test set is 72 rows. 67/72 is 0.931 and 61/72 is 0.847, so the reported 0.083 gap is exactly 6 test items on a single split. Wilson 95 percent intervals are 0.848 to 0.970 and 0.747 to 0.912, which overlap fully. Even in the most favourable case, where all six disagreements run one direction, an exact McNemar test gives p = 0.031. Please state the test-set size and either add repeated splits or cross-validation with intervals, or state explicitly that the gap is not statistically established on this sample.

On Experiment 2: `get_sae_entropy` applies softmax to raw ReLU SAE activations. Those are non-negative magnitudes rather than logits, and the resulting quantity is scale dependent. Holding a sparsity pattern fixed and rescaling the activations moves it across nearly the full range, from about 10.1 nats to near zero, with no change in which features are active, so it behaves largely as a function of how much the top activation dominates. It is also on a different support from the model entropy, 24,576 against 50,257, with maximum entropies of 10.11 and 10.82 nats, so 1.08 and 5.78 cannot be ordered against each other. Please either state these two limitations where the reversal is claimed at line 91, or replace the measure with something comparable, for instance decoding the SAE reconstruction through the unembedding and taking the entropy of that next-token distribution.

Also please report the entropy means to two or three significant figures with units and a dispersion measure, rather than the current 16 digits, and reconcile "8%" at line 44 with "8.3 percentage point gap above a 50% baseline" at line 60. The latter phrasing conflates the gap between the two classifiers with distance above chance.

## Condition 4. State the model, layer and SAE per experiment and figure.

There are currently three different models across the submission. The paper says GPT-2. Experiments 1 and 2 load `gpt2`, which is the 124M model with `d_model` 768, with the `gpt2-small-res-jb` SAE at `blocks.8.hook_resid_pre`. The backend notebook loads `gpt2-large` while its response payload hardcodes `"model": "gpt2"`, and the EDF client buckets a 1280-dimensional vector. The phase portrait figure is titled "GPT-2 Medium".

Please state model, layer, SAE release and token position for each experiment and each figure. The methods details that currently live only in the notebooks belong in the paper: n = 240, the 4 templates by 60 referents design, the 70/30 split, the seed, and the rule that labels are a median split on output entropy.

## Condition 5. Verify the batched last-token extraction. This is the one item that may change numbers.

Both experiments call `model.to_tokens` on batches of 4 and then take `acts[:, -1, :]` and `logits[:, -1, :]`. GPT-2's tokenizer defaults to right padding and TransformerLens honors it, setting the pad token to eos. For any batch containing prompts of unequal token length, index -1 is a pad position rather than the final content token. Prompt lengths do vary within batches, for instance "South Korea" against "Japan", and the fictional referents tokenize into several pieces each. Because referent length correlates with the real versus fictional split, which in turn correlates with the entropy label, the probe may be reading padding structure in part.

Please either re-run with `padding_side="left"`, per-prompt tokenization, or indexing from the attention mask, and report whatever the corrected numbers are, or state this as a known limitation. I recognise that if the numbers move, Conditions 1 through 3 may need revisiting, so please treat this one first.

## Condition 6. Fix the source links.

These are quick but they are explicit requirements in the review criteria.

Software citation. The criteria ask that all mentioned software be formally cited. TransformerLens, SAE-Lens, scikit-learn, MNE-Python, pyEDFlib, NumPy and Matplotlib are all used, MNE-Python is named in prose with no citation, and the `gpt2-small-res-jb` SAE needs attribution to its author. `myst.yml` still carries the template's `sklearn1`, `sklearn2` and `jupyter` keys in the doi-ignore list with no matching bib entries.

Source code. The six "View the Source Code" links point to files with an `.ipynb` extension that are not notebooks. They are plain Python text and will not open in Jupyter or nbviewer. Two of them also require pasting an ngrok URL that no longer resolves, so the phase portrait and EDF demonstrations cannot be reproduced by a reader. Please ship valid notebooks or `.py` files, pin library versions, and link an archived tagged release with a DOI rather than raw main-branch URLs. The sonification evidence is currently a YouTube link, which is not archival either.


Disclosure per the Gen AI policy: I used Amazon Kiro to help verify this review, specifically to re-derive the reported accuracies and confidence intervals, check the bibliography entries against arXiv and Crossref, inspect the linked notebooks, and confirm the TransformerLens padding default. The reading, the judgments and the recommendation are mine, and I take responsibility for them.


Thanks,
Aditya Singh

virajsharma2000:
@aditya14896  thanks for the detailed feedback. I am analyzing, hopefully will be able to address by early next week.

virajsharma2000:
Hi Aditya

I analyzed the two expriments I had submitted, a lot changed and I thought about how I had come up with the "feel the model" when I was checking the circuit tracer for figuring out model misalignment looking at attribution graph and huge amout of SAE feature texts. My goal and main idea for the pape ris not to prove that using SAEs as replacement model for interpretibilty is bad, it is to show how AST as a proposition can be more viable than the whole aattribution features approach.

So I removed  the old experiments and added new one - which is easier and simpler - it exposes a reconstruction issue. 

i did the following:

Condition 1 -Removed. The classifier-disagreement claim and Section 3.2 (Experiment 2) have been taken out of the paper.

Condition 2: Now I think its no longer applicable. This condition was with  the old Experiment 1, which has also been removed and replaced with a new experiment.

Condition 3 :  The SAE-entropy thingie no longer applies, since Table 2 is removed. The new experiment reports sample size (20/20) and a significance test (p < 0.0001).

Condition 4: Addressed. Backend, phase portrait, and EDF notebooks now consistently saying GPT-2 Large. The backend's /health endpoint and the phase portrait figure title were corrected to match. The paper text now states model/layer  in each relevant section.

Condition 5 : Fixed. model.tokenizer.padding_side = "left" is applied in the new experiment's code from the start.

Condition 6 - Little bit addressed. Added citations for TransformerLens, SAE-Lens, scikit-learn, MNE-Python, pyEDFlib, NumPy, Matplotlib, and the SAE author. Bib entries matching these citation keys still need to be added. The archived DOI release, notebook file format issue, and dead ngrok dependency have not been fixed; a note has been added to the paper disclosing that the phase portrait, sonification, and EDF demos are test scripts requiring a live backend session.

Let me know if this works - my 10 th exams are close..I will have to fix quickly :)

thanks

PR #1234, opened 2026-06-19
Paper: Vibes, meet rigor: Evaluating and improving AI performance on complex scientific code

cliffckerr:


cliffckerr:
Hi @chongshenng, this paper is now out of draft stage and is ready for review! I had a question about the AI disclosures though. While the [Gen AI Policy](https://github.com/starsimhub/scipy_proceedings/blob/2026/genai_policy.md) says *what* to write, it doesn't say *where* to write it. I added a section to the manuscript called "Disclosures", but let me know if this is the wrong place. Thanks!

anizmo:
I am available to review this PR.

shahchiragh:
Confirmation comment - I will be able to review this paper. 

Thanks,
Chirag

chongshenng:
Hello @cliffckerr! Great to see so much progress on your submission 😊. For the record, I am Chong Shen Ng (@chongshenng) and I will serve as the editor for your submission.

On "Disclosures", I reviewed the section that you mentioned and that looks good to me. Please reach out to me if any further assistance is needed.

cliffckerr:
Thank you @chongshenng, appreciate it!

cliffckerr:
Hi @anizmo and @shahchiragh -- I just pushed a new revision of the paper, so if you haven't started your reviews yet, please use the new version. Thank you!

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+
+```{figure} ./figures/fig4_judge_agreement.png
+:label: fig-judge-agreement
+:width: 100%
+
+**(a)** Agreement between the two independent rubric judges. Each point is one graded answer, plotting the Anthropic judge's score (x) against the OpenAI judge's score (y); the dashed line is perfect agreement. The judges agree closely, across both agent configurations (color) and model providers (marker). **(b)** Distribution of score differences between the two judges. An unbiased judge would have a distribution centered on 0; as it is, the Claude judge has a slight preference for Claude models (shifting the distribution to the right), while the Anthropic judge is the opposite (shifting to the left).
```
anizmo, line 117: Claude and Anthropic are the same judge. I assume the second one is supposed to be OpenAI. Please fix this, as it currently inverts the meaning.

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
+---
+title: "Vibes, meet rigor: Evaluating and improving AI performance on complex scientific code"
+abstract: |
+  Scientists apply rigorous methods to their research, but rarely to the AI tools they use to write code. We tested different large language models (LLMs) in combination with domain-specific tools (including MCP servers and custom-written skills) to find the optimal combination for writing complex domain-specific code. To evaluate LLMs' understanding of Starsim, a Python-based disease modeling framework, we wrote a quantitative proficiency exam. Using this exam, we tested different LLMs and varied their access to tools. Scores ranged widely, from ~50% for Claude Haiku 4.5 and GPT-5 mini without tools to 98.7% for Claude Opus 4.8 with tools. While LLM choice had the biggest impact on exam scores, tool access also significantly improved performance for both the cheapest (Haiku) and most capable (Opus) models, with more modest benefits for intermediate models. Thus, to improve LLM performance on domain-specific problems, we recommend developing a set of AI tools with the help of quantitative evaluation.
```
anizmo, line 4: The Results section gives 54.7% and 56.6%, and those numbers are for single-turn chat, not "agent-without-tools". "~50%" is an understatement, and "without tools" is ambiguous here.

anizmo, line 4: The abstract calls this "GPT-5 mini", but the Methods section says "GPT-5.4-mini". Please standardize.

anizmo, line 4: This 98.7% ceiling conflicts with Figure 2, which shows Opus-high-with-skills hitting 99.4%. If 98.7 is an average across effort levels, please clarify that.

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+
+```{figure} ./figures/fig1_cost_vs_score.png
+:label: fig-cost-score
+:width: 90%
+
+Performance as a function of model cost. Shape shows LLM type, and color shows model configuration. Individual runs are shown as small symbols; large symbols are the mean for that model-configuration combination. Not all models were run with all configurations.
+```
+
+Since skills tended to be evoked relatively infrequently without nudging (see below), we will focus on results that included nudging. Skills (with nudging) improved the performance of Haiku running in agent mode by 6.7% (from 68.9% to 75.6%), Sonnet by 2.0% (from 94.1% to 96.1%), and Opus by 1.7% (from 97.0% to 98.7%). While these gains may seem small, another way of looking at it is that skills reduced Opus' *mistakes* from 3.0% to 1.3% – a 57% improvement.
```
anizmo, line 83: Percentage points (pp) and percent are used interchangeably here. Please use "pp" when talking about absolute differences.

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+
+In order to quantify skill utilization, we logged every time any agent invoked a skill. The combined set of skill uses across all runs is what we defined the denominator to be, i.e. "100% utilization". (While technically this number is dependent on the number of runs, we found it had mostly saturated in the 5 runs included here.) Out of these possible skill uses, we counted how many actual uses any given run had.
+
+As shown in @fig-skill-utilization, without nudging, agents invoked skills only 7.8% of the possible times they could have. This accounts for the relatively small performance gain provided by simply including skills. When provided with the additional prompt nudge, skill usage increased to 28.4% – a 3.6-fold increase. We attribute to skill under-use to agent overconfidence rather than an optimal efficiency tradeoff, since the nudged agents not only had better performance, they also sometimes had lower cost (see @fig-cost-score). Surprisingly, skill usage was relatively uniform across the questions, even though the difficulty of the questions varied markedly, and skills were of much more help for some questions than others.
```
anizmo, line 126: Typo: There is a stray "to" in "We attribute to skill under-use to agent overconfidence"

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+
+Each combination (LLM, configuration, and effort level where applicable) was run five times to account for stochasticity in both answering and marking. Answers were graded by LLM-based marking agents rather than by hand, with marks point-checked and cross-validated against the schema by the human experts who wrote the exam. Hand-marking was infeasible: a completed exam was typically about 2,500 lines, roughly half of it executable Python, amounting to approximately 400,000 lines of output across all experiments.
+
+## Results
+
+### Performance scales with cost and skills
+
+Unsurprisingly, costlier models perform better. The cheapest models with single-turn ("chat") configurations, GPT-mini-chat (USD0.11) and Haiku-chat (USD0.21), also performed by far the worst (56.6% and 54.7%, respectively). The best-performing model, Opus-nudged (98.7%), was also among the most expensive (USD13.38). These results are shown in @fig-cost-score. Opus outperformed Sonnet on both performance and cost.
```
anizmo, line 74: This formatting looks a bit odd. I suggest formatting it as "$0.11" or "USD 0.11".

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+
+Beyond trying different models, few developers rigorously test whether other changes to their workflow help. This is because writing and executing these evaluations is typically time-consuming, the results of the evaluations are difficult to interpret and quantify, and AI workflows and tooling are evolving rapidly. Even among engineering leaders who track AI at all, a minority do so systematically [@leaddev2025ai]. Here we describe our process for quantitatively testing our assumptions about how to build a good AI assistant.
```
anizmo, line 11: The LeadDev survey cited here polls engineering leaders, which isn't quite the same demographic as scientists writing research code. This feels like a minor overreach.

anizmo (commented):
## MAJOR FEEDBACK

- Statistical Claims: Consider softening the use of "significantly," as effect sizes are relatively small (1.7 to 2.0 pp, n=5). Adding paired CIs and reporting mean +/- SD or IQR would really strengthen this section.

- Baseline Comparison: Adding a "docs-only, no network" baseline would help clearly isolate the unique value of your curated skills compared to standard retrieval.

- Tool Utilization: It would be helpful to highlight in the abstract and conclusion that the best scores required active prompting, given that unprompted tool use was 7.8 percent.

- Evaluating Judges: Since R-squared indicates reliability rather than validity, it would be great to see human vs. LLM discrepancies. Using Bland-Altman or ICC instead of correlation, and exploring the 40-point outlier, would provide a clearer picture.

- Construct Validity: Because the exam and skill authors overlap, briefly acknowledging the potential risk of "teaching to the test" would improve overall transparency.

- Reproducibility: Including a Data Availability Statement and pinning exact model API strings, snapshot dates, and the Starsim version would be incredibly helpful for future researchers.

- Cheating Detection: Quantifying the honeypot triggers in a table, rather than using anecdotes, would turn an interesting observation into solid evidence and help ground the AI-safety framing.

## MODERATE FEEDBACK

- Skill Metrics: Creating a human-labeled relevance key for the denominator might provide a more accurate measure of true recall.

- Cost Analysis (Fig 1): A summary table detailing model, configuration, score, raw tokens, and cost would add great clarity. It may also be worth noting that Haiku 4.5 is an older generation.

- Effort Non-Monotonicity: The U-shaped improvements might be best framed as expected within-run noise due to the small sample size (n=5).

- Opus 4.6 Retrospective: Detailing the exact evaluation methodology for the 70 percent to 91 percent jump would fully support its inclusion in the Discussion.

- Relative Error: Adding a ceiling-effect caveat to the 3.0 percent to 1.3 percent error drop would be beneficial, as would standardizing the improvement metric (57 percent vs. 60 percent) throughout the text.

## FINAL VERDICT

- Accept with Revisions

It is a really good paper, well written and it could benefit by implementing the feedback provided. Feel free to mention your input on the feedback.

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Marking is robust to judge choice
+
+Because the answers are marked by LLMs, a natural concern is whether the results are an artifact of a particular judge. @fig-judge-agreement addresses this by plotting the Anthropic judge's score against the OpenAI judge's score for every marked answer. The two judges agree closely (Pearson $R^2$ = 0.96) with only a small mean difference (the Anthropic judge is higher by 2.2% on average). The high cross-provider correlation indicates that the scores reflect properties of the answers rather than the idiosyncrasies of a single marker, and gives us confidence in the comparisons above.
+
+That said, the panel does reveal a small but consistent own-provider bias: each judge scores answers written by its own provider's models slightly more generously. On average, there was a preference of 2.1% for a judge towards its own work versus the other provider's. While much smaller than the variance in scores (in one extreme case, there was a 40-point gap between the Anthropic and OpenAI judges' scores on the same exam), it is still notable since even with a rigorous and explicit marking schema, it shows that a model's "personal" bias is still present.
```
shahchiragh, line 111: Judge bias (2.1%) is disclosed, but not clear how much it's actually correcting for in the reported numbers. Since most of the headline comparisons are between Claude models scored partly by a Claude judge, this bias could be inflating the Haiku/Sonnet/Opus numbers a bit relative to what an independent human-anchored score would show.

shahchiragh, line 111: The 40 point outlier gap mentioned in passing is also seems concerning. That's a huge single-case disagreement. Was it an outlier or symptomatic of something in the marking pipeline? A distribution of just how many answers show large judge disagreement would help more than the aggregate R^2 number

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+Since skills tended to be evoked relatively infrequently without nudging (see below), we will focus on results that included nudging. Skills (with nudging) improved the performance of Haiku running in agent mode by 6.7% (from 68.9% to 75.6%), Sonnet by 2.0% (from 94.1% to 96.1%), and Opus by 1.7% (from 97.0% to 98.7%). While these gains may seem small, another way of looking at it is that skills reduced Opus' *mistakes* from 3.0% to 1.3% – a 57% improvement.
+
+### Model effort quickly saturates
+
+Claude models include a configurable "effort" parameter; higher effort consumes more tokens but is supposed to produce better results. As shown in @fig-effort-score, performance saturated above medium effort (extra-high and max effort levels were also run for a subset of trials, which also showed no further improvement; results not shown). Interestingly, skills (with nudging) improved performance in both the lowest-performance case (Sonnet-low, 92.0→95.1%) and the highest-performance case (Opus-high, 98.3→99.4%), but less in moderate-performance cases (Sonnet-medium, 96.7→96.1%; Opus-low 97.4→97.4%).
```
shahchiragh, line 87: Figure 3's caption explicitly flags that the error bars are min/max, not confidence intervals, which is honest, but it also means several of the effort-level comparisons (e.g., the "less improvement in moderate cases" claim) could just be noise given the ranges shown?

papers/cliff_kerr/main.md
shahchiragh, line 126: I could be understanding this wrong but it appears the "skill utilization" denominator is defined relative to itself. 

Let say 100% is the union of all observed skill invocations across the very runs being measured, not an independent estimate of how often a skill actually would have been useful. Which would make the 7.8% -> 28.4% numbers a relative comparison between two conditions, not an absolute utilization rate. 

It might be worth being explicit that this could shift a fair bit depending on run count.

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+
+In humanoid mode, we ran the three Anthropic models (Haiku 4.5, Sonnet 4.6, and Opus 4.8). For Sonnet and Opus we also varied the "effort" level (low, medium, and high; exploratory runs at extra-high and max levels performed similarly to high and were not pursued further, to save tokens). These models were run with three configurations, slightly different from pipeline mode. "Chat only" was unavailable, since the Claude Agent SDK is multi-turn by default. "Agent only" was enforced through prompt instructions rather than Docker firewalls (cheating attempts, such as `curl` from Bash and `urllib` from Python, were surprisingly frequent, but were intercepted and blocked). "Agent + skills" was similar to pipeline mode. Humanoid mode also used a fourth configuration, "skills + nudged", in which the system prompt included the following nudge, added because exploratory runs showed plugin usage was low:
```
shahchiragh, line 64: The skills+nudged condition brings two effects: increased skill invocation and increased directive confidence framing. 

It might strengthen the causal story to have a condition with the nudge language but skills disabled, just to isolate how much of the gain is "use this tool" versus "trust this source over your own judgment."

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+```{figure} ./figures/fig2_effort_vs_score.png
+:label: fig-effort-score
+:width: 90%
+
+Performance as a function of effort for Sonnet and Opus models. Note that vertical bars show minimum and maximum scores, not confidence intervals.
+```
+
+### Marks are lost for multiple reasons
+
+No single factor accounted for a majority of marks lost. As shown in @fig-lost-marks, marks were lost approximately equally due to the answer being (a) incorrect, (b) omitted, or (c) for a different reason. Interestingly, the primary improvement in marks was due to a reduction in the number of omitted answers, rather than incorrect answers. This suggests that skills can be helpful in prompting the LLM to bring the solution to completion when it may have otherwise become stuck.
```
shahchiragh, line 98: Minor: the "marks lost" breakdown in the figure is great but it would be more useful broken down by question difficulty or topic.
Since the discussion notes skill usage was _surprisingly uniform_ across questions of varying difficulty - pairing that with where marks were actually being lost would make the "agents don't know what they don't know" argument more concrete.

papers/cliff_kerr/main.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Starsim-AI
+
+We created an open-source set of agent tools to improve LLMs' performance when working with Starsim, called [Starsim-AI](https://github.com/starsimhub/starsim_ai). Specifically, we added Model Context Protocol (MCP) servers for both Starsim and [Sciris](https://docs.sciris.org), which is a scientific Python library used widely in the codebase [@sciris]. We also created a set of "skills" for Starsim, which consist of problem-solving and feature-oriented Markdown files. Each skill carries a short natural-language description that the agent uses to decide when the skill is relevant, plus a body that is loaded into context only when triggered.
```
shahchiragh, line 31: Is there any external validation done? (e.g., a non-affiliated Starsim user grading a subset by hand)
My thoughts - the same team wrote Starsim, wrote the skills, wrote the exam, and is judging correctness. 
Theres's reasonable justification (most qualified to judge correctness), but it does mean there's no independent replication

shahchiragh:
Very useful paper - methodological, transparent, and sound study. Genuinely interesting. 

Some minor comments as part of the feedback:
- External validation point - isn't disqualifying, but consider stating more explicitly rather than justify away in one sentence. 
-  Judge bias 2.1% - disclosed but not corrected in headline numbers. Most comparisons are within Claude family scored partly by a claude judge. 
- Skill utilization % - fine as relative comparison but should be labeled as such rather than implying absolute
- nudge+skills conditions - not fully clear how much of the gain is retrieval benefit vs. just more directive/confident prompting

Final acceptance check:  Some quantified human-anchor check on the LLM marking to actually underpin whether the headline numbers are trustworthy after all. 

**Overall:** 

Accept with revisions

papers/cliff_kerr/main.md
cliffckerr: Fixed

papers/cliff_kerr/main.md
cliffckerr: Fixed

cliffckerr: Fixed

cliffckerr: Fixed. I left ~50%, since I think trying to be more precise summarizing across multiple trials for two models would add too much complexity.

papers/cliff_kerr/main.md
cliffckerr: Fixed

papers/cliff_kerr/main.md
cliffckerr: Fixed (to US$0.11).

papers/cliff_kerr/main.md
cliffckerr: I agree they are different demographics, but one would think that "engineering leaders" would be more systematic in their exploration of AI than scientists would -- hence the "even" (as in, scientists would likely be even less systematic).

papers/cliff_kerr/main.md
cliffckerr: Fixed

papers/cliff_kerr/main.md
cliffckerr: We agree that this is a limitation; added to the Discussion: "Third, the same authors who wrote Starsim also wrote the exam, the marking scheme, and evaluated the LLMs' fidelity to the marking scheme; independent replication would increase the robustness of these results."

papers/cliff_kerr/main.md
cliffckerr: We appreciate the suggestion, but we are not entirely sure how a prompt of "Use the skills provided" would work in a scenario where no skills are provided.

papers/cliff_kerr/main.md
cliffckerr: We did prepare figures like that, e.g.:
image

While the right panel initially looks interesting, the higher bars are actually just because these questions are worth more marks. Upon human review, we also felt that there was too much ambiguity to support these five categories of lost marks, hence why we simplified for the manuscript. We think the figure in the manuscript is the best compromise of information content and interpretability.

papers/cliff_kerr/main.md
cliffckerr: We agree this is a limitation, but in our view a fairly modest one since the judges' agreement was far greater than their differences. Added to the Results: "Human review of marking discrepancies revealed that most cases were the result of genuine ambiguity; in the case of the 40-point gap, for example, the OpenAI model did _not_ use Starsim to solve the problem, which led the Anthropic judge to award the solution 0 points, while the OpenAI judge awarded full points. The difference hinged on the judge's interpretation of the marking instruction "Use Starsim to answer questions unless it would add unnecessary complexity". In this case, the human reviewer agreed with the Anthropic judge's assessment (i.e., Starsim would have made the solution simpler, not more complex), but in other cases the OpenAI judge was deemed more accurate. However, note that in the vast majority of cases, both judges agreed."

cliffckerr: Addressed in the previous comment (added to Results).

papers/cliff_kerr/main.md
cliffckerr: While it's true that the sample sizes are relatively small, fairly consistent trends were found across models (including the high and extra-high effort runs which are not shown in the figure since we felt the token cost was prohibitive to complete the runs across all trials). We changed the figure to show median and IQR as well as min/max.

papers/cliff_kerr/main.md
cliffckerr: That's correct, it is defined relative to itself, which is indeed a limitation. We did additional analysis to show that it indeed saturated:
image

Added to Results: "In order to quantify skill utilization, we logged every time any agent invoked a skill. The combined set of skill uses across all runs is what we defined the denominator to be, i.e. "100% utilization". (While technically this number is dependent on the number of runs, we found it had effectively saturated after 3 runs, and 5 were included here.) Out of these possible skill uses, we counted how many actual uses any given run had. While this is a relative rather than absolute measure of skill use, it is not possible to independently determine when an agent "should have" invoked a skill, except by seeing when they actually invoked them.

Note: as part of this analysis, we discovered that the denominator had been calculated incorrectly (it included several runs that were not included in the denominator), leading to an underestimate of skill use. This has been fixed in the figure and manuscript text.

cliffckerr:
Thanks for the review, @anizmo! Responses to your comments:

- **Major feedback**
  1. **Statistical claims**: Done; "significant" has been removed in cases where it may have had an unintended statistical connotation.
  2. **Baseline Comparison**: We appreciate the suggestion, and we agree it would be an interesting condition. However, to avoid additional complexity, and since running in this configuration would not be a realistic scenario, we have decided to omit it.
  3. **Tool Utilization**: Added to the abstract: "For all models, tool utilization was low without specific prompt instructions."
  4. **Evaluating Judges**: Added to the Results: "Human review of marking discrepancies revealed that most cases were the result of genuine ambiguity; in the case of the 40-point gap, for example, the OpenAI model did _not_ use Starsim to solve the problem, which led the Anthropic judge to award the solution 0 points, while the OpenAI judge awarded full points. The difference hinged on the judge's interpretation of the marking instruction "Use Starsim to answer questions unless it would add unnecessary complexity". In this case, the human reviewer agreed with the Anthropic judge's assessment (i.e., Starsim would have made the solution simpler, not more complex), but in other cases the OpenAI judge was deemed more accurate. However, note that in the vast majority of cases, both judges agreed." We did not use a Bland-Altman plot since panel B is effectively a kernel density version of a Bland-Altman plot.
  5. **Construct Validity**: We agree that this is a limitation; added to the Discussion: "Third, the same authors who wrote Starsim also wrote the exam, the marking scheme, and evaluated the LLMs' fidelity to the marking scheme; independent replication would increase the robustness of these results."
  6. **Reproducibility**: Added a new section "Data availability": "All code and materials are publicly available except the Starsim exam itself, which is intentionally kept private to avoid questions and solutions becoming part of models' training data, which would reduce its usefulness; however, the exam is available on request. Materials that are available publicly include [Starsim](https://github.com/starsimhub/starsim) (version 3.3.0), [Starsim-AI](https://github.com/starsimhub/starsim_ai) (version 1.4), the [code to run the evaluations](https://github.com/starsimhub/scipy2026_code) (minus the exam and solutions), and the [code to make the figures](https://github.com/starsimhub/scipy2026_paper) for this paper. We used `uv` to ensure exact reproducibility between environments."
  7. **Cheating Detection**: There were attempts that were obviously cheating and more borderline cases; we were not able to quantitatively classify these or trace their provenance. We added the following text to the Discussion: "While concerning, these attempts at cheating comprised a relatively small proportion of sessions (<10%). We manually reviewed the logs from these sessions, but were not able to determine conclusively why some agents chose the path of darkness."
- **Moderate feedback**
  1. **Skill Metrics**: We appreciate the suggestion, but we feel this would be too time consuming, and would be imperfect since there is some ambiguity in what questions would benefit from invoking a skill (since human annotations do not know what is already in the models' training data).
  2. **Cost Analysis**: Added as the new Table 1 in the manuscript.
  3. **Effort Non-Monotonicity**: I agree that it's due to a small sample size effect, but because of the n=5 models being evaluated, rather than the n=5 runs done of each. Added to the Discussion: "Our finding that skills helped most with the cheapest models (Haiku, GPT-5.4-mini) and the most capable models (Opus), but less so with intermediate models (Sonnet, GPT-5.5), was surprising. Since this capability categorization is constantly changing (i.e., today's most capable models will be tomorrow's intermediate models), this seems more likely to be an artifact of the five models tested here rather than a general principle of how LLMs use skills."
  4. **Opus 4.6 Retrospective**: We agree this would be interesting, but unfortunately it is not possible. Added to the Discussion: "Unfortunately, since these are not open-weight models, and since the agent harness code is not public, we are unable to determine exactly what caused such large performance improvements."
  5. **Relative Error**: We believe there is _not_ a ceiling effect, which is what makes this result interesting: skills had the same ~60% benefit for low-performing as well as high-performing models. We have however taken the reviewer's suggestion to use "pp" rather than "%" to talk about absolute differences in percentage points, however.

cliffckerr:
Thanks for your comments @shahchiragh, much appreciated. I believe all your comments have been addressed as replies to your comments above -- thank you!

cliffckerr:
Hi @chongshenng -- my response to the reviewers is now complete.

chongshenng:
Thanks @cliffckerr for carefully addressing the reviewer comments! 

Thank you also to @anizmo and @shahchiragh for reviewing this submission -- is the revision acceptable in the current state?

anizmo:
> Thanks @cliffckerr for carefully addressing the reviewer comments!
> 
> Thank you also to @anizmo and @shahchiragh for reviewing this submission -- is the revision acceptable in the current state?

Hello @chongshenng , I went through the changes and they look good to me. Thank you @cliffckerr for working on the feedback and providing your input as well.

anizmo (approved):
The changes look good to me.

shahchiragh:
Thank you @cliffckerr for carefully addressing the comments/suggestions! I understand they were quite a few to address but overall the paper reads great.  @chongshenng the revisions look good to me 🚀

shahchiragh (approved):
The revisions look good to me 🚀

chongshenng (approved):

PR #1253, opened 2026-06-20
Paper: Right Predictions, Wrong Reasons: Explanation Drift Monitoring in Production

ab93:
If you are creating this PR in order to submit a draft of your paper, please name your PR with `Paper: <title>`. An editor will then add a `draft` label; this will trigger GitHub Actions to run automated checks on your paper and build a preview. You may then work to resolve failed checks and ensure the preview build looks correct. If you have any questions, please tag the proceedings team in a comment on your PR with `@scipy-conference/2026-proceedings`.


sanjaybk7:
I'll be reviewing this paper for SciPy Proceedings 2026. Confirming as Reviewer.

chongshenng:
Hello @ab93, I am Chong Shen Ng (@chongshenng). I will serve as the editor for your submission. Please reach out to me if any assistance needed 😊.

EngineerKhan:
I'll be reviewing this paper. Confirming as Reviewer.

EngineerKhan:
Overall, I found this to be a clear, practically motivated paper addressing an interesting gap in production ML monitoring. Treating explanation distributions as a monitorable production signal is a useful idea, and the proposed combination of per-feature SHAP PSI, adversarial validation, importance/rank changes, and sign changes provides a practical framework for doing so.

My main concern is about distinguishing three related but different claims: 

1. Explanation monitoring can detect changes in attribution behavior
2. It can detect changes that conventional input monitoring would miss, and 
3. Those changes indicate that a model is making predictions for the “wrong reasons.” 

I think the current case study provides good evidence for the first claim, but the second would benefit from raw-input drift baselines. In particular, because the experiment deliberately changes the population from age <= 45 to age > 45, conventional input monitoring would presumably detect a substantial shift as well. Comparing raw-feature PSI with SHAP PSI, and raw-feature adversarial AUC with SHAP adversarial AUC, would make the incremental value of explanation-space monitoring much clearer.

Overall, I think the central idea and implementation are useful, and my comments are primarily aimed at sharpening the claims and strengthening the empirical demonstration.

_(I used LLM to refine/summarize my thoughts in this review)_

sanjaybk7:
Thanks for the submission. I went through the paper as well as the shap-monitor package. Overall I find this a useful and practically relevant contribution for the SciPy community. The problem it addresses is a real one — a model holding its accuracy steady while its reasoning quietly changes underneath, which is especially painful in delayed-label settings — and the paper is written clearly, with an honest limitations section. The tool is open-source and built on a standard scientific-Python stack, so it fits the proceedings well. I am inclined towards acceptance, but I have a few points which I would request the authors to address in revision. I have grouped them into major and minor.

Major comments

The case study does not fully isolate the blind spot that is being claimed. The central claim is that explanation drift catches what input-drift and performance monitoring miss. The case study demonstrates the gap against performance monitoring quite nicely, since the F1 stays essentially flat (0.71 → 0.72). But the shift used here is a direct covariate shift on age, so ordinary input-drift monitoring on the age feature would also fire very strongly in this setup. As such, this particular study does not really show the advantage over input-drift monitoring. The more convincing cases are actually the other two patterns you describe — the version regression on identical inputs, and the pipeline bug that pins a feature to a plausible value — but these are only described, not shown. It would strengthen the paper a lot if you could add even one small case study for one of those, or at the minimum report the input-drift signals side by side, so the reader can clearly see what explanation drift is adding beyond "the data has moved."
Related work on explanation/attribution drift is a bit thin. At present the input-drift tools (Evidently, Alibi Detect) are cited, but there is existing work specifically on explanation shift / attribution-based drift detection that is not discussed. Because of this, contribution (1) — framing explanation drift as a monitoring signal — reads as slightly overstated. Kindly position the work against the prior explanation-shift literature and state clearly what is genuinely new here.
Interpretation of the adversarial-validation AUC. Since the SHAP values are themselves a function of X, an AUC of 0.97 will be driven partly by the input shift itself and not purely by a change in reasoning. It would be good if the authors could comment on how much of the separability is due to the reasoning change versus the underlying covariate shift, so that the "0.50 → 0.97" story is not over-attributed to reasoning alone.
Please add some measure of uncertainty. The headline numbers (adversarial AUC, PSI, the percentage changes) are all single-run point estimates. The limitations section itself notes that attributions have variance, so repeated splits or error bars on the case-study numbers would make the results considerably more convincing.

Minor comments

The age PSI of around 11 is a very large value. Since the ln(c/r) term can blow up when a reference bin has near-zero mass, kindly clarify the binning strategy and confirm that PSI is stable and well-behaved for these continuous SHAP distributions.
Reproducibility is in good shape — I was able to confirm the case study runs from examples/lightgbm_example.ipynb (it pulls the UCI Adult dataset directly, dependencies are pinned via poetry.lock, and the package API matches the paper). Two small requests: please reference this notebook explicitly from the paper so readers can find the reproduction, and confirm that it regenerates all three figures (the PSI and SHAP-distribution plots, not just the drift score).
A couple of sentences need a small fix. The compare_versions sentence in the "Analyzing drift" section reads as incomplete, and the "…and transfers directly" phrase in the PSI paragraph is a bit confusing.
The myst.yml suppresses the DOI check for a few keys (sklearn, lightgbm, etc.) that do have standard citations. Adding the DOIs would be good for the bibliography.

To be clear, all of the above is meant constructively — the core idea is a nice one and I would genuinely like to see it in the proceedings. Addressing the case-study point (major comment 1) in particular would make the contribution much stronger. Happy to look at the revised version. Thanks again for the good work.

(I used LLM to refine/summarize my thoughts in this review)

ab93:
Thank you both for the comments. Working on addressing them. Appreciate it!

ab93:
Hi all, here is the summary of the changes after the review.

### Case study didn't isolate the blind spot

The raw-input baselines were added. And what I saw was input-drift monitoring detects the change as well for the study. The paper now says so plainly. What explanation drift adds there is per-feature, e.g. on `race` and
3 others the input distribution stays quiet while the attributions cross the warn band.

A second study was added, where model versions are scored on byte-identical inputs. Input
drift is zero by construction and accuracy is preserved, so explanation drift is the only
label-free signal that fires. It's reported as a family of constructions, including a null
control that correctly flags nothing and also just a hyperparameter change.

### Related work / contribution

Added a Related work section covering explanation shift and SageMaker Model
Monitor. Contribution (1) is rewritten to claim the delineation of *when* the signal helps.

### AUC interpretation

New limitation noting that attributions are a function of the inputs, so much of Study A's
separability is inherited from the covariate shift rather than added by it. The independent
claim rests on Study B.

### Uncertainty

Headline figures now carries an interval over repeated subsamples. Documented some nitpicks, e.g. with-replacement bootstrapping inflates a genuine no-drift control, because duplicated rows leak across cross-validation folds.

### Other Minor Fixes

1. Binning documented in a new subsection. PSI is well behaved under the null, but the large
   `age` value turns out to be dominated by the epsilon floor rather than being a meaningful
   magnitude, so it's now reported as saturated.
2. `examples/paper_notebooks` cited; it regenerates every number and figure.
3. Some DOI fixes

chongshenng:
Thanks @ab93 for addressing the reviewer comments! And thank you @EngineerKhan and @sanjaybk7 for providing reviews -- is the revision acceptable in the current form?

EngineerKhan:
It's good to me. Thanks for updating it.

sanjaybk7:
Thanks @ab93 for the thorough revision — it addresses my comments well. The added raw-input baselines and Study B (byte-identical inputs, with a null control that correctly flags nothing) isolate what explanation drift contributes beyond input monitoring, which was my main concern. I also appreciate the candid AUC caveat that much of Study A's separability is inherited from the covariate shift, the uncertainty intervals on the headline figures, and reporting the large PSI as a saturation flag rather than a magnitude. The reproduction notebooks in examples/paper_notebooks/ are a valuable addition.

I approve this submission for the Proceedings in its current form.

chongshenng (approved):

PR #1256, opened 2026-06-20
Paper: ArcadeDB in Python: An In-Process Multi-Model Database

tae898:
Submission for SciPy 2026 Proceedings.

**ArcadeDB in Python: An In-Process Multi-Model Database** — an embedded, multi-model database for scientific Python that unifies documents, graphs, and vectors in one in-process runtime, with Python bindings over a Java engine (via JPype, with a jlink-trimmed bundled JRE). A single `pip install` provides documents (SQL), a property graph (OpenCypher), and HNSW vector search in one process, with no server to deploy.

The paper centers on a hybrid retrieval workflow that composes vector search → SQL filtering → graph traversal over one dataset in one process and one transaction, and includes a reproducible, resource-capped benchmark suite comparing ArcadeDB against SQLite, DuckDB, LadybugDB, and Chroma on the Cross Validated (stats.stackexchange.com) corpus across tabular, graph, and vector lanes (OLTP/OLAP).

The binding (Apache-2.0), the benchmark suite, the datasets, and the run manifest are openly available.


jameslamb:
👋🏻 hi @tae898 , I've been assigned as one of the reviewers for this paper. Looking forward to working with you.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+
+A growing share of scientific and applied machine-learning work is data-shaped in three
+different ways at once. A retrieval-augmented question-answering system stores documents and
+their metadata (records), models who-answered-what or what-cites-what (relationships), and
+retrieves passages by semantic similarity (vectors). A bioinformatics or recommender pipeline
+has the same character. Python is the lingua franca for all of it, but no single
+Python-embeddable data store covers all three access patterns. The usual response is to
+assemble a stack: SQLite or DuckDB for records, NetworkX or a graph database for
+relationships, and a vector index such as FAISS, hnswlib, or Chroma for embeddings
+[@duckdb2019; @sqlite; @networkx2008; @faiss2019; @hnswlib; @chroma].
```
jameslamb, line 33:
```suggestion
retrieves passages by semantic similarity (vectors). Python is popular for these workloads, but no single
Python-embeddable data store covers all three access patterns. The usual response is to
assemble a stack: SQLite or DuckDB for records, NetworkX or a graph database for
relationships, and a vector index such as FAISS, hnswlib, or Chroma for embeddings
[@duckdb2019; @sqlite; @networkx2008; @faiss2019; @hnswlib; @chroma].
```

Proposing some simplifications:

* remove the sentence about bioinformatics... it does not add any information
* replace the phrase "lingua franca" with simpler, more direct language

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+
+Assembling specialists is a reasonable default, and for many projects the right one. But its
+costs fall hardest on the local, exploratory, reproducibility-sensitive work that
+characterizes scientific Python. Data must be copied and kept in sync across systems with
+different consistency models. Joining a graph traversal to a vector neighborhood means moving
+identifiers between processes by hand. Reproducing an experiment means standing up and pinning
```
jameslamb, line 39: The claims in this paragraph read as pretty exaggerated, to the point that I think it distracts from the content of the paper.

For example, the idea that using DuckDB for SQL and `networkx` for graph data is equivalent to doing data processing "by hand" is not really true or fair, in my opinion.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+several services instead of installing one package. None of that is fundamental to the
```
jameslamb, line 40:
```suggestion
several services instead of installing one package.
```

I'd remove this sentence, it doesn't add any information to the paper. The introduction already used the words "science" / "scientific" to try to orient this towards SciPy, the paper doesn't need continued generic "and you can use this for science" types of phrases.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+its SQL dialect, the graph with OpenCypher, and vectors through an HNSW index. Workload type
+is a separate axis: transactional (OLTP) and analytical (OLAP) queries run over the same data.
+ArcadeDB is ACID, and recent versions add Graph Analytical Views (GAV) that accelerate graph
+analytics without giving up transactional guarantees.
+
+The engine, however, is Java. Our contribution is what makes it usable in-process from scientific Python.
+**`arcadedb-embedded`** is a binding built on JPype [@jpype; @arcadedbpython]. It runs the
+engine inside the Python process over a *bundled* per-platform Java runtime, exposes a
+NumPy/pandas-friendly API, and ships as ordinary wheels (`pip install arcadedb-embedded`, no
+Java installation, no server). The focus throughout is enablement. We do not claim the engine is fastest at any one task, and
```
jameslamb, line 56:
```suggestion
Java installation, no server). We do not claim the engine is fastest at any one task, and
```

This seems like as good a place as any to leave this comment... filler statements like this one I'm recommending removing are strong evidence that a significant portion of the text was generated by an LLM.

I see you added an AI use disclosure that says you used AI tools to help with this paper, and that is allowed under the proceedings' AI use policy (https://github.com/scipy-conference/scipy_proceedings/blob/2026/genai_policy.md). But from this point forward I think the paper would really benefit from your making remaining edits manually and from you doing a copy-editing pass for reducing the "AI voice" in the paper. I'm only a few paragraphs in and I've already found it pretty distracting (personal context: I am exhausted from reading AI-generated text at work and in open source every day), and I suspect other readers might as well.

I'll avoid leaving more comments about this specific thing in this first round of review. I hope you'll consider this editing pass.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+we show where it is not. We claim that documents, graph and vectors *in one Python process*
+remove the boundaries that fragment these workflows.
+
+Concretely, this paper contributes:
+
+1. **An in-process multi-model database for Python.** We describe the binding design: an in-process JVM via JPype, transaction and lifecycle
```
jameslamb, line 62: `JPype` should be a citation, please: https://www.osti.gov/doecode/biblio/46107

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+management, and NumPy/pandas interop. Platform-agnostic Java bytecode plus a per-platform
+bundled JRE yields wheels for four platforms × five Python versions, installable with no Java
+present.
+2. **A hybrid in-process workflow as the central demonstration.** A single retrieval pipeline
+   composes vector search → SQL filtering → graph traversal over one dataset in one process
+   and one transaction. We argue, with a capability matrix, that no single Python-embeddable
+   alternative expresses this composition, and that the binding is what makes it possible from
+   Python at all.
+3. **An honest characterization of the embedded-from-Python experience.** A matched, resource-capped, reproducible benchmark suite compares the engine (*as reached
```
jameslamb, line 71:
```suggestion
3. **An honest characterization of the embedded-from-Python experience.** A reproducible benchmark suite compares the engine (*as reached
```

I think it's sufficient just to say "reproducible" here, I'm not sure what "matched" means and "resource-capped" is an implementation detail of the benchmark that doesn't belong in this high-level summary.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+analytical SQL [@duckdb2019]. For relationships, NetworkX [@networkx2008] provides in-memory
+graph analysis, while LadybugDB [@ladybugdb], the maintained continuation of the Kùzu project
+[@kuzu2023], is an embedded, analytics-oriented *graph database* with a Cypher interface and
+vector indexing.
+For vectors, hnswlib provides the reference HNSW implementation
+[@hnsw2018; @hnswlib], FAISS provides exact and approximate similarity search [@faiss2019],
+and Chroma packages HNSW behind a convenient Python API [@chroma]. Each is excellent within
+its model. What none provides is *all three models in one embedded engine*. SQLite has no
+graph or vector type, DuckDB has no transactional graph or Cypher, LadybugDB is graph-first
+(graph and vectors, but no document/SQL OLTP store), and Chroma is vector-only.
```
jameslamb, line 93: The parallel structure in this sentence is not consistent

```
SQLite has no {thing}
DuckDB has no {thing}
LadybugDB is graph-first (graph and vectors, but no {thing})
Chroma is vector-only
```

That change of format raises the effort for readers to understand the differences (I know I had to slow down and think hard about what this was saying).

I think this would be more easily understood as a feature matrix table, would you consider that instead?

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
+---
+# Keep this title identical to the one in `myst.yml`
+title: "ArcadeDB in Python: An In-Process Multi-Model Database"
+abstract: |
+  Scientific Python workflows increasingly mix three kinds of data access: transactional
+  reads and writes over records, traversals over relationships, and similarity search over
+  vector embeddings. In practice these are split across separate systems (a relational or
```
jameslamb, line 7:
```suggestion
  Scientific Python workflows sometimes mix three kinds of data access: transactional
  reads and writes over records, traversals over relationships, and similarity search over
  vector embeddings. In practice these are split across separate systems (a relational or
```

The claim that workloads that mix these 3 types of queries are "increasingly" used is difficult to substantiate. I don't think the paper's impact would be decreased at all by just saying that such workloads "often" or "sometimes" mix those types of queries.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+[@hnsw2018; @hnswlib], FAISS provides exact and approximate similarity search [@faiss2019],
+and Chroma packages HNSW behind a convenient Python API [@chroma]. Each is excellent within
+its model. What none provides is *all three models in one embedded engine*. SQLite has no
+graph or vector type, DuckDB has no transactional graph or Cypher, LadybugDB is graph-first
+(graph and vectors, but no document/SQL OLTP store), and Chroma is vector-only.
+
+**The composable-stack counterpoint.** A natural objection is that you do not need one engine.
+You compose specialists. @datta2025composable articulates this "Python is all you need"
+position: a composable, Python-native data stack assembled from best-of-breed components. We
+take that argument seriously, and for warehouse-scale and team settings it is compelling. Our
```
jameslamb, line 98: > _for warehouse-scale and team settings it is compelling_

I think there are 2 overlapping topics being discussed throughout the paper (at least up to this point):

* in-process databases vs. services you communicate with over a network
* combining Python libraries specialized to each of 3 different query patterns vs. this `arcadedb-embedded` 

I recommend devoting a paragraph to the decision to use in-process datastores vs. server-based datastores, and for the rest of the paper only compare `arcadedb-embeded` to those other in-process options. That'd simplify things and make the comparisons a bit sharper, in my opinion.

jameslamb (changes requested):
I started reviewing tonight. It's late in my timezone so posting a partial review before I've finished the paper (I stopped right before `## Architecture: An Embedded Java Engine with Python Bindings`), so you have some initial feedback to look at.

Really interesting project! I'll try to review the rest of the paper soon.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+function call, and queries are method calls that return Python objects.
+
+**In-process JVM via JPype.** On first use, the binding starts an embedded JVM inside the
+Python process using JPype [@jpype], which bridges CPython and the JVM in a single address
+space and lets Python call Java methods directly. The engine's Java objects are exposed
+through a thin Pythonic facade, so application code never touches JPype directly. Because the
+JVM is in-process, there is no inter-process serialization on the query path: result rows
+cross the in-process boundary, not a network socket. The JVM is configured from Python via a
+keyword argument at database creation, notably the heap size, which matters for large vector
+indexes (discussed later):
```
jameslamb, line 129:
```suggestion
On first use, the binding starts an embedded JVM inside the Python process using JPype [@jpype].
JVM settings like heap size are configured directly from Python, and CPython and the JVM
share a single address space, so data can be passed around without needing network calls or
inter-process serialization. The engine's Java objects are exposed through a Pythonic layer,
so applications never need to interact with JPype directly.
```

What does "in-process boundary" mean?

Would you consider something like this proposed re-write? I think it's significantly simpler and more direct.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+the same query can be a transactional point operation (OLTP) or an analytical scan (OLAP). This
+single-handle design is what later lets a hybrid workflow thread results from one model directly
+into the next without leaving the process.
+
+**NumPy and pandas.** Scientific Python speaks arrays and frames, and the binding fits in.
+Embeddings are passed as `float32` NumPy arrays through a helper,
+`arcadedb.to_java_float_array(vec)`, that hands the buffer to the engine without a Python-side
+element-by-element copy; this is the path the vector and hybrid workloads use. Query results
+come back as ordinary lists of dict-like rows that drop straight into a `pandas.DataFrame`, and
+our workflows use pandas to load and prepare the source data. Bulk graph construction is
```
jameslamb, line 158:
```suggestion
come back as ordinary lists of dict-like rows that drop straight into a `pandas.DataFrame`. Bulk graph construction is
```

This clause beginning with "and our workflows" seems out of place in this paragraph... was it maybe included from some other work related to ArcadeDB?

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+available through a `graph_batch` context manager that amortizes edge creation. Together these
+keep the engine feeling native to a NumPy/pandas workflow rather than like a foreign service.
```
jameslamb, line 160: This paragraph is making a claim about how the API of `arcadedb-embedded` naturally fits with `numpy` / `pandas` codes... show us some code, please. Directly showing some code would make these easier to understand and might allow the description to be simplified.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+
+For bulk results the binding also exposes columnar exports, and the choice between them is
+about types rather than speed. `to_columns()` returns NumPy arrays, which fits until a column contains nulls. NumPy has no
+missing value for integers, so a nullable `INTEGER` must widen to `float64` with `NaN` holes.
+That loses the type and any precision above $2^{53}$, and a nullable boolean degrades to a
+Python list. `to_arrow()` reads the same buffer
+into a `pyarrow.Table`, keeping the validity bitmap the buffer already carries, so those
+columns stay `int64` and `bool`. It is not a second serialization path. The columnar transport already emits packed columns
+with a null bitmap, and Arrow's string layout is the same offsets-plus-blob representation. So
+the export adds no Java and no bytes to the wheel, and pyarrow is an optional extra.
```
jameslamb, line 170: I understand all the individual words in this paragraph but I'm not confident that I understand the paragraph itself.

I *think* the purpose here is to say:

* "the library can populate query results directly into `numpy` and `pyarrow` data structures"

and 

* "... but `numpy` and `pyarrow` are not required dependencies"

Is it roughly that? If so that's interesting and valuable, but I think the paragraph needs a full re-write for clarity.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+engine's `.jar` files are identical everywhere. The only platform-specific dependency is the
+Java runtime itself. `arcadedb-embedded` therefore *bundles a per-platform JRE* inside the
+wheel: the build produces wheels for four platforms (Linux x86-64, Linux ARM64, macOS Apple
+Silicon, and Windows x86-64), each carrying the matching JRE, across Python 3.10–3.14, for
+twenty wheels in total. The user runs `pip install arcadedb-embedded` and gets a working
+multi-model database with **no Java installation and no `JAVA_HOME`**, and with no server to
+deploy or operate. The JVM is an implementation detail sealed inside the package
+([](#fig-arch)). The cost is wheel size, about 67 MB per wheel (Linux x86-64; the other
+platforms are within a few MB), dominated by the runtime. We keep it small by trimming the JRE
+with `jlink` to only the modules the engine needs and by excluding JARs the engine does not
```
jameslamb, line 187: This needs a citation. I'm unsure what the best resource is, maybe https://openjdk.org/jeps/282 ?

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+require. That is the price of "no Java to install," paid once at install time.
+
+About 8 MB of the wheel is an optional in-process HTTP server with a web UI, inert unless the
+program calls `create_server()`. It is 12 JARs totalling 7.65 MB, plus 0.8 MB of JRE modules
+`jlink` pulls in only for them. It costs about 10 ms of one-time JVM startup for the longer
+classpath, and nothing measurable in resident memory. The wheels benchmarked here were built without it and
+are about 62 MB; embedded behaviour is identical either way.
```
jameslamb, line 194: I was very confused by this paragraph. I found it to be way over-specific in some places and lacking in critical information in otherp laces.

My biggest concern is that it seems to be saying "the wheels published to PyPI contain some optional components which affect startup time, but later in the paper I show benchmarks based on some custom build that doesn't have those components".

This paragraph left me as a reader wondering about some important questions:

* "well if you're presenting benchmarks from a custom build, what performance should I expect in my application where I `pip install arcadedb-embedded` and get a different build?"
* "what is the real-world purpose of `create_server()`? Why would I ever call it? Also haven't you been saying throughout this paper that not needing any servers or network calls is a defining benefit of this project?"

And instead of listing out the exact number of JARs and the compressed size to the nearest 10kb (details readers do not need), I recommend reframing these things in terms of the user benefit, for example "wheels are around 70 MB compressed, larger than projects like NumPy but not so large that they'd be prohibitively expensive to download over a weak internet connection or use in a space-constrained application like AWS Lambda".

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+inflated the vector step to ≈100 ms and the bulk load by ≈45×. Both are instances of the
+release cadence noted at the end of the benchmark section. All of it runs in a single process
+after a one-time bulk load. The
+three steps pass 200 vector candidates to the SQL filter, 50 survivors to the graph traversal,
+and return the top 10 answers. Timings were measured on the same host and 8-core cap as the
+comparison tables below.
+
+What makes this a *contribution*, not merely a convenience, is that no single
+Python-embeddable alternative can express it. [](#tbl-capability) lays out the capability
+matrix. SQLite has no graph or vector model, DuckDB has no transactional graph or Cypher,
```
jameslamb, line 323: This "{database1} has {x} but not {y}, {database2} has {y} but not {z}, etc." paragraph was already included earlier in the paper (https://github.com/scipy-conference/scipy_proceedings/pull/1256/changes#r3788607186)

The restatement here is not adding new information. I think one of them should be removed or converted to a reference to the other.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,651 @@
...
+Engines run their shipped defaults except where stated: SQLite is configured WAL +
+`synchronous=NORMAL` (its documented recommendation; the durability implications are stated
+with the tabular results), and ArcadeDB's JVM heap is pinned per tier.
+We report the median with the full [min–max] range: database benchmark repetitions are
+right-skewed (GC pauses, page-cache state), so the median resists outliers while the range
+exposes them, following established guidance for performance reporting
+[@raasveldt2018fair; @hoefler2015scientific]. A host-side sidecar samples each container's cgroup memory
+and CPU and reads the kernel peak. The data is the **Cross Validated** (`stats.stackexchange.com`)
+public data dump [@crossvalidated], a statistics and machine-learning Q&A corpus and the most
+SciPy-relevant Stack Exchange site, comprising 425,735 posts, 345,754 users, and 1,242,391
```
jameslamb, line 374:
```suggestion
public data dump [@crossvalidated], a statistics and machine-learning Q&A corpus, comprising 425,735 posts, 345,754 users, and 1,242,391
```

Remove "and the most SciPy-relevant Stack Exchange site". You can't substantiate that claim, and it's a distraction from the purpose of the paper.

jameslamb (changes requested):
I've finished a first review. I stopped leaving specific comments as I got further in because some of my earlier comments apply to multiple cases in the paper.

Overall, the project sounds really compelling and the topics covered are interesting! I wasn't familiar with ArcadeDB and found it pretty interesting. I could imagine myself wanting Python bindings to use it in exactly the way this paper describes.

However, I think that interesting content is drowned out by the overwhelming use of "AI voice" in this paper. I strongly suspect that significant portions of it were LLM-generated and only lightly edited. I see many signs of LLM-generated text here, including 0-information filler statements, excessively-specific descriptions, and "not X, but Y" clauses.

The one that I think may be most off-putting to this journal's audience is the use of forced-sounding references to SciPy and "scientific python". For example, the description in `myst.yml` says "An embedded, multi-model database **for scientific Python**". That isn't how the project describes itself in its own documentation. Looking in https://github.com/humemai/arcadedb-embedded-python, the only references to words like "science", "scientific", and "scientific Python" are limited to:

* code related to handling scientific notation for numeric types
* test code

That's it. The paper reads as if an LLM was told *"write a paper explaining `arcadedb-embedded` and comparing it to related projects. Make it relevant to participants in the SciPy conference."* This is at best distracting and at worst insulting. Even if that's not what happened, I stand by this suggestion... remove those things. The project being in Python, related to array/graph/vector data structures, and being interoperable with `numpy`, `pandas`, and `pyarrow` is more than enough to be relevant to this audience.

tae898:
Thanks for the kind review. I apologize for including a lot of LLM voice. A lot of my efforts were spent on the experiments and the source code of ArcadeDB python bindings, and this has made me be less attentive on the paper text. I'll fix them.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+  document store, a graph engine, and a vector index), and the data is copied between
+  them and kept consistent by application code. ArcadeDB is an Apache-2.0 multi-model
+  database engine, written in Java and so out of reach of a Python program that wants to
+  stay in one process. We present `arcadedb-embedded`, a Python package that runs the
+  engine inside the Python interpreter through JPype, over a Java runtime bundled in the
+  wheel. A single `pip install` provides documents, a property graph, and vector
+  search with no Java installation, and results come back as NumPy arrays, pandas
+  DataFrames, or Arrow tables. We demonstrate a hybrid retrieval workflow that chains
+  vector search, SQL filtering, and graph traversal over one dataset. We benchmark the engine from Python against SQLite, DuckDB,
+  LadybugDB, and Chroma, and report its costs.
```
jameslamb:
```suggestion
  them and kept consistent by application code. ArcadeDB is an open-source multi-model
  database engine that runs in-process and can support all of these query patterns, and here we
  present `arcadedb-embedded`, a Python client for it.
  ArcadeDB is written in Java. The Python client runs it inside a Python process via JPype, using
  a vendored Java runtime. The library supports returning results as NumPy arrays, pandas
  DataFrames, or Arrow tables. We demonstrate a hybrid retrieval workflow that chains
  vector search, SQL filtering, and graph traversal over one dataset. We benchmark the engine from Python against SQLite, DuckDB,
  LadybugDB, and Chroma, and report its costs.
```

Will you please consider this re-write? I think  itmore clearly conveys what the paper is about.

Some of my sources of confusion:

* the phrase "a single `pip install` provides documents" is confusing / misleading
* the reference to ArcadeDB and its license is kind of jarring here (there's no explicit transition explaining that that tool can be used to for all those workkloads)
* the phrase "the engine" is used throughout the paper but it's very unclear to me what that is... is that actually ArcadeDB itself (including its preferred query interface, result serialization, etc.)? Or is it some lower-lever `libarcadedb` or similar that you've built bindings to?

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+property graph (OpenCypher), and vectors (a graph index, searched with the
+`vectorNeighbors` function). The first argument of `query` selects the language. All three
+models share one transaction scope and one set of record identifiers, so the hybrid
+workflow below can pass results from one model into the next without leaving the process.
+
+**NumPy, pandas, and Arrow.** Embeddings are passed as `float32` NumPy arrays through
+`arcadedb.to_java_float_array(vec)`, which hands the buffer to the engine without a
+Python-side element-by-element copy. Results come back through four methods on the result
+set. `to_list()` gives a list of dicts, `to_dataframe()` a `pandas.DataFrame`,
+`to_columns()` a dict of NumPy arrays, and `to_arrow()` a `pyarrow.Table`. The benchmark suite linked at the end of the paper uses these same calls.
```
jameslamb:
```suggestion
`to_columns()` a dict of NumPy arrays, and `to_arrow()` a `pyarrow.Table`.
```

I don't think this sentence is adding information, recommend removing it.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+
+```python
+import numpy as np
+
+# db as above; embeddings is an (n, 384) float32 array, one row per Question;
+# the Question type and its vector index are created in the next section.
+seed = np.asarray(embeddings[0], dtype=np.float32)
```
jameslamb: > _embeddings is an (n, 384) float32 array,_

It's at this point then that I can no longer test the code and that anyone reading the paper has to do manual work to try to keep testing by themselves.

It'd be useful to try to make these self-contained and copy-pastable.

If the code to generate the data inline is too cumbersome for the paper, maybe you could have it load a publicly-available real-world dataset, like from `sklearn.datasets` or HuggingFace datasets via https://pypi.org/project/datasets/ or something?

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+hits = db.query("sql",
+    "SELECT id, distance FROM (SELECT expand(vectorNeighbors(?, ?, ?, ?)))",
+    "Question[embedding]", arcadedb.to_java_float_array(seed), 10, 100)
+df = hits.to_dataframe()                 # pandas.DataFrame
+
+posts = db.query("sql", "SELECT id, score, title FROM Question")
+cols = posts.to_columns()                # {"id": ndarray, "score": ndarray, "title": list}
+table = db.query("sql", "SELECT id, score, title FROM Question").to_arrow()
+                                         # pyarrow.Table; nullable INTEGER stays int64
+```
```
jameslamb: 2 comments about this code sample.

1. Why is it necessary to execute the query twice? That deserves explanation. For example, is the query not actually executed in the database until some output-requesting function like `.to_arrow()` is run? Similar to how a Spark query won't run until a `.collect()` or similar?
2. the `.to_columns()` example assigns the query to a variable and then the `.to_columns()` output to a separate variable, but the `.to_arrow()` is a one-liner... is that necessary? If not, make them consistent (I am indifferent on which pattern, depending on the answer to Question 1)

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+`to_arrow()` reads the same buffer and keeps its null bitmap, so the column stays `int64`.
+Both read one packed columnar buffer from the engine, and the Arrow path needs no extra
+Java code in the wheel. [](#sec-transport) measures the four export paths.
+
+**Cross-platform packaging.** Bringing an engine written in another language into Python
+usually means compiling a native C/C++ extension, with tooling such as scikit-build-core
+[@schreiner2024skbuild]. `arcadedb-embedded` ships a JVM instead. Java bytecode is
+platform-agnostic, so the engine's `.jar` files are identical everywhere and the only
+platform-specific part is the Java runtime.
+`arcadedb-embedded` bundles a per-platform JRE inside the wheel, trimmed with `jlink`
```
jameslamb:
```suggestion
`arcadedb-embedded` bundles a per-platform Java Runtime Environment ("JRE") inside the wheel, trimmed with `jlink`
```

This is by my count the 6th time in the paper already that the use of a vendored JRE is mentioned. It'd be helpful to consolidate those if possible.

The paper is also inconsistent between saying there's a vendored "JVM", "JRE", and "Runtime". It would be worthwhile to clarify that (the most technically correct would be to say that the JRE is vendored and used to create a JVM at runtime, I think).

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+The document model is created and queried with SQL, and writes run inside ACID
+transactions. The snippet in [](#sec-arch) creates a type and a `UNIQUE_HASH` index for
+point lookups, then runs a parameterized `INSERT` inside `db.transaction()`. Records can
+be document-typed (schema-flexible) or property-typed for indexing, and both are stored in
+the same database as the graph and vectors below.
+
+Loading a corpus row by row pays the language-boundary cost once per row, so bulk ingest
+goes through `insert_many`, which serializes a batch to one string and loops over it on
+the Java side. As with the graph below, the columns a type does not need are never read
+into Python:
```
jameslamb:
```suggestion
To minimize the impact of per-transaction overhead, bulk ingestion
goes through `insert_many`, which serializes a batch to one string and loops over it on
the Java side. As with the graph below, the columns a type does not need are never read
into Python:
```

I'm confident there is other overhead in each transaction besides whatever "the language-boundary cost" is (that is not explained in the paper so far). Would you consider this simplification?

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+
+### Property graph
+
+Vertex and edge types are declared with SQL. OpenCypher creates them implicitly on first
+write, as in Neo4j, but we declare them so that properties can be typed and indexed. Bulk
+loading goes through a `graph_batch` context manager. `create_vertices` and `new_edges` each cross into the JVM once for a
+whole list and the batch commits on flush, instead of one crossing and one transaction per
+element. `batch_size` is the number of edges buffered before an automatic flush, and
+`create_vertices` returns the record ids, so edges reuse them instead of querying them
+back. Reading the corpus column by column keeps the same discipline on the pandas side.
```
jameslamb:
```suggestion
back.
```

This sentence is vague (what "discipline"?) and looks like another LLM artifact to me. I recommend removing it.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+Relationship queries then use OpenCypher through the same handle:
+
+```python
+import pandas as pd
+
+for t in ("Question", "Answer", "Userx"):
+    db.command("sql", f"CREATE VERTEX TYPE {t}")
```
jameslamb:
```suggestion
db.command("sql", f"CREATE VERTEX TYPE Answer")
db.command("sql", f"CREATE VERTEX TYPE Question")
db.command("sql", f"CREATE VERTEX TYPE Userx")
```

The `for` loop only saves 1 line here... I think it's be simpler and clearer to just make it literal `db.command()`, just like the edge type commands below.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+db.command("sql", "CREATE EDGE TYPE HAS_ANSWER")
+db.command("sql", "CREATE EDGE TYPE AUTHORED_BY")
+
+data = "datasets/prepared/stats.stackexchange.com"
+
+# one column-selected read per table; only these columns cross into Python
+questions = pd.read_parquet(f"{data}/questions.parquet", columns=["id", "score"])
+answers   = pd.read_parquet(f"{data}/answers.parquet", columns=["id", "score", "author"])
+users     = pd.read_parquet(f"{data}/users.parquet", columns=["id", "reputation"])
+answer_author = answers.pop("author").tolist()
```
jameslamb: These code examples all show a data ingestion path for Parquet like:

* read into a `pandas` Dataframe
* convert to Python objects (`.todict()` / `.tolist()`)
* pass to ArcadeDB

Just at face value that seems prohibitively expensive relative to using something like DuckDB that can query over Parquet files directly (taking advantage of their data layout and schema to make reads efficient).

Are these examples supposed to show the preferred ingestion path to do analysis on a bunch of Parquet files? Or are they trying to demonstrate an analysis that just starts from having `pandas` DataFrames?

1. that should be clarified
2. a section dedicated to data ingestion should be added to the paper. That feels like a significant missing part of this right now... it's unclear how users should get their data into ArcadeDB, and that's a big part of story for the end-to-end time it takes to run an analysis on a dataset

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+    f"WHERE q.id IN {fids} "
+    f"RETURN q.id AS qid, ans.id AS aid, ans.score AS ascore, "
+    f"usr.reputation AS rep ORDER BY ascore DESC LIMIT 10").to_list()
+```
+
+Over the complete Cross Validated corpus (213,761 questions, 208,986
+answers, 108,101 linked users), the workflow runs warm in ≈13 ms (median over 20 reps
+after 5 warmups), of which vector ≈5.9 ms, SQL ≈5.0 ms and Cypher ≈2.0 ms. Nineteen of the twenty reps
+fall between 11.4 and 16.2 ms. The twentieth, 31.8 ms, includes one SQL step that stalled
+to 24 ms, the managed-runtime tail in [](#tbl-latency), and is included in the median.
```
jameslamb: Just say "Over 20 repetitions, the runtime ranged from `{min}` to `{max}`, with a median of `{median}` and mean of `{mean}`".

Listing those summary stats for all but the worst case and then saying "oh but the worst case is still in the media" is really confusing.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+Timings were measured on the same host and 8-core cap as the comparison tables below.
```
jameslamb:
```suggestion
Timings were measured on the same system used to compute other benchmarks shown in the comparison tables below. See {reference} for specific details on the environment.
```

I recommend pushing any details about the environment that the benchmarks were run on down into footnotes or an appendix. At this point in the paper, "the same ... 8-core cap ..." is an unnecessary and distracting level of detail.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+
+A GAV accelerates the Cypher traversal (measured at 2.4× on the analytical suite, see
+below). The same traversal written in ArcadeDB's SQL `MATCH` answers in ≈3 ms, and both languages run the
+same traversal over the same storage.
```
jameslamb: What is the goal of his paragraph?

I'm not familiar with GAV so maybe I am just misunderstanding, but to me it reads as "the numbers above could have been even better if I had done `{x}`" which leaves me thinking..... "well then why didn't you just do that?".

Recommend either explaining that clearly here or removing this paragraph.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+
+## Benchmarks
+
+Is each model competitive enough for practical use, and what does it cost? We compare
+ArcadeDB, *as reached from Python*, against the strongest Python-embeddable
+specialist in each lane, namely SQLite and DuckDB (tabular), LadybugDB (graph), and
```
jameslamb: "the strongest" implies 1 choice, it was confusing to see both SQLite and DuckDB here. I recommend either changing "the strongest" to something like "popular specialized tools" or dropping one of those tools.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+environment file (link at the end of the paper).
+
+All runs were on a single host, a 12th-gen Intel Core i9-12900HK with 20 logical cores, of
+which 8 reached each container via `--cpuset-cpus 0-7`. It has 61 GiB usable RAM and a
+Samsung 980 PRO 2 TB NVMe SSD (PCIe 4.0) holding the databases and datasets, on Linux
+kernel 7.0.0 (x86-64) and Docker 29.5.3. The versions are `arcadedb-embedded` 26.8.1 (the
+PyPI wheel), DuckDB 1.5.4, LadybugDB (`ladybug`) 0.18.1, Chroma 1.5.9, and SQLite 3.46.1,
+the version Python's `sqlite3` module binds in the `python:3.12-slim` image the containers
+are built from. ArcadeDB releases roughly monthly and `arcadedb-embedded` packages each
+release, so these numbers describe 26.8.1 rather than a fixed ceiling. Embeddings are
```
jameslamb:
```suggestion
are built from. ArcadeDB and `arcadedb-embedded` v26.8.1 were used. Embeddings are
```

This sentence seems really out of place here, and I also don't understand it.

The rest of the paragraph is filled with highly-specific detail of the form "I used version `{v}` or `{project}`". This should also just plainly state the version of ArcadeDB and the Python client.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+384-dimensional and computed in the binding's repository with `all-MiniLM-L6-v2` over each
+question's title and de-HTML'd body, L2-normalized, stored as `float32`.
+
+**Workloads.** OLTP is a mixed point-operation workload issued by id and reported as
+sustained throughput (ops/s). The tabular lane issues 5,000 operations (60% reads, 20%
+updates, 10% inserts, 10% deletes) and the graph lane 2,000 (50% point lookups, 35% one-hop
+traversals, 15% vertex inserts). OLAP is a fixed suite of analytical queries. Each query is timed
+seven times and averaged, the averages are summed into a suite time, and that suite time is
+reported as the median over the five repetitions, like every other cell. The tabular suite is five `GROUP BY`/aggregate/top-N
```
jameslamb: I'm finding this description really really difficult to understand. Is this standard for database benchmarks?

The level of indirection is really intimidating. There are multiple different levels of repetition and aggregation here, making it pretty hard for me to understand what the reported timings mean.

I think this would be more impactful if it were simpler... run the end-to-end workloads `n` times, report the summary statistics on the timings, share all the queries in an appendix.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+against LadybugDB's ≈684 MiB. The on-disk graph is ≈44× LadybugDB's columnar store
+(≈1,841 vs ≈41 MiB).
+
+**Vector ([](#tbl-vector)).** At matched graph degree (`maxConnections` $= 2M$), recall@10
+is higher than Chroma's (0.980 vs 0.972). Build time is longer (≈546 s vs ≈319 s for
+1.24 M vectors) and query latency ≈3.5× higher (≈4.2 ms vs ≈1.2 ms), still single-digit
+milliseconds. Peak memory is lower than Chroma's (≈21.1 GiB vs ≈24.6 GiB, 14% less), because
+the engine keeps vectors on disk while Chroma's hnswlib index holds the whole set in RAM.
+
+:::{table} Vector lane (Cross Validated corpus, 1,242,391 vectors): Chroma, ArcadeDB at matched HNSW degree (Chroma $M=16$, ArcadeDB `maxConnections`=32, both `ef_construction`=100, `ef_search`=100). Build = insert+index (s), Query = mean latency per query in a rep (ms), recall@10 vs exact ground truth. Median [min–max] over 5 reps. Peak = container memory, DB = on-disk size (MiB).
```
jameslamb:
```suggestion
:::{table} Vector lane (Cross Validated corpus, 1.2M+ vectors): Chroma, ArcadeDB at matched HNSW degree (Chroma $M=16$, ArcadeDB `maxConnections`=32, both `ef_construction`=100, `ef_search`=100). Build = insert+index (s), Query = mean latency per query in a rep (ms), recall@10 vs exact ground truth. Median [min–max] over 5 reps. Peak = container memory, DB = on-disk size (MiB).
```

Choosing this place on the diff to leave a general comment about this "Benchmarks" section. I really think it'd be helpful to do a pass over this section for simplicity. Totally makes sense to me that you want to be extremely precise about details that are needed to reproduce the benchmark (like the exact versions of software used) but Things like `Vector lane (Cross Validated corpus, 1,242,391 vectors)` could be `Vector lane (Cross Validated corpus, 1.2M+ vectors)` or similar without negatively affecting the paper.

Anything you could do to soften this section and make it more easily understandable would be helpful. As currently written, it's extremely dense and I'm struggling to understand the answer to the questions from the beginning (_"Is each model competitive enough for practical use, and what does it cost?"_ ).

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+≈6,400 (≈14×) on the mixed point workload, while ArcadeDB in turn runs ≈29× DuckDB's
+fully-durable ≈219. Under the *strict* pairing, per-commit fsync for both
+(`arcadedb.txWalFlush=2`, and SQLite at its library defaults of a rollback journal with
+`synchronous=FULL`), the two converge on the disk's fsync floor, ≈262 against ≈187 ops/s.
+The SQLite strict figure is in the suite's append log rather than the
+curated per-run table. Comparing each engine at its own default would mostly compare
+durability settings, so both pairings are reported. On analytical SQL DuckDB answers the
+suite in ≈9 ms versus ArcadeDB's ≈1,400 ms.
+
+:::{table} Tabular lane (Cross Validated corpus): SQLite, DuckDB, ArcadeDB. OLTP is a mixed point read/insert/update workload (ops/s, higher better), OLAP an analytical aggregation suite (ms, lower better). Durability: SQLite WAL+NORMAL and ArcadeDB async WAL are bounded-loss, DuckDB fsyncs per commit. Median [min–max] over 5 reps. Peak = container memory, DB = on-disk size at the end of the OLTP rep (MiB).
```
jameslamb: 1. Would you consider using the term "query pattern" or "query type" instead "lane"? I think that'd be much clearer, and help to differentiate from "workload".
2. In an earlier comment I mentioned that you'd used "record" and "document" interchangeably. Here's now a third word used to refer to the same workload ("tabular"). It'd be helpful to choose one phrase and use it throughout the paper... that'd help readers to connect the concepts they learned in the introductory sections to the results they see in the benchmarking section.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+
+
+
+(sec-transport)=
+### Result transport into Python
+
+The same engine runs the same query in all four arms, and only the way results cross
+into Python differs. `iter_dicts()` converts one row at a time, `to_json_list()`
+serializes one JSON string per batch for Python to parse, `to_columns()` reads a packed
+columnar buffer with `numpy.frombuffer`, and `to_arrow()` wraps that same buffer as a
```
jameslamb:
```suggestion
columnar buffer with `numpy.frombuffer()`, and `to_arrow()` wraps that same buffer as a
```

minor style note: in most other places, inline markup of functions has a trailing `()` (<-- I like this!). This `numpy.frombuffer()` reference should too. It's worth checking for other cases like that, I think it's informative.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+| 1,000 | 20.12 | 2.97 | 2.28 | 1.49 |
+| 10,000 | 197.29 | 25.87 | 15.87 | 14.20 |
+| 100,000 | 1,978.27 | 278.22 | 135.51 | 109.03 |
+:::
+
+At 100k rows ([](#tbl-transport)) `iter_dicts` is 18× slower than `to_arrow` (1,978
+against 109 ms), and the whole difference is on the Python side. At 10 rows every arm
+finishes under 2.1 ms, so the choice only starts to matter as results grow.
+
+`to_arrow()` is also the fastest arm at every size, 0.65–0.90× of `to_columns()`, because
```
jameslamb: > "fastest arm at every size"

I don't understand this phrasing of "arm". Is this a baseball metaphor? It'd be helpful to use plain language here ("fastest method"?).

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+| | dtype | integer? | missing |
+|---|---|---|---|
+| `to_columns()` | `float64` | no | 1,000 `NaN` |
+| `to_arrow()` | `int64` | yes | 1,000 nulls |
+
+A Python user cannot change how the engine works, but they do choose the transport, and on
+a large result that choice costs more than the engine does. That is why it is measured
+here.
+
+## When to use it
```
jameslamb, line 533: This"When to use it" section did not hold the content I was hoping for.

The "where it fits" part is a short summary of the topics from the abstract  and then the "Costs to plan for" is a collection of overly-precise facts measured during the benchmarking.

The "Benchmarks" section was very dense and I was expecting this "When to use it" section to offer high-level takeaways from it to help users understand whether to adopt `arcadedb-embedded`. I was expecting statements like _"`arcadedb-embedded` is most useful in multi-model applications on small datasets, where reproducibility and correctness are more important than maximizing throughput"_, not _"each wheel is 70 MB"_.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+**Where it fits.** Local, single-machine work that touches more than one data model over
+the same data, such as retrieval-augmented prototypes that need documents, graph, and embeddings,
+and record- or graph-oriented applications that also want vector search beside the data.
+
+**Costs to plan for.** Memory is the baseline above plus an enlarged heap for large vector
+indexes, set from Python via `jvm_kwargs`. Startup is a one-time ≈0.3 s on the first
+operation. JPype calls hold the global interpreter lock (GIL), so Python-driven
+parallelism is limited as with any native extension, though engine-internal parallelism
+is unaffected. Python results cost ≈1.3×
+(vector search) and ≈1.6× (full-table scan) a pure-Java baseline on the same engine, the
```
jameslamb: What does "cost" mean here?

If this is only about run time, it'd be clearer to write something like the following:

```text
Queries run via `arcadedb-embedded` are generally 1.3x-1.6x
slower than the same queries run using ArcadeDB directly
in a Java application.
``` 

This seems like a really important point, it is worth clarifying.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+difference being result materialization. Each wheel is ≈70 MB, downloaded once.
+
+## Conclusion
+
+Applications that need records, relationships, and vectors over the same data usually run
+three systems and application code to keep them consistent. ArcadeDB can be packaged as a
+Python wheel with a bundled JVM, and one engine then provides documents, graph, and
+vectors from one process. The hybrid workflow chains vector search, SQL filtering, and
+graph traversal without data leaving the engine. The comparison puts each specialist ahead on
```
jameslamb: This is 3 consecutive sentences that describe the exact same thing with 3 different terminologies and in an inconsistent order:

1. "records, relationships, and vectors"
2. "documents, graph, and vectors"
3. "vector search, SQL filtering, and graph traversal"

It seems like a superset of the competing terminologies used throughout the paper (maybe this set + "tabular"), pointing it out so you could use it as a reference of things to look for if you do an editing pass for standardizing on terms. I think the paper would benefit from that.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+its own ground, leaves the engine competitive on the rest, and prices the difference mainly
+in memory. The ArcadeDB project wrote the engine. This paper contributes the
```
jameslamb: I do not understand that phrase "prices the difference mainly in memory".

This feels to me like it may be another LLMism.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+|---|---|--:|--:|--:|
+| tabular read | SQLite | 0.004 | 0.007 | 0.05 |
+| tabular read | DuckDB | 0.92 | 1.89 | 2.9 |
+| tabular read | ArcadeDB | 0.06 | 0.17 | 34.9 |
+| graph point | LadybugDB | 0.41 | 1.24 | 1.8 |
+| graph point | ArcadeDB | 0.18 | 0.74 | 1.5 |
+| graph hop | LadybugDB | 1.44 | 4.34 | 11.0 |
+| graph hop | ArcadeDB | 0.21 | 0.81 | 90.3 |
+| vector query | Chroma | 1.18 | 1.36 | 1.4 |
+| vector query | ArcadeDB | 3.86 | 7.11 | 145.9 |
```
jameslamb:
```suggestion
| Lane / op | Backend | p50 | p99 | max |
|---|---|--:|--:|--:|
| tabular read | SQLite | **0.004** | **0.007** | **0.05** |
| tabular read | DuckDB | 0.92 | 1.89 | 2.9 |
| tabular read | ArcadeDB | 0.06 | 0.17 | 34.9 |
| graph point | LadybugDB | 0.41 | 1.24 | 1.8 |
| graph point | ArcadeDB | **0.18** | **0.74** | **1.5** |
| graph hop | LadybugDB | 1.44 | 4.34 | **11.0** |
| graph hop | ArcadeDB | **0.21** | **0.81** | 90.3 |
| vector query | Chroma | **1.18** | **1.36** | **1.4** |
| vector query | ArcadeDB | 3.86 | 7.11 | 145.9 |
```

For all of these tables, I recommend helping users by bolding or in some other way visually emphasizing the best results in each grouping.

Today, this table renders like this:

Image

With the bolding, it'd be much easier to see the interesting patterns, which to me look like:

* on "tabular read", SQLITE has MUCH lower latency than DuckDB and ArcadeDB
* on "graph hop", ArcadeDB has lower median and p99 latency than ArcadeDB, but its worst-case performance ("max") was MUCH worse

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,584 @@
...
+Graph point and 1-hop p99 (0.74 ms, 0.81 ms) beat LadybugDB's (1.24 ms, 4.34 ms), and
+tabular read p99 (0.17 ms) beats DuckDB's (1.89 ms) ([](#tbl-latency)). SQLite's
+memory-mapped reads at 0.007 ms are far ahead of everything in the table. The tail is
+longer, though. ArcadeDB's worst single operation runs 35–146 ms across the three lanes
+against 0.05–11 ms for the specialists, which matters under a hard per-request bound but
+not under a p99 target.
+
+:::{table} Per-operation query latency on the Cross Validated corpus (ms, median of per-rep summaries over 5 reps): median (p50), tail (p99), and worst single operation (max).
+:label: tbl-latency
+| Lane / op | Backend | p50 | p99 | max |
```
jameslamb:
```suggestion
| Lane / op | Backend | median | p99 | max |
```

The text of the paper uses the word "median". Use that in the table too, not `p50`.

jameslamb (changes requested):
## Conclusion

To @cdlindsey and the rest of the Proceedings committee, my final recommendation is **NO, do not include this paper as-is in the Proceedings**.

---

## Why so definitive?

I received an email from the proceedings committee that said the following:

> _please let us know if you approve this submission as it is, in the Proceedings. The deadline for authors to work on your comments was August 30th... Please let us know your decision by September 6th; if it is yes or no._

I see you pushed changes on August 29th: https://github.com/scipy-conference/scipy_proceedings/pull/1256/changes/fcba75107e93db813c6eddce42de5c1b902f43bd

I reviewed thoroughly tonight in response to that.

Details below.

---

### Why a No?

See the review comments attached to this review. I think the paper is much improved from when I last reviewed on August 15th, but the sum total of the criticisms in those comments is too large, in my opinion, to recommend inclusion in the Proceedings.

I do think `arcadedb-embedded` seems like a very interesting project and could see a paper on it being a great addition to the Proceedings! Maybe with a few more review-edit-review feedback cycles that could have happened this year for this paper.

Short summary:

* inconsistent use of terminology, making it hard to connect the different sections (example: "records" and "documents" used interchangeably)
* "Benchmarks" section is excessively detailed, and I strongly suspect it was mostly AI-generated without a manual editing pass.
  - _It's several consecutive dense paragraphs of numbers without any plain-language explanation of the result for readers (in any section)._
  - _The sentence about it in the Conclusion is very unclear: "The comparison puts each specialist ahead on its own ground, leaves the engine competitive on the rest, and prices the difference mainly in memory."_
* unclear language throughout

---

### A note on AI use

Throughout these reviews, I've commented on "LLMisms" and AI use. I just want to repeat what I wrote in https://github.com/scipy-conference/scipy_proceedings/pull/1256#discussion_r3788541889

> _[AI use] is allowed under the proceedings' AI use policy (https://github.com/scipy-conference/scipy_proceedings/blob/2026/genai_policy.md). But from this point forward I think the paper would really benefit from your making remaining edits manually and from you doing a copy-editing pass for reducing the "AI voice" in the paper._

I don't think that happened here. I find it very unlikely that you authored https://github.com/scipy-conference/scipy_proceedings/pull/1256/changes/fcba75107e93db813c6eddce42de5c1b902f43bd manually, given the excessive detail in the commit message.

I am not a "No" on this paper because AI was used for the writing, but I do think that in this case the paper would have greatly benefited from more manual review and copyediting for simplicity, clarity, and consistency.

tae898:
@jameslamb 

Thanks for the thorough review. I'd like to address the points, but is the review window over for this year then?

tae898:
> ## Conclusion
> To @cdlindsey and the rest of the Proceedings committee, my final recommendation is **NO, do not include this paper as-is in the Proceedings**.
> 
> ## Why so definitive?
> I received an email from the proceedings committee that said the following:
> 
> > _please let us know if you approve this submission as it is, in the Proceedings. The deadline for authors to work on your comments was August 30th... Please let us know your decision by September 6th; if it is yes or no._
> 
> I see you pushed changes on August 29th: [fcba751](https://github.com/scipy-conference/scipy_proceedings/commit/fcba75107e93db813c6eddce42de5c1b902f43bd)
> 
> I reviewed thoroughly tonight in response to that.
> 
> Details below.
> 
> ### Why a No?
> See the review comments attached to this review. I think the paper is much improved from when I last reviewed on August 15th, but the sum total of the criticisms in those comments is too large, in my opinion, to recommend inclusion in the Proceedings.
> 
> I do think `arcadedb-embedded` seems like a very interesting project and could see a paper on it being a great addition to the Proceedings! Maybe with a few more review-edit-review feedback cycles that could have happened this year for this paper.
> 
> Short summary:
> 
> * inconsistent use of terminology, making it hard to connect the different sections (example: "records" and "documents" used interchangeably)
> * "Benchmarks" section is excessively detailed, and I strongly suspect it was mostly AI-generated without a manual editing pass.
>   
>   * _It's several consecutive dense paragraphs of numbers without any plain-language explanation of the result for readers (in any section)._
>   * _The sentence about it in the Conclusion is very unclear: "The comparison puts each specialist ahead on its own ground, leaves the engine competitive on the rest, and prices the difference mainly in memory."_
> * unclear language throughout
> 
> ### A note on AI use
> Throughout these reviews, I've commented on "LLMisms" and AI use. I just want to repeat what I wrote in [#1256 (comment)](https://github.com/scipy-conference/scipy_proceedings/pull/1256#discussion_r3788541889)
> 
> > _[AI use] is allowed under the proceedings' AI use policy (https://github.com/scipy-conference/scipy_proceedings/blob/2026/genai_policy.md). But from this point forward I think the paper would really benefit from your making remaining edits manually and from you doing a copy-editing pass for reducing the "AI voice" in the paper._
> 
> I don't think that happened here. I find it very unlikely that you authored [fcba751](https://github.com/scipy-conference/scipy_proceedings/commit/fcba75107e93db813c6eddce42de5c1b902f43bd) manually, given the excessive detail in the commit message.
> 
> I am not a "No" on this paper because AI was used for the writing, but I do think that in this case the paper would have greatly benefited from more manual review and copyediting for simplicity, clarity, and consistency.

Thanks James again for the thorough review.

- I used interchangeable terms, e.g. records vs. documents, but I can see that it is confusing. I'll pick one and use it throughout.
- The benchmarks section is detailed on purpose. A database embedded in Python is not a common thing, so I wanted a thorough benchmark so people know where it stands against other DBs. But you are right that the numbers can be more easily explained. I'll add a plain takeaway per table and move the protocol detail to footnotes.
- On AI use: the acknowledgements disclose it, as the policy asks. I want to say again that all the core thoughts, the experiments, and the structure are mine.

papers/taewoon_kim/main.md
tae898: - Yes, "a single `pip install` provides documents ..." was confusing. I'll change it to "one `pip install` gives a database with documents, a property graph, and vector search, and needs no Java installation".
- Good point on Apache-2.0. I'll drop it from the abstract and mention each license once: ArcadeDB's in the intro and the binding's in the conclusion.
- "The engine" is ArcadeDB itself, the unmodified Java jars from https://github.com/ArcadeData/arcadedb. There is no lower-level `libarcadedb`. My package (https://github.com/humemai/arcadedb-embedded-python) adds the JVM lifecycle, the bundled runtime, and the conversion of results into NumPy, pandas, and Arrow. So the queries are ArcadeDB's and the result serialization is mine. I'll define this once in the intro so the term is clear after that.
- One change to your suggestion: "Python client" to "Python package". ArcadeDB can also run as a server with an HTTP API, but this package is not about that. It loads the engine into the Python process and talks to it through JPype, so there is no server. It is closer to `sqlite3` than to `psycopg2`.

papers/taewoon_kim/main.md
tae898: Fair point. Fully self-contained snippets would not fit the page: the workflow needs documents, a who-answered-what graph, and embeddings from the same data, and no ready-made sklearn or HuggingFace dataset ships all three, so the loading code alone is longer than the paper allows. The full code was already in the benchmark suite linked at the end of the paper, but I made it easier: one command now rebuilds the data and runs this section.

papers/taewoon_kim/main.md
tae898: The query runs when `db.query()` is called, so it is not lazy like Spark. It returns a cursor, and each `to_*()` reads that cursor once, so the same rows in a second format need a second query. I made both lines one-liners and added a comment saying this.

papers/taewoon_kim/main.md
tae898: Agreed on all three. The bundled JRE is now mentioned in three places that are read on their own: the abstract, the contribution list (where JRE and JVM are spelled out), and the figure caption. The packaging paragraph has the details. And the terms are fixed: the wheel bundles a JRE, and the binding starts a JVM from it at first use. Two places said "bundled JVM", which was wrong, and "Java runtime" is gone.

papers/taewoon_kim/main.md
tae898: Took the simplification, with one change: the cost that `insert_many` removes is per row, not per transaction. Loading one row at a time makes one Python-to-Java call and one statement parse per row. So the sentence now says "pays a Python-to-Java call per row".

papers/taewoon_kim/main.md
tae898: Removed.

papers/taewoon_kim/main.md
tae898: applied

tae898: I haven't pushed the commit yet. I'll continue tomorrow.

papers/taewoon_kim/main.md
tae898: Good question. pandas is only there to read the Parquet files. I added an "Ingestion" subsection that explains how data gets in and gives the load times, and an Ingest column to the graph table. The load is slower than DuckDB and SQLite, but most of that is the unique hash index, not the Python path.

papers/taewoon_kim/main.md
tae898: applied.

papers/taewoon_kim/main.md
tae898: applied

papers/taewoon_kim/main.md
tae898: The Cypher step in that measurement already runs over a GAV, so the paragraph now just says that. GAV is explained once, in the Property graph section, and the graph benchmark measures what it is worth.

papers/taewoon_kim/main.md
tae898: Changed to "widely used Python-embeddable specialists"

papers/taewoon_kim/main.md
tae898: The comparator versions are in a footnote now. I kept one sentence saying both projects release monthly, since that matters for how to read the numbers, and it now says plainly that the numbers are for 26.8.1.

papers/taewoon_kim/main.md
tae898: Simplified. The main text now just says each workload ran in five fresh containers and we report the median, and the details are in a footnote. Timing each query a few times within a run, then repeating the whole run, is the usual way for database benchmarks (the paper cites Raasveldt et al. for it). The full query text stays in the linked suite, since an appendix would count toward the word limit.

papers/taewoon_kim/main.md
tae898: Agreed. I went through the whole section again. Each result paragraph now opens with the answer in plain words and keeps only the few numbers that back it up. The durability settings and other details went to footnotes.

papers/taewoon_kim/main.md
tae898: Done. The three models are called documents, graph, and vectors throughout, in the examples and in the benchmark tables. "Lane" is now "model", so the benchmark section uses the same names as the section that introduces them. "Tabular" and "record" are gone.

papers/taewoon_kim/main.md
tae898: Applied, and swept the rest of the paper to find more.

papers/taewoon_kim/main.md
tae898: changed to "method"

papers/taewoon_kim/main.md
tae898: You're right. The section now says when `arcadedb-embedded` is the right choice, when a single-model tool is better and why, and what it costs. The wheel size and the other small numbers are gone.

papers/taewoon_kim/main.md
tae898: It means time. I moved this to the end of the result transport section and wrote it plainly: Python is a thin layer over the engine, the query itself does the same work either way, and a vector search takes about 10% longer from Python than from Java. The overhead only shows up on large results, and the transport table shows how much the method used to return them matters.

papers/taewoon_kim/main.md
tae898: Done. It's documents, graph, and vectors everywhere now, in that order. "Records", "relationships", and "tabular" are gone.

papers/taewoon_kim/main.md
tae898: Rewritten in plain words.

papers/taewoon_kim/main.md
tae898: Applied to all tables. Best values are in bold.

papers/taewoon_kim/main.md
tae898: applied.

papers/taewoon_kim/main.md
tae898: One more thing on this: while checking, I found the snippet did not match the files the suite produces. It is fixed, and all the snippets now run fine.

papers/taewoon_kim/main.md
tae898: Kept just the median in the end, to match how the rest of the paper reports.

jameslamb:
> _is the review window over for this year then?_

This I'm not sure of, hopefully @cdlindsey can tell us. I would be willing to review again if that's permitted but it's possible the paper has missed the window this year. Let's see what he says.

cdlindsey:
yeah, email proceedings@scipy.org if I haven't gotten back (I'm overseas and wifi is not predictable), I'm checking window

tae898:
@jameslamb @cdlindsey 

Thank you for the extra time you put for my paper. The two rounds of reviews helped me shape the paper better, and I think now it reads much better as well.

chongshenng:
Hello @tae898, I'm Chong Shen and I'll help @cdlindsey with editorial duties for this paper given that he is unavailable at the moment. 

@jameslamb Can you please review the revisions and make a decision?

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,579 @@
...
+for assembling one from interoperable Python libraries. It has costs when all three access
+patterns apply to the same data. Each system holds its own copy, with its own consistency
+model, so every write has to reach all of them. A query that spans models passes identifiers from one library to another. Reproducing
+an experiment means pinning several packages instead of one.
+
+This paper describes the alternative, one embedded multi-model database that holds
+documents, a property graph, and vector indexes. That database is ArcadeDB [@arcadedb], an
+Apache-2.0 Java engine that keeps all three models in one store. In this paper
+"the engine" always means ArcadeDB itself, its unmodified Java code including its query
+languages; there is no lower-level library. Documents are queried with
```
jameslamb:
```suggestion
documents, a property graph, and vector indexes. That database is ArcadeDB [@arcadedb], an
Apache-2.0 Java application that keeps all three models in one store (referred to as "the engine" throughout the rest of the paper). Documents are queried with
```

This "there is no lower-level library" seems out of place here. It makes sense as a response to my earlier request for clarification on use of the word "engine" throughout, but it's unnecessary and distracting within the paper. Recommend removing it.

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,579 @@
...
+its SQL dialect, the graph with OpenCypher, and vectors through a graph index of the
+Hierarchical Navigable Small World (HNSW) family. Transactional (online transaction processing, OLTP) and analytical (online analytical
+processing, OLAP) queries run over the same data. ArcadeDB is ACID (atomic, consistent,
+isolated, durable). Its Graph Analytical Views (GAV), an ArcadeDB-specific structure, speed
+up graph analytics while the base graph stays transactional.
+
+`arcadedb-embedded` binds the Java engine through JPype [@jpype; @arcadedbpython]. The
+benchmarks report the tasks on which the engine is slower than the specialists.
```
jameslamb:
```suggestion
`arcadedb-embedded` binds the Java engine through JPype [@jpype; @arcadedbpython].
The paper contains benchmarks comparing `arcadedb-embedded` to the specialists.
```

Just keep this generic. "report the tasks on which [it] is slower" is confusing because it makes me think "well are there other tasks where it's FASTER? where are those reported?"

papers/taewoon_kim/main.md
```diff
@@ -0,0 +1,579 @@
...
+Data enters all three models as Python objects: `insert_many()` takes a list of dicts
+for documents, `graph_batch()` takes lists of dicts and ids for vertices and edges, and
+embeddings go in as `float32` arrays. The examples below start from the Parquet files the benchmark
+suite prepares and read them with pandas, selecting only the columns a type needs, because
+ArcadeDB has no Parquet or Arrow reader. Loading costs more than in the specialists. On the
+Cross Validated corpus, ArcadeDB needs about 17.5 s for the 426k posts and about 27 s for the graph.
+SQLite and DuckDB read the same Parquet files natively in about 0.3 s, and LadybugDB bulk-loads
+the graph with `COPY` in about 0.6 s ([](#tbl-document), [](#tbl-graph)). Little of that gap is
+the Python boundary. Most of the load time goes to the unique hash index on `id` that the
+point-lookup workload relies on.
```
jameslamb:
```suggestion
the graph with `COPY` in about 0.6 s ([](#tbl-document), [](#tbl-graph)). Most of the load time goes to the unique hash index on `id` that the
point-lookup workload relies on.
```

I suspect this is another yet another LLMism. LLMs love to insert 0-information vague statements like "Little of that gap is the Python boundary.". Recommend just removing this.

jameslamb (approved):
I did a quick read-through tonight (I'm sorry, but I cannot spare the time for another thorough line-by-line review). I focused only on the written prose and did not re-check code samples or the visual appearance of the preview.

It is MUCH improved from the first version I read thanks to what appears to me to be your human-edited changes (my concerns about "AI voice" and "LLMisms" in the paper have been mostly addressed). Thank you for working through the review comments.

I left a few more tiny suggestions but they should not be considered blocking.

@cdlindsey @chongshenng I **support** inclusion of the paper in the Proceedings as-is.

--

side-note: very cool project, I'm glad to have learned about it through this experience 😊

papers/taewoon_kim/main.md
tae898: applied

papers/taewoon_kim/main.md
tae898: applied

papers/taewoon_kim/main.md
tae898: applied

tae898:
@jameslamb @cdlindsey @chongshenng @rowanc1 

Everything looks good now.
Thank you all for taking time to review my paper!  :smile:

chongshenng:
Many thanks @jameslamb for the re-review, and @tae898 for addressing the comments promptly. Happy to accept the paper as-is.

chongshenng (approved):

